PR #38, opened 2024-03-02
Add initial Dockerfile setup

isZumpo:
Hi, I am really happy to see this project coming along nicely! One issue I always had with the original BirdNET-Pi project was that is did not support docker. Which made it really hard to try out before deciding to run it on a raspberry, since it pretty much required a dedicated VM...

While the go approach for this project makes it way easier to try out, I still believe a docker image could be beneficial! Therefore, I would like to make a contribution on getting us started with a initial dockerfile.

This PR allows for BirdNET-Go to be built and bundled into a container. The container will by default execute
the realtime command, starting the server. The image is currently based on debian bookworm since I found that easiest to get running. However, in the future it might be better bundle it in a alpine image to reduce the size even more. It might also be of interest to utilize Github Actions to build a docker for every release. Would happily contribute to that in the future too!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+Dockerfile
+.dockerignore
+.gitignore
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM golang:1.22.0-bookworm as build
+
+ARG TESNSORFLOW_VERSION=v2.14.0
+ARG PLATFORM=linux_amd64
+
+# Download and configure precompiled TensorFlow Lite C library
+RUN curl -L \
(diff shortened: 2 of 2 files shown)


tphakala:
Thank you! Contributions are much appreciated

isZumpo:
Wow that was quick! One issue I experienced that you might be aware of how to fix is that if I don't copy in the .git folder while building the project in docker, then it will try to grab birdnet-go stuff from github rather than locally. I am not that experienced with go... Do you know if there is some way to work around that?

tphakala:
I guess it is fetching go dependencies, like GORM, SQLite driver etc. modules, but I am not sure. I need to test this and see what it does.

isZumpo:
Just double checked. Appears to not be the case anymore, probably some unrelated issue I had while constructing the dockerfile. Removing .git works fine now!

PR #39, opened 2024-03-02
Add basic github action for building docker image

isZumpo:
Implements a basic github action job that will for every commit to the main branch, build and push a docker
image to the latest tag. The latest tag will get overwritten after each commit.

In the future it might be better to instead decide on some better tagging strategy.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
 Dockerfile
 .dockerignore
 .gitignore
+.git
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,4 @@ Dockerfile
 .dockerignore
 .gitignore
 .git
+.github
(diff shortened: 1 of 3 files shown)


isZumpo:
@tphakala there appears to be some sort of permission error now when the action runs:

**denied: installation not allowed to Create organization package**

I did not experience this when running it in my fork. Perhaps there is a setting that must be adjusted for the repository?

tphakala:
It was a privilege issue, now fixed.

isZumpo:
> It was a privilege issue, now fixed.

Great! Now it seems to be running as expected. Could you also make the docker packages public? Then it should be accessible for everyone.

tphakala:
done!

tphakala:
I updated workflow to support multiplatform builds, at the moment it builds linux/amd64 and linux/arm64

PR #40, opened 2024-03-02
Define docker volumes & add docker documentation

isZumpo:
Defines a /config and /data volume for the docker image where the end-user can mount in filesystems from the host. Allowing for these volumes to persist outside the container.

I also added a very basic example on how to run the container in the README.md. It can probably be improved a bit, but it is a start!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
 Dockerfile
 .dockerignore
 .gitignore
+.git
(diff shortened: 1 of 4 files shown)


isZumpo:
This PR is dependent on this PR: https://github.com/tphakala/birdnet-go/pull/39

tphakala:
Merged, thank you for contributing

PR #42, opened 2024-03-03
feat(docker): add multiplatform build support

tphakala:


isZumpo:
Nice! I was fiddling a bit with getting alsa-utils into the image, but had to go away for some errands. Guess you made it before me :+1:

PR #43, opened 2024-03-03
Enable usage of rootless docker

isZumpo:
I'm using podman and had some issues with the most recent change. This adds the runtime user to the audio group, and adds instructions on how to mount in the audio devices to the container.


Dockerfile
tphakala, line 1: `RUN usermod -a -G audio root`
is not required, root already has access to sound device.
"ERROR: User 'root' is not member of audio group, add user to audio group by executing" message is wrong in this respect, which I need to fix in audio capture code.

tphakala (approved):

Dockerfile
isZumpo: Yes it appears that you are right. Seems to be not required anymore. Good catch!

PR #50, opened 2024-03-10
support RTSP by using `ffmpeg` to convert the stream

### LINKED ISSUE #12: Add rtsp source support for realtime analysis
Realtime analysis should support RTSP streams as source for audio data

example

`birdnet realtime --rtsp rtsp://user@pass@url`

twt--:
adding very basic support for RTSP.

usage: `birdnet-go realtime --rtsp rtsp://...`
if an rtsp URL is not provided, it will default to using the audio device.

I've been running this for the past week without issues.

resolves #12

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,13 +37,14 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
     libasound2 \
     sox \
+    ffmpeg \
     && rm -rf /var/lib/apt/lists/*
 
 COPY --from=build /root/src/BirdNET-Go/bin /usr/bin/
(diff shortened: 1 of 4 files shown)


tphakala:
Thank you, much appreciated!

enjikaka:
Is multiple RTSP streams supported?

tphakala:
No, not yet.

tphakala:
But I think there is a workaround, I have not tested this but I don't think that anything prevents you from running multiple instances of birdnet-go which each point to same SQLite file. Enable Web UI for only one of processes and you should see detections from each stream in single view.

PR #54, opened 2024-03-13
Add in a --rtsptransport flag to realtime command

PeteLawrence:
By default ffmpeg will use UDP when streaming from an RTSP source. Some setups (in my case when running the docker container in a Kubernetes cluster) won't work with a UDP connection, and so require a TCP connection instead.

This PR adds in an --rtsptransport flag, which defaults to udp, but can be set to tcp if required.

This is my first attempt working with Go, so please feel free to point me in the right direction if there's a better way to be doing something!


PeteLawrence:
I've been using this for a little bit now and seems to be working as expected, so happy for you to merge.

tphakala:
Thanks! I modified RTSP transport config layout slightly, I added a RTSP node which contains URL and Transport settings, so config format is now

```
realtime:
  rtsp:
    url: rtsp://localhost:554/etc
    transport: tcp
```

@twt-- fyi

PR #88, opened 2024-03-28
Test docker image before pushing to registry

isZumpo:
As mentioned before, I thought it would be nice to have some sort of verification of the docker images. This pr adds such a feature by starting the image and giving it 5 sec to respond to a http request. Only if it responds within time, then the docker image(s) will be pushed to the registry.

Had to do a bit of fighting with the github action runner since it does not contain any alsa devices, causing the container to not start properly. I was able to work around this by installing the alsa dummy soundcard. Unfortunately, there appears to be certain issues with ubuntu-22 due to recent kernel changes, meaning that I had to downgrade to ubuntu-20.04 for this to work. However, there is a discussion ongoing on getting the ubuntu-22 runner behaving as expected in the [works](https://github.com/actions/runner-images/discussions/9074).

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -8,7 +8,8 @@ on:
 
 jobs:
   push-docker-image:
-    runs-on: ubuntu-latest
+    # Ubuntu-20.04 required until this is fixed: https://github.com/actions/runner-images/discussions/9074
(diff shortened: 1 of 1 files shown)


tphakala:
Thank you, this is highly appreciated.

PR #102, opened 2024-04-05
Fix go linting workflow missing tensorflow dependency

isZumpo:
The workflow was failing to run since it did not have the
required tensorflow dependencie. This commit fixes it
by making use of the makefile to install the dependency inside
the workflow.


isZumpo:
Nvm. Did not fix it :(

isZumpo:
I will work on it on my fork instead. Then reopen once it works as expected

Issue #69, opened 2024-03-21
Implement audio file purging


tphakala:
Old wav files should be purged once disk space is starting to get low, purging should be flexible so that users can select which species recordings should be deleted last.

isZumpo:
Should only the audio file be removed or the whole database entry for a given detection?

MueJosh:
> Should only the audio file be removed or the whole database entry for a given detection?


Definitely just the audio files. (It makes much more sense this way)

isZumpo:
> > Should only the audio file be removed or the whole database entry for a given detection?
> 
> Definitely just the audio files. (It makes much more sense this way)

Agreed! The database entries should be of so little size that it does not make any sense to clear them up in comparison to the clips. They could also be useful for long term statistics and insights.

isZumpo:
I got started on a commit where I made a system that would check and make sure that we don't go above 90% disk usage for the given filesystem where the clips are saved. Deleting the oldest clips to always keep it below for example 90%. However, I just kept thinking of edge cases that makes it very hard to not end up in unfortunate situations.  For instance. what if the user decides to put a large blob of data that by itself fills up 90% of the filesystem, then the clips use lets say 1% of the filesystem. Should we then just go ahead and delete all bird clips to keep it below 90%? Feels like there are so many ways to lose important clips that way...

Instead of diving into complicated disk usage algorithms to not end up in these situations above, I thought we instead could come up with some simpler retention rules. One simple idea to get started that would be good enough for a first policy would be this:

We add these two variables to the config:
**MinEvictionHours** - Minimum time before recording is considered for eviction
**MinRecordingsPerSpecies** - Number of recordings to always keep per species

Then every x minutes or on new detections, we would query the database for detections older than **MinEvictionHours** where there are a minimum of **MinRecordingsPerSpecies** for that species. Then delete all clips for those, maybe nulling the database entry for the detection itself?

Does this sound like a reasonable first attempt on implementing some sort of retention strategy? Of course in the future we could consider things as disk utilization, and confidence, rarity of the detection as parameters to some policy.

tphakala:
Audio file purging has been implemented

PR #121, opened 2024-04-15
feat: Add initial retention policy

isZumpo:
As [mentioned](https://github.com/tphakala/birdnet-go/issues/69#issuecomment-2043282550), I wanted to give the simple retention policy a try. Anything is better than nothing as my raspberry keeps running out of space otherwise. I am still trying out this change locally and might make some more changes to it. However, I thought it could still be a good idea to open up this pull request to get some feedback on the idea itself and the code.

The simple retention policy works by every minute, querying the database for clips older than **minEvictionHours**, for these clips, it will then remove those for species that have more than **minClipsPerSpecies**. Always removing the oldest clips first.

I am a bit uncertain of where to place all the logic. Right now I jammed it into the realtime.go and interface.go, but there might be better places to put it? In addition, it has to be decided if this is a feature which should be enabled by default or not. In addition, the spectrogram images are currently saved to the clips folder too, I wonder if it might be better to move them to some other cached directory?

## Summary by CodeRabbit

- **New Features**
	- Introduced automated clip cleanup based on retention policies to manage storage efficiently.
- **Enhancements**
	- Updated settings to allow configuration of clip retention parameters.
- **Tests**
	- Added tests to ensure proper identification and removal of clips according to the new retention policies.

### DIFF
--- a/internal/analysis/realtime.go
+++ b/internal/analysis/realtime.go
@@ -7,6 +7,7 @@ import (
 	"os/signal"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/tphakala/birdnet-go/internal/analysis/processor"
 	"github.com/tphakala/birdnet-go/internal/analysis/queue"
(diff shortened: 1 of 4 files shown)


tphakala:
Thank you, merged.

isZumpo:
That went way quicker than expected. Suppose that we will have to work out those questions as we go instead.

PR #123, opened 2024-04-16
feat: Prometheus Endpoint

aster1sk:
I understand there is interest in keeping birdnet-go as lightweight as possible so instead of building out a full-blown Web UI, perhaps an alternative option could be to export metrics to Prometheus and visualized with existing tools like Grafana.

This PR include a bool configuration option to enable a very basic Prometheus CounterVec as well as all of the boilerplate Go performance metrics.

To enable the Prometheus endpoint, simply include the following in your configuration file.

```
realtime:
    prometheus: true
```

The current implementation is very simple, the following is sample output from development / testing :
```
# HELP birdnet_detections How many birdnet_detections partitioned by common name.
# TYPE birdnet_detections counter
birdnet_detections{name="American Robin"} 4
birdnet_detections{name="Downy Woodpecker"} 2
birdnet_detections{name="House Sparrow"} 10
birdnet_detections{name="Northern Cardinal"} 6
birdnet_detections{name="Pileated Woodpecker"} 1
birdnet_detections{name="Rock Pigeon"} 1
birdnet_detections{name="Ruby-crowned Kinglet"} 1
```

This could easily be improved upon with :
* Dog or human counter events
*  Measuring confidence either by species or overall
* Measuring processing time

## Summary by CodeRabbit

- **New Features**
	- Added support for Prometheus metrics in realtime processing, including tracking BirdNET detections and exposing metrics via an HTTP endpoint.


tphakala:
Thank you, merged.

isZumpo:
This is very nice! I would like to add some more metrics in the future for observability of the health of the application etc :+1: 

Just one question, would it be better to offer the /metrics view at the same port used for serving the webpage itself at? In this case 8080. That would make the port forwarding a bit easier. However, I suppose one benefit of having it on a separate port is that it allows more control of who should have access to the metrics?

tphakala:
I have been thinking of adding Go telemetry for some time , so this is more than welcome. I'd prefer to keep telemetry endpoint separate from main application, for example if people wish to publish their dashboard to Internet it would be at least some level security issue if application telemetry would be exposed as well.

internal/analysis/realtime.go
```diff
@@ -44,6 +47,20 @@ func RealtimeAnalysis(settings *conf.Settings) error {
 	// Initialize database access.
 	dataStore := datastore.New(settings)
 
+	// Initialize Prometheus (if enabled)
+	if settings.Realtime.Prometheus {
+		settings.Realtime.PrometheusDetectionCounter = prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "birdnet_detections",
```
d0ugal, line 54: As this is a counter it probably should have ended with `_total` or `_count`. You'll probably see warnings like this when using PromQL.

> PromQL info: metric might not be a counter, name does not end in _total/_sum/_count/_bucket

aster1sk: Hey nice catch! I don't know anything about PromQL but I see no reason why this small change shouldn't be applied. 

I think my (selfish) goal of this PR was simply to introduce an option to use existing visualization systems instead of building yet another HTML dashboard ( not saying there's anything wrong with the existing one <3 ).

Feel free to make this change if @tphakala's on board.

d0ugal: Thanks - this is just a minor nit. I am using the metric and it works get btw 😄

PR #144, opened 2024-05-05
Default date picker to local time zone

farski:
`new Date().toISOString()` always returns the current time in UTC, so when it's after UTC midnight, but still before UTC midnight of the previous day (say, at 10 PM in New York), the date picker defaults to tomorrow, which will never contain any data.

This changes to use `toLocaleString` to ensure the date picker always reflects the browser's local today date. It uses the Swedish (`sv`) locale simply because they use an `YYYY-MM-DD` date format, which is the most useful in this case


tphakala:
merged, thank you

PR #151, opened 2024-05-07
Clear clip path from note during cleanup when clip is missing

### LINKED ISSUE #149: "Failed to remove clips" forever
It seems like when the process to prune clips is unable to (because the file was removed some other way), it doesn't do anything to prevent trying in the future. If it finds that a clip file has already deleted, it probably should update the DB/etc to reflect that, just as it would do normally.

farski:
When the clip cleanup process tries to delete a clip file, and the OS reports that the file does not exist, this will now detect that condition and still attempt to clear the clipPath from the clip.

Closes #149


isZumpo:
Looks great to me :+1:

PR #156, opened 2024-05-09
Reworked webui reader

matthew73210:
Reworked the webui.log reader. Placed the functions it what seems the right folders. #153 & #138 
Couldn't for the life of me get the templating working like settings, just created a work around in logs.html. Pretty much just index.html copied and removed some parts.
Need to work on making it dynamic as it depends on refreshing the browser.
Being a gitnoob i commited my changes to config, can ignore them.

Might need to be carefull about CSS attacks cus the function reads the webui log and posts as is, if someones puts some code in there ? I think code rabbit will flag it up again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,3 @@ output/soundscape.wav.csv
 birdnet.db
 birdnet.txt
 custom_confidence_list.txt
-conf/config.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ output/soundscape.wav.csv
 birdnet.db
 birdnet.txt
 custom_confidence_list.txt
+conf/config.yaml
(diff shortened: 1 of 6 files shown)


matthew73210:
The webui doesn't have much going on, but the basis is there to read the other logfile too. Might add two links under logs. Or just paste all logs to webui.log and have a single one?

.gitignore
tphakala, line 1: Please do not include this in in pull request, this is unnecessary change

internal/conf/config.yaml
matthew73210, line 1: Not sure what's going on here.

.gitignore
matthew73210: Sorry my bad.

tphakala:
abdandoning this for now

PR #177, opened 2024-05-26
refactor: Refactor configuration package to improve settings handling

tphakala:
Settings are now accessible through the conf.Setting() function. This allows direct access to settings without the need to import a Settings pointer into a class or function.

Accessing Settings

```
// Accessing a setting
if conf.Setting().Debug {
    fmt.Println("Debug is enabled")
} else {
    fmt.Println("Debug is disabled")
}

// Accessing a nested setting
if conf.Setting().Realtime.Audio.Export.Enabled {
    fmt.Println("Audio export is enabled")
} else {
    fmt.Println("Audio export is disabled")
}
```

Updating Settings at Runtime

```
// Copy current settings
newSettings := *conf.Setting()

// Modify the necessary setting
newSettings.Realtime.Audio.Export.Enabled = true

// Update settings
if err := conf.UpdateSettings(&newSettings); err != nil {
    log.Fatalf("Error updating settings: %v", err)
}
```

### DIFF
--- a/internal/conf/config.go
+++ b/internal/conf/config.go
@@ -3,12 +3,12 @@ package conf
 
 import (
 	"embed"
-	"errors"
 	"fmt"
 	"io/fs"
 	"log"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/spf13/viper"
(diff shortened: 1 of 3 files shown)


isZumpo:
Look great! Though, linting is failing on the PR. Would it be possible to enable branch protection rules, such that all actions would have to pass before allowing a merge?

PR #186, opened 2024-06-01
feat: Initial support for multiple RTSP streams

tphakala:
This PR adds support for multiple RTSP streams, streams are configured by listing --rtsp setting multiple times for each stream

### DIFF
--- a/internal/conf/config.go
+++ b/internal/conf/config.go
@@ -90,8 +90,8 @@ type Settings struct {
 		}
 
 		RTSP struct {
-			Url       string // RTSP stream URL
-			Transport string // RTSP Transport Protocol
+			Transport string   // RTSP Transport Protocol
+			Urls      []string // RTSP stream URL
 		}
 
 		MQTT struct {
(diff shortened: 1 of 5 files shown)


internal/conf/config.go
```diff
@@ -90,8 +90,8 @@ type Settings struct {
 		}
 
 		RTSP struct {
-			Url       string // RTSP stream URL
-			Transport string // RTSP Transport Protocol
+			Transport string   // RTSP Transport Protocol
+			Urls      []string // RTSP stream URL
```
xconverge, line 94: @tphakala I think you need to rename this now too, to ```urls``` and maybe set it to [] 

https://github.com/tphakala/birdnet-go/blob/6bc12a5e81ec3bc311cfbcf778819025b6e04a92/internal/conf/defaults.go#L57

tphakala: This is already fixed in latest PR I submitted yesterday

xconverge: I still see the config as a single URL  where I linked on the master branch ..

tphakala: you are correct, missed changes for defaults

jasonmadigan:
This is great to have! Any chance we might be able to add labels to rtsp inputs (to note where detections occur?)

tphakala:
> This is great to have! Any chance we might be able to add labels to rtsp inputs (to note where detections occur?)

Yes, I have been thinking of doing this.

PR #190, opened 2024-06-02
Minor cleanup

xconverge:
Just starting to look into the code base to understand why I am getting some VERY strange (not possible) bird results and found some minor cleanup


tphakala:
Latest BirdNET range filter model is likely to be defective, there is discussion about it here https://github.com/kahst/BirdNET-Analyzer/issues/296

PR #200, opened 2024-06-05
feat: make range filter model selectable between latest and previous …

tphakala:
…"legacy" version.

Also introduces a sliglty changed birdnet config layout with new config node birdnet.rangefilter with mode and threshold (renamed from birdnet.locationfilterthreshold) settings

```
# BirdNET model specific settings
birdnet:
  sensitivity: 1.0      # sigmoid sensitivity, 0.1 to 1.5
  threshold: 0.8        # threshold for prediction confidence to report, 0.0 to 1.0
  overlap: 0.0          # overlap between chunks, 0.0 to 2.9
  threads: 0            # 0 to use all available CPU threads
  locale: en            # language to use for labels
  latitude: 00.000      # latitude of recording location for prediction filtering
  longitude: 00.000     # longitude of recording location for prediction filtering
  rangefilter:
      model: latest     # model to use for range filter: "latest" or "legacy" for previous model
      threshold: 0.01   # rangefilter species occurrence threshold
```

### DIFF
--- a/internal/conf/config.yaml
+++ b/internal/conf/config.yaml
@@ -22,7 +22,9 @@ birdnet:
   locale: en            # language to use for labels
   latitude: 00.000      # latitude of recording location for prediction filtering
   longitude: 00.000     # longitude of recording location for prediction filtering
-  locationfilterthreshold: 0.01 # rangefilter species occurrence threshold
+  rangefilter:
+      model: latest     # model to use for range filter: "latest" or "legacy" for previous model
+      threshold: 0.01   # rangefilter species occurrence threshold
 
 # Realtime processing settings
 realtime:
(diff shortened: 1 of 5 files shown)


xconverge:
Thanks a lot @tphakala !

internal/conf/config.yaml
```diff
@@ -22,7 +22,9 @@ birdnet:
   locale: en            # language to use for labels
   latitude: 00.000      # latitude of recording location for prediction filtering
   longitude: 00.000     # longitude of recording location for prediction filtering
-  locationfilterthreshold: 0.01 # rangefilter species occurrence threshold
+  rangefilter:
+      model: latest     # model to use for range filter: "latest" or "legacy" for previous model
```
xconverge, line 26: this should be ```version: latest```

internal/conf/defaults.go
```diff
@@ -26,7 +26,8 @@ func setDefaultConfig() {
 	viper.SetDefault("birdnet.locale", "en")
 	viper.SetDefault("birdnet.latitude", 0.000)
 	viper.SetDefault("birdnet.longitude", 0.000)
-	viper.SetDefault("birdnet.locationfilterthreshold", 0.01)
+	viper.SetDefault("birdnet.rangefilter.model", "latest")
```
xconverge, line 29: this shoudl be ```birdnet.rangefilter.version```

PR #201, opened 2024-06-05
Improve devcontainer experience

isZumpo:
* Add port forwarding of 8080 in devcontainer such that it will automatically start working, rather than waiting for vscode to detect that it is open
* Air now goes by a new repo name and the reference has been updated
* Now when starting the devcontainer you will be prompted for which recording device to use:
image
* A dev-user is added for the buildenv stage, meaning that most steps will not run as root directly anymore. For devcontainer users, this has the advantage of being less likely to be able to create root-owned files on the host system

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,7 @@
 {
 	"name": "Go Build Environment",
 	"dockerFile": "../Dockerfile",
+	"forwardPorts": [8080],
 	"build": {
 		"context": "..",
 		"target": "buildenv"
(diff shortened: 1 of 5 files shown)


isZumpo:
Will have to try out the last commit a bit before opening this pull request up for merge/review.  I want to be sure that there will be no file permission issues when mounting in volumes from the host

isZumpo:
> Will have to try out the last commit a bit before opening this pull request up for merge/review. I want to be sure that there will be no file permission issues when mounting in volumes from the host

Ok skipping that last commit since it requires more verification. All else should be good though :)

PR #203, opened 2024-06-08
feat: add range print command which lists all species included by ran…

tphakala:
…ge filter model

### DIFF
--- a/cmd/rangefilter/print.go
+++ b/cmd/rangefilter/print.go
@@ -1,0 +1,50 @@
+// print.go range print command code
+package rangefilter
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/tphakala/birdnet-go/internal/birdnet"
+	"github.com/tphakala/birdnet-go/internal/conf"
+)
+
+// PrintCommand creates the print subcommand
+func PrintCommand(settings *conf.Settings) *cobra.Command {
+	printCmd := &cobra.Command{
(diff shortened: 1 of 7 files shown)


cmd/rangefilter/print.go
```diff
@@ -0,0 +1,50 @@
...
+	// Define flags for the print subcommand
+	printCmd.Flags().Float64Var(&settings.BirdNET.Latitude, "latitude", settings.BirdNET.Latitude, "Latitude for range filter")
+	printCmd.Flags().Float64Var(&settings.BirdNET.Longitude, "longitude", settings.BirdNET.Longitude, "Longitude for range filter")
+	printCmd.Flags().Float32Var(&settings.BirdNET.RangeFilter.Threshold, "threshold", settings.BirdNET.RangeFilter.Threshold, "Threshold for range filter")
+	printCmd.Flags().StringVar(&settings.BirdNET.RangeFilter.Model, "model", settings.BirdNET.RangeFilter.Model, "Model for range filter")
+
+	return printCmd
+}
+
+// GetCurrentDateFormatted returns the current date formatted as a string based on the given format ("eu" for European and "us" for US).
```
xconverge: [nit] Seems like a good candidate (and basically every in this SW when reasonable) for ISO8601 YYYY-MM-DD and remove the formatter/complexity unless you have a specific reason to be using those formats, just my opinion though!

tphakala: Good point

PR #220, opened 2024-06-18
refactor: improve thumbnails code, add cache size metrics for telemetry

tphakala:


tphakala:
I removed attributions from table view, I'll add them visible as "on hower" elements. I will also add config switch to enable thumbnails for daily totals table as images take a lot of space.

PR #235, opened 2024-06-27
refactor: add MQTT client unit tests and use node name as MQTT client id

tphakala:


tphakala:
@coderabbitai full review

PR #238, opened 2024-06-28
fix: Hourly occurences not being calculated correctly when using MySQL

PeteLawrence:
When using a MySQL database the hourly occurrences table does not render correctly due to DATE_FORMAT being called on a time field. This MR switches the DATE_FORMAT call for TIME_FORMAT


tphakala:
Merged, thank you

PR #240, opened 2024-06-29
refactor: telemetry package refactored to modular structure

tphakala:
The telemetry package has been refactored to a modular structure to improve code organization and maintainability. The package now consists of multiple modules, each responsible for a specific aspect of telemetry data collection and processing.

Added metrics collection to MQTT module
Added BirdNET processing time metrics

### DIFF
--- a/internal/imageprovider/imageprovider.go
+++ b/internal/imageprovider/imageprovider.go
@@ -13,7 +13,7 @@ import (
 
 // ImageProvider defines the interface for fetching bird images.
 type ImageProvider interface {
-	fetch(scientificName string) (BirdImage, error)
+	Fetch(scientificName string) (BirdImage, error)
 }
 
 // BirdImage represents the metadata of a bird image.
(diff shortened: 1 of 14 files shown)


internal/imageprovider/imageprovider.go
```diff
@@ -1,95 +1,120 @@
...
 	"unsafe"
 
 	"github.com/tphakala/birdnet-go/internal/telemetry"
+	"github.com/tphakala/birdnet-go/internal/telemetry/metrics"
 )
 
+// ImageProvider defines the interface for fetching bird images.
 type ImageProvider interface {
-	fetch(scientificName string) (BirdImage, error)
+	Fetch(scientificName string) (BirdImage, error)
```
isZumpo, line 16: Why is this function made public?

internal/imageprovider/imageprovider.go
```diff
@@ -1,95 +1,120 @@
...
+	c.nonBirdImageProvider = provider
+}
+
+// SetImageProvider allows setting a custom ImageProvider for testing purposes.
+func (c *BirdImageCache) SetImageProvider(provider ImageProvider) {
+	c.birdImageProvider = provider
+}
+
+// initCache initializes a new BirdImageCache with the given ImageProvider.
+func InitCache(e ImageProvider, t *telemetry.Metrics) *BirdImageCache {
```
isZumpo, line 55: Same with this one

internal/imageprovider/imageprovider.go
```diff
@@ -1,95 +1,120 @@
...
+		c.metrics.IncrementCacheHits()
 		return birdImage.(BirdImage), nil
 	}
 
-	// Fetch the bird image from the image provider
 	fetchedBirdImage, err := c.fetch(scientificName)
 	if err != nil {
-		// TODO for now store an empty result in the cache to avoid future queries that would fail.
-		// In the future, look at the error and decide if it was caused by networking and is recoverable.
-		// And if it was, do not store the empty result in the cache.
```
isZumpo, line 81: Still think this todo is valid and should not be removed

internal/imageprovider/imageprovider.go
tphakala: For unit tests to work

internal/imageprovider/imageprovider.go
tphakala: Some comments were removed unintentionally, like this

internal/imageprovider/imageprovider.go
isZumpo: Can't the unit tests be in the same package as before?

internal/imageprovider/wikipedia.go
```diff
@@ -163,47 +167,40 @@ func (l *wikiMediaProvider) queryAuthorInfo(thumbnailURL string) (*wikiMediaAuth
...
 
 	if len(wikipediaUserLinks) == 1 {
-		// Return the href and inner text of the Wikipedia user link
 		wikipediaLink := wikipediaUserLinks[0]
 		href = extractHref(wikipediaLink)
 		text = extractText(wikipediaLink)
 		return href, text, nil
 	}
 
-	// Check if all the links have the same href value
```
isZumpo, line 206: All of these comments might also be useful to detangle the reason for the differerent if cases

internal/imageprovider/imageprovider.go
isZumpo: Ah ok I understand :)

Issue #118, opened 2024-04-10
Recommended hardware documentation


tphakala:
There should be a recommended hardware documentation

- Low cost quick start setup
- Mobile or remote setup (maybe)
- Cost-effective high fidelity setup

Also recommended storage solutions (sd cards, USB drives, NAS, etc)

florisre:
Where from should this page be linked to?

PR #245, opened 2024-07-05
Documentation of Recommended Hardware

florisre:
This is my initial list of my reccomendations. Feel free to add upon this! 

Starts work on https://github.com/tphakala/birdnet-go/issues/118 and will eventually close that.

### DIFF
--- a/doc/recommended_hardware.md
+++ b/doc/recommended_hardware.md
@@ -1,15 +1,15 @@
 # List of Recommended Hardware
-BidrNet-Go works with many many different setups. This is meant to give an overview of what has been tried and found working. 
+BirdNet-Go works with many,  many different setups. This is meant to give an overview of what has been tried and found working.
 ## Microphones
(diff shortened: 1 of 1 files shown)


isZumpo:
Nice work! I am also using the AOM-5024. One thing related to it which I had issues with was finding a online store that sold these units without insane shipping costs (within europe). In the end I went with Reichelt, where you still had to pay ~10 euro shipping for a ~3 euro mic. Maybe a small table could added at some point in the future with resonable priced retailers for the hardware itself for different parts of the world.

florisre:
We should probably also link to this document somewhere, maybe installation instructions or as an additional wiki page? 

@isZumpo feel free to add this! I eventually sourced mine via eBay from some dude who sold ready soldered microphones and sets of parts to self-solder. Many pages had astronomically high shipping cost or would require you to at least buy 100 pieces. Maybe we could even set up collective buys?

PR #270, opened 2024-08-26
refactor: major rewrite of ffmpeg process management

tphakala:


jasonmadigan:
This works really well and seems to have solved some of the odd crashing issues/instability I've occasionally seen after running the latest dev image for a few hours. 

Wondered when we'd see this in a release? (Had been running this as a hassio addon from alexbelgium, but I think the addon generally tracks stable releases).

tphakala:
Sorry I am having a slight mental block about next release, there are still few things I need to get working before I deem it releaseworthy. I

joriws:
Sorry commenting old issue, I am using pure rtsp input for Birdnet-Go (BNG). Sometimes something goes wrong, surveillance camera (or something) stops RTSP feed. 

BNG does not react anything. On CLI I need to start container - but what I'd like to see is better stream handling with BNG like if no new audio frames coming from FFMPEG - restart the FFMPEG/stream. This works if no initial data at launch of BNG, but already established BNG runtime it just silently (no log) sits.

I've had many occations of this. Maybe go2rtc would be the solution to keep RSTP stable, but I'd like to see effort also from BNG-side.

tphakala:
@joriws maybe this PR solves it? I have not tested it yet https://github.com/tphakala/birdnet-go/pull/365

PR #320, opened 2024-11-15
fix: only validate required input fields

petterip:


tphakala:
Hi @petterip, a licensing note about your contribution here.

I'm planning to relicense BirdNET-Go from CC BY-NC-SA 4.0 to AGPL-3.0, an OSI-approved open source license. It would keep the copyleft, share-alike spirit but remove the non-commercial restriction and make the project properly open source. There are no plans to commercialize the project; this is to align better with the open source community and make packaging easier. Full context is in #4243, and consent is tracked in #4244.

Since your work is part of the codebase, I'd like your agreement before relicensing. Could you reply confirming:

Thanks a lot for this contribution.

PR #341, opened 2024-12-22
fix: remove redundant copy of reset_auth.sh

isZumpo:
For some reason the reset_auth.sh script was copied over to the build
stage, not used there, and then copied over to the runtime stage. This
is redundant and can be removed.

Updates Dockerfile to remove the redundant copy. Instead copies directly
from the host into the runtime stage. In addition, removes the chmod
since the file is already executable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,6 @@ USER dev-user
 WORKDIR /home/dev-user/src/BirdNET-Go
 
 COPY --chown=dev-user ./Makefile ./
-COPY --chown=dev-user ./reset_auth.sh ./
 
 # Download TensorFlow headers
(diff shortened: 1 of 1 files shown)


tphakala:
Hi @isZumpo, a quick licensing note.

The project has no contributor licence agreement, so you keep copyright over your contributions. You've sent a fair few over the years, so before I relicense I'd like your explicit agreement. Could you reply here confirming:

I'm tracking consent in #4244. Thanks for all the work, and happy to answer any questions about the change.

isZumpo:
@tphakala sounds good to me

I agree that my contributions to birdnet-go, past and future, may be licensed under AGPL-3.0.

PR #363, opened 2025-01-02
feat: add birdnet-user to docker runtime image

### LINKED ISSUE #322: Feature request - docker security
## Issue Type
- [ ] Feature request

## Description
This is a very cool service - thanks.
However, the container is configured to run as root, and if configured to run as a user, it fails with:

` 2024/11/20 21:51:55 Error loading settings: error initializing viper: error creating directories for config file: mkdir /.config: permission denied`

Please can the service be enhanced to run without root priviliges.

isZumpo:
Fixes #322 

**Changes:**

   - Introduces a non-root user `birdnet-user` to the Docker image to enhance security and prevent potential vulnerabilities.
   - The entrypoint script utilizes `gosu` to downgrade to the less privileged `birdnet-user` (gid=10002, uid=10002), ensuring the container does not run as the root user.
   - Adjusts volume mount permissions for `/data` and `/config` to the new user during the downgrade process.
   - Supports running the container with an arbitrary user/group ID by passing `--user` into the `docker run` command. Safeguards are in place to ensure the specified user/group ID has the required permissions to the mounted volumes and belongs to the audio group.
   - Introduces the `CONFIG_PATH` environment variable to allow setting the configuration path. This is particularly useful when running inside a container, as the configuration might be mounted in a different path than expected when running on a real machine.

These changes should not leave any existing volumes in a bad state and should be easy for everyone to use. However, it would be good if someone else could try out these changes to be on the safe side before we merge anything.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,6 +58,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libasound2 \
     ffmpeg \
     sox \
+    gosu \
     && rm -rf /var/lib/apt/lists/*
 
 ARG TFLITE_LIB_DIR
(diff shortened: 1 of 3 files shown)


tphakala:
Sorry for delay, I will review this during weekend. This is needed but I am unsure how to handle this without breaking existing deployments on upgrade.

isZumpo:
I think it should be quite safe since it still will start as root (just like before) and downgrade itself. Still good to review it properly though since we don't want to mess up a bunch of installations.

tphakala:
closing this as it got obsoleted by https://github.com/tphakala/birdnet-go/pull/636

Issue #351, opened 2024-12-27
MQTT message(s) sent while detection(s) are ignored


HWittingen:
## Issue Type
- [x] Bug report
- [ ] Feature request
- [ ] Question / Support

## Description
In the config.yaml the realtime/interval is set to 15. Which makes that next same species detections within 15 seconds are dropped and not registered. 
But although not registered the detections clearly does get messaged through MQTT if that is integration path is set to on.

## Steps to Reproduce (for bugs)
1. [First Step]
2. [Second Step]
3. [and so on...]

## Expected behavior
I'd expected that when a detection is dropped due to above mentioned setting this also means that it's excluded rfrm any further (external) reporting too. It should by fully dropped.  

## Actual behavior
Although the detection does not get registered it is messaged to any interested MQTT client if MQTT integration is on.

## Screenshots
![Screenshot_20241227_120117_Edge](https://github.com/user-attachments/assets/451cbef6-b0cf-42f3-ac7e-a8c5a9d91927)
![Screenshot_20241227_120150_IoT MQTT Panel Pro](https://github.com/user-attachments/assets/c9a50341-2d5a-4d8b-acc0-dd0bea73ce35)

### Version Information (build date)
Provide the following line from your system:
```
$ ./birdnet-go
[Paste the output line here, which should look like this:]
BirdNET-Go build date: YYYY-MM-DDTHH:MM:SSZ, using config file: /path/to/config.yaml
```

### System Information
- Device: 
  - [ ] Raspberry Pi (specify model, e.g., Raspberry Pi 4 Model B)
  - [ ] Other SBC (please specify)
  - [ ] PC/Server (please specify)
- OS: [e.g. Raspberry Pi OS, Ubuntu, Debian]
- OS Version: [e.g. Raspberry Pi OS Bullseye, Ubuntu 20.04.2 LTS]
- Hardware specs:
  - CPU: [e.g. Raspberry Pi 4 Broadcom BCM2711, quad-core Cortex-A72 (ARM v8) 64-bit SoC @ 1.5GHz]
  - RAM: [e.g. 4GB]
  - Storage type and size: [e.g. 32GB microSD card, 128GB SSD]
- Additional hardware (if applicable):
  - USB Audio device: [e.g. Brand, Model]
  - Other peripherals: [e.g. Specific microphone, GPS module]

### For Non-Raspberry Pi Linux Users
Have you tested with the latest dev branch container image?
- [ ] Yes
- [ ] No
[...]

PR #393, opened 2025-01-15
Event tracker fixes

tphakala:
- Event tracker functionality was using hard coded interval values, fixed code to use realtime.interval value instead
- MQTT publish action frequency was not controlled by event tracker


tphakala:
Fixes https://github.com/tphakala/birdnet-go/issues/351

PR #407, opened 2025-01-24
Fix docker build process for nightly builds

tphakala:


d0ugal:
Thanks for this. I tested the nightly build on my raspberry pi and it worked great.

PR #438, opened 2025-02-09
Fix failing test and linter error

vibragiel:
This fixes the `TestGetClipsQualifyingForRemoval` test, that has been failing for a while.

image

And this linter error:

image

Or I think so, because I wasn't able to replicate the build environment.


vibragiel:
Okay I saw you took care of this already :rocket:

PR #446, opened 2025-02-10
security: Add CSRF protection to login and authentication routes

tphakala:
- Configure CSRF middleware with custom settings
- Add CSRF token generation and rendering in login page
- Expose CSRF token in login template
- Modify security functions to support local subnet detection
- Improve security middleware configuration

Proposed solution to fix https://github.com/tphakala/birdnet-go/issues/314


tphakala:
@petterip can you please review this?

Issue #502, opened 2025-02-28
container keeps growing - clips not being deleted


uSlackr:
## Issue Type
- [x ] Bug report
- [ ] Feature request
- [ ] Question / Support

## Description
My container keeps growing - now at 10GB and disk space is 97.5%.  It appears clips aren't going away.  I found the following in the log:
`Error cleaning up clips:  parsing time "20250130T184446Z.mp3": extra text: ".mp3"`

My Retention Policy is `Usage`
Max Usage: `80%`
Minimum Clips: `10`

## Steps to Reproduce (for bugs)
1. Run container
2. Set Retention to Usage
3. Let run
4. Error logged at failed deletion

## Expected behavior
Disk space usage shouldn't grow past 80%
## Actual behavior
Disk fills

## Screenshots
[If applicable, add screenshots to help explain your problem]

### Version Information (build date)
Provide the following line from your system:
```
$ ./birdnet-go
[Paste the output line here, which should look like this:]
BirdNET-Go build date: YYYY-MM-DDTHH:MM:SSZ, using config file: /path/to/config.yaml
```

### System Information
- Device: 
  - [ ] Raspberry Pi (specify model, e.g., Raspberry Pi 4 Model B)
  - [ ] Other SBC (please specify)
  - [x ] PC/Server proxmox docker in lxc
- OS: proxmox
- OS Version: [e.g. Raspberry Pi OS Bullseye, Ubuntu 20.04.2 LTS]
- Hardware specs:
  - CPU: [e.g. Raspberry Pi 4 Broadcom BCM2711, quad-core Cortex-A72 (ARM v8) 64-bit SoC @ 1.5GHz]
  - RAM: [e.g. 4GB]
  - Storage type and size: [e.g. 32GB microSD card, 128GB SSD]
- Additional hardware (if applicable):
  - USB Audio device: [e.g. Brand, Model]
  - Other peripherals: [e.g. Specific microphone, GPS module]

### For Non-Raspberry Pi Linux Users
Have you tested with the latest dev branch container image?
- [ x] 
- [ ] No
- [ ] Not applicable (Windows/macOS user)

If not, first test with the latest dev branch container image and see if the issue is already fixed.
```
$ docker pull ghcr.io/tphakala/birdnet-go:dev
```

## Additional context
I'm I am able to delete files from inside the docker container consoile
## Possible Solution
[If you have a suggestion for fixing the bug or implementing the feature]

tphakala:
[...]

PR #503, opened 2025-02-28
fix: Fix diskmanager file type protection check and add unit testing

tphakala:


tphakala:
This PR fixes https://github.com/tphakala/birdnet-go/issues/502

PR #616, opened 2025-04-10
Allow files with .m4a extensions to be cleaned up

Phaeton:
AAC Export Type creates `.m4a` files. This PR allows these files to be pruned per the retention policy.


tphakala:
Thank you for this fix

PR #631, opened 2025-04-15
feat: add host IP resolution and subnet checking for containerized en…

tphakala:
…vironments

- Introduced `GetHostIP` function to resolve the host IP address when running in a container, including fallback mechanisms for Linux.
- Enhanced `RealIP` method in the server to return the actual host IP if the client appears to be localhost in a container.
- Added `IsInHostSubnet` function to check if a given IP is in the same subnet as the host, improving local network connection validation.
- Updated `IsInLocalSubnet` to utilize the new host subnet checking when running in a container.

These changes improve the handling of IP addresses in containerized deployments, enhancing network security and connectivity.

## Summary by CodeRabbit

- **New Features**
	- Improved detection of the host machine's IP address, including enhanced support for containerized environments.
- **Bug Fixes**
	- Addressed potential issues with nil client IP addresses during local network checks to prevent errors.
- **Enhancements**
	- More accurate client IP reporting when running inside containers.
	- Improved subnet membership checks for containerized deployments, ensuring correct local network detection.
	- Updated systemd service configuration to ensure proper Docker host gateway resolution inside containers.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1219,6 +1219,7 @@ Restart=always
 ExecStart=/usr/bin/docker run --rm \\
     -p ${WEB_PORT}:8080 \\
     --env TZ="${TZ}" \\
+    --add-host="host.docker.internal:host-gateway" \\
     ${AUDIO_ENV} \\
     -v ${CONFIG_DIR}:/config \\
     -v ${DATA_DIR}:/data \\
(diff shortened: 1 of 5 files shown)


PaulOckenden:
If this was supposed to make audio available from the local LAN without logging in using docker based installs I'm sorry to have to report that it hasn't worked. 

Shout if you want me to run any debug things to help you.

Issue #691, opened 2025-05-02
Can't publish MQTT


dziban303:
## Issue Type
- [x] Bug report
- [ ] Feature request
- [ ] Question / Support

## Description
Birdnet is no longer able to publish MQTT messages, though it was working last week.

```
irdnet  | 2025/05/02 06:42:33 🔄️ Retrying e12b87e5: Publish detection to MQTT, attempt 4/4
birdnet  | 2025/05/02 06:42:36 ❌ Message Publishing: Message Publishing operation timed out: operation timeout
birdnet  | 2025/05/02 06:42:59 Total time for processing all notes: 465.78µs
birdnet  | 2025/05/02 06:42:59 Retrieved 5 detections in 677.055µs
birdnet  | 2025/05/02 06:43:03 ⚠️ Job e12b87e5: Publish detection to MQTT permanently failed after 4 attempts: job execution timed out after 30 seconds: context deadline exceeded
```

Additionally, sending test messages fails as well

![Image](https://github.com/user-attachments/assets/0532db45-257f-4168-a050-3f825e2b19ae)

```
birdnet  | 2025/05/02 07:02:15 ✅ Message Publishing: Running Message Publishing test...
birdnet  | 2025/05/02 07:02:20 ❌ Message Publishing: Message Publishing operation timed out: operation timeout
```

Mosquitto sees birdnet trying to connect:
```
1746187871: New connection from 172.23.0.1:43540 on port 1883.
1746187871: New connection from 172.23.0.1:43544 on port 1883.
1746187871: New client connected from 172.23.0.1:43544 as auto-35E2602A-07A8-2894-90B4-D64D095BB35A (p2, c1, k30).
1746187871: Client <unknown> closed its connection.
```
It seems to be trying to connect twice, I don't know enough about mqtt to know if that's very unusual or not. I don't see any other clients doing that, fwiw.

I haven't made any changes to the config between when it last worked and now, though there is an issue when editing the config in the web interface in that it demands an mqtt password if you make any changes to mqtt settings there. I have anonymous publication turned on in my mqtt server, a username and password is unnecessary:

![Image](https://github.com/user-attachments/assets/8adb2a67-dbb0-4ed0-9fda-9b16bc535543)

So any modifications to mqtt settings need to be made in the `config.yaml` file. 

I've also tried using `mqtt://<ip>` as well as `tcp://<ip>`, I've tried various topics. I even created a user and password and added those to the config, still fails.

[...]

PR #693, opened 2025-05-02
fix(mqtt): Remove action-based reconnect to prevent race condition

tphakala:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling when fetching bird images to prevent crashes if the image cache is unavailable or encounters errors.
  - Updated MQTT action behavior to avoid immediate reconnect attempts, relying on background processes for reconnection and providing clearer error messages when the client is disconnected.

- **Tests**
  - Added unit tests for MQTT action execution, covering both connected and disconnected scenarios to ensure correct error handling and message publishing behavior.

### DIFF
--- a/internal/analysis/processor/actions.go
+++ b/internal/analysis/processor/actions.go
@@ -298,32 +298,20 @@ func (a *MqttAction) Execute(data interface{}) error {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


tphakala:
Fixes MQTT client creating two connections as reported here https://github.com/tphakala/birdnet-go/issues/691

PR #1405, opened 2025-10-19
fix(settings): allow clearing BirdNET model/label paths to revert to defaults

### LINKED ISSUE #1404: Bug: Cannot clear BirdNET model/label paths in UI settings to revert to defaults
## Bug Description

When users try to clear the custom BirdNET model path and label path fields in the UI settings page to revert to the embedded defaults, the fields appear to clear in the UI but the values are not actually removed from the config file.

## Steps to Reproduce

1. Set custom model path and label path in Settings > Main > BirdNET Model Configuration
2. Save the settings
3. Later, delete/clear both fields to revert to embedded defaults
4. Click Save
5. Restart BirdNET-Go

**Expected**: The `modelPath` and `labelPath` fields should be removed from config.yaml (or set to empty), causing BirdNET-Go to use the embedded defaults

**Actual**: The config file timestamp updates, but the `modelPath` and `labelPath` values remain unchanged in the config file

## Root Cause

The issue is in the settings merge logic at `internal/api/v2/settings.go`:

In the `deepMergeMaps` function (lines 641-672), the code only clears values when they are `nil` (null in JSON):

```go
if v == nil {
    // If src value is explicitly null, set it to null
    result[k] = nil
    continue
}
```

When the frontend sends an empty string `""` to clear a field, it's not `nil`, so it falls through to line 668 which preserves it. However, empty strings for these configuration fields should signal "revert to default/embedded model".

## Affected Code

- **Frontend**: `frontend/src/lib/desktop/features/settings/pages/MainSettingsPage.svelte` (lines 1224-1242)
- **Backend**: `internal/api/v2/settings.go` (`deepMergeMaps` function, lines 641-672)
- **Settings store**: `frontend/src/lib/stores/settings.ts`

## Solution Options

**Option 1 (Recommended)**: Frontend sends `null` instead of empty string
- Modify the settings store to convert empty strings to `null` for `modelPath` and `labelPath` before sending to API
- This is cleaner and more semantically correct (null = "no value")

**Option 2**: Backend treats empty strings specially
- Update `deepMergeMaps` or add post-processing to treat empty strings as `nil` for specific fields
- Less clean as it requires field-specific logic

## User Report

From Discord user leccelecce:
> I set the new model path in my config.yml. I then saw it can also be specified in the UI config pages. I deleted the values out to disable it, clicked Save and restarted, but it seems it doesn't actually update the config file.
> 
> Oddly, I can see the config file timestamp does change when I click save, but it's not actually resetting the two fields back to "".

## Related Files

- `/frontend/src/lib/desktop/features/settings/pages/MainSettingsPage.svelte`
- `/frontend/src/lib/stores/settings.ts`
- `/internal/api/v2/settings.go`
- `/internal/conf/config.go`

tphakala:
## Summary

Fixes #1404 - Users can now properly clear the BirdNET model path and label path fields in the UI settings to revert to embedded defaults.

### Changes Made

**Frontend** ([settings.ts:814-819](frontend/src/lib/stores/settings.ts#L814-L819)):
- Modified `settingsStore.saveSettings()` to convert empty strings to `null` for `modelPath` and `labelPath`
- This signals to the backend that users want to remove custom paths and revert to embedded defaults

**Backend** ([config.go:744-745](internal/conf/config.go#L744-L745)):
- Added `omitempty` tags to `BirdNETConfig.ModelPath` and `BirdNETConfig.LabelPath` struct fields
- Ensures empty strings are not written to config.yaml, properly cleaning up the file

### Problem Solved

**Before this fix**:
1. User sets custom model path in Settings > Main > BirdNET Model Configuration
2. User later deletes/clears both fields to revert to embedded defaults
3. Fields appear to clear in the UI
4. Config file timestamp updates but `modelPath` and `labelPath` values remain unchanged ❌

**After this fix**:
1. User clears the fields in the UI
2. Frontend sends `null` instead of empty string `""`
3. Config file is properly cleaned (fields removed or set to empty)
4. BirdNET-Go uses embedded defaults ✅

### Technical Details

The fix leverages the existing `deepMergeMaps` function which already handles `null` values correctly:

```go
// deepMergeMaps at line 652-654
if v == nil {
    result[k] = nil
    continue
}
```

The flow is:
1. Frontend sends `null` when fields are cleared
2. Backend's `deepMergeMaps` sets those map values to `nil`
3. When unmarshaling into the `BirdNETConfig` struct, `nil` becomes `""` (empty string)
4. When marshaling to YAML, the `omitempty` tag omits empty strings from the output

### Testing

- ✅ Frontend linters pass (`npm run check:all`)
- ✅ Backend linter passes (`golangci-lint run -v`)
- ✅ TypeScript type checking passes
- Manual testing recommended to verify config file cleanup

## Summary by CodeRabbit

* **Bug Fixes**
  * Model path and label path settings can now be properly cleared and reset to defaults. Empty or whitespace-only values are normalized when saved, allowing users to revert to default behavior.

### DIFF
--- a/frontend/src/lib/desktop/features/settings/pages/MainSettingsPage.svelte
+++ b/frontend/src/lib/desktop/features/settings/pages/MainSettingsPage.svelte
@@ -1223,7 +1223,7 @@
         <div class="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 gap-x-6">
           <TextInput
             id="model-path"
-            value={settings.birdnet.modelPath}
+            value={settings.birdnet.modelPath ?? ''}
             label={t('settings.main.sections.customClassifier.modelPath.label')}
             placeholder={t('settings.main.sections.customClassifier.modelPath.placeholder')}
             helpText={t('settings.main.sections.customClassifier.modelPath.helpText')}
@@ -1233,7 +1233,7 @@
 
           <TextInput
             id="label-path"
-            value={settings.birdnet.labelPath}
+            value={settings.birdnet.labelPath ?? ''}
             label={t('settings.main.sections.customClassifier.labelPath.label')}
             placeholder={t('settings.main.sections.customClassifier.labelPath.placeholder')}
             helpText={t('settings.main.sections.customClassifier.labelPath.helpText')}
--- a/frontend/src/lib/stores/settings.test.ts
+++ b/frontend/src/lib/stores/settings.test.ts
@@ -2,12 +2,13 @@ import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { get } from 'svelte/store';
 import { settingsStore, settingsActions } from './settings';
 import type { BirdNetSettings, RealtimeSettings, SettingsFormData, Dashboard } from './settings';
+import { settingsAPI } from '$lib/utils/settingsApi.js';
 
 // Mock the settings API
 vi.mock('$lib/utils/settingsApi.js', () => ({
   settingsAPI: {
     load: vi.fn(),
-    save: vi.fn(),
+    save: vi.fn().mockResolvedValue(undefined),
   },
 }));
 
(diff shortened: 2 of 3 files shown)


tphakala:
@coderabbitai review

PR #1508, opened 2025-11-18
Add url encoding to user given city location

oskari:
Sometimes you just might live somewhere with eg. umlauts in the name :)  
"Piikkiö, Finland" resulted in `jq: parse error: Invalid numeric literal at line 1, column 8` because the curl would result in `Invalid HTTP request received` when not using url encoded strings

## Summary by CodeRabbit

* **Bug Fixes**
  * Setup configuration now URI-encodes city and country inputs before geocoding lookups, preventing malformed location queries.
  * Improves handling of locations with special characters, reducing failures or misrouting in external map/geocoding requests and making initial configuration more reliable.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3209,8 +3209,10 @@ configure_location() {
                     fi
                     
                     # Split input into city and country
-                    city=$(echo "$location" | cut -d',' -f1 | xargs)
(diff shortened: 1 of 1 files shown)


tphakala:
Thanks for contributing!

PR #1687, opened 2025-12-30
fix(ui): prevent comment section from collapsing when typing

### LINKED ISSUE #1683: Unable to add note/comment for a detection
## Issue Type

- [x] Bug report
- [ ] Feature request
- [ ] Question / Support

## Description

UI to add note/comment to detection does not work

## Steps to Reproduce (for bugs)

1. In the "Review" tab for a detection, I expand the "Add Comment" section

Image

2. I click on the text box and attempt to type a comment. The first keystroke collapses the comment section, so it's impossible to add a comment

Image

## Expected behavior

I expect to be able to type a comment

### Version Information (build date)

Provide the following line from your system:

nightly-20251223

tphakala:
## Summary

Fixes #1683

When users tried to add a comment in the Review tab, the comment section would collapse on the first keystroke, making it impossible to add comments.

**Root cause:** In Svelte 5, `$effect` creates reactive dependencies on any `$state` variables that are READ during execution. The line `showCommentSection = !!comment` was reading the `comment` $state variable, creating a dependency. When the user typed (changing `comment`), the effect would re-run, resetting `comment` back to the original value from `detection.comments` and collapsing the section.

**Fix:** Use the local `firstComment` variable instead of the `comment` $state when setting `showCommentSection`. This avoids creating the reactive dependency while preserving the same behavior.

## Changes

- **ReviewCard.svelte**: Changed `showCommentSection = !!comment` to `showCommentSection = !!firstComment`
- **ReviewModal.svelte**: Same fix applied
- **Added tests**: `ReviewCard.test.ts` with tests that verify the comment section stays open when typing

## Test plan

- [x] Added regression test that fails before fix, passes after
- [x] Verified with `npm test -- --run src/lib/desktop/components/review/ReviewCard.test.ts` - 2 tests pass
- [x] Lint/format check passes

---

Thanks to @mikeyk730 for the clear bug report with screenshots!

## Release Notes

* **Bug Fixes**
  * Fixed an issue where the comment section in the review modal would unintentionally reset or collapse while users were typing.

* **Tests**
  * Added comprehensive test coverage for comment persistence behavior during user input and empty detection scenarios.

### DIFF
--- a/frontend/src/lib/desktop/components/modals/ReviewModal.svelte
+++ b/frontend/src/lib/desktop/components/modals/ReviewModal.svelte
@@ -43,9 +43,11 @@
       lockDetection = false;
       ignoreSpecies = isExcluded;
       const firstComment = safeArrayAccess(detection.comments || [], 0);
-      comment = firstComment?.entry || '';
-      // Show comment section if there's already a comment
-      showCommentSection = !!comment;
+      const firstCommentValue = firstComment?.entry || '';
+      comment = firstCommentValue;
+      // Use firstCommentValue (local variable) instead of comment ($state) to avoid
+      // creating a reactive dependency that would reset the section when typing
+      showCommentSection = !!firstCommentValue;
     }
   });
 
--- a/frontend/src/lib/desktop/components/review/ReviewCard.svelte
+++ b/frontend/src/lib/desktop/components/review/ReviewCard.svelte
@@ -46,7 +46,9 @@
           ? detection.comments[0]?.entry || ''
           : '';
       comment = firstComment;
-      showCommentSection = !!comment;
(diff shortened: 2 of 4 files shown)


tphakala:
**Re: Gemini's suggestion to remove `as any` casts**

The `as any` cast is an established pattern in this codebase for Svelte 5 component tests. There are 62 occurrences across 6 test files (CollapsibleSection, WeatherInfo, StatsCard, DataTable, Pagination, ReviewCard). 

This is a necessary workaround for the type incompatibility between @testing-library/svelte (designed for Svelte 4) and Svelte 5's `$props()` pattern. Each usage is intentionally documented with an eslint-disable comment.

PR #1716, opened 2026-01-06
fix: normalize species config keys for case-insensitive matching (#1701)

### LINKED ISSUE #1701: Species Configuration Inconsistency: Include/Exclude vs Custom Config
## Issue Type

- [x] Bug report

## Summary

BirdNET-Go has three ways to configure species behavior (include list, exclude list, and custom config with thresholds), but they use **inconsistent matching logic**. This causes silent failures where users configure thresholds that never get applied because the matching rules are different.

## The Three Configuration Sections

```yaml
realtime:
  species:
    include:          # Always include these species (bypass range filter)
      - "American Robin"
    exclude:          # Always exclude these species
      - "House Sparrow"
    config:           # Custom thresholds and actions per species
      "american robin":
        threshold: 0.75
        interval: 30
```

### Include/Exclude Lists (Case-Insensitive, Flexible)

**Code Location:** `internal/birdnet/range_filter.go:197-200`

```go
func matchesSpecies(label, speciesName string) bool {
    scientificName, commonName, _ := observation.ParseSpeciesString(label)
    return strings.EqualFold(scientificName, speciesName) || 
           strings.EqualFold(commonName, speciesName)
}
```

**Matching Logic:**
- ✅ **Case-insensitive** (`strings.EqualFold`)
- ✅ **Common names** work: `"American Robin"`
- ✅ **Scientific names** work: `"Turdus migratorius"`
- ✅ **Any capitalization** works: `"american robin"`, `"AMERICAN ROBIN"`

**Examples that work:**
```yaml
include:
  - "American Robin"      # ✅ Works
  - "american robin"      # ✅ Works
  - "AMERICAN ROBIN"      # ✅ Works
  - "Turdus migratorius"  # ✅ Works (scientific name)
  - "turdus migratorius"  # ✅ Works
```

### Custom Config (Case-Sensitive, Common Name Only)

**Code Location:** `internal/analysis/processor/processor.go:770`

```go
func (p *Processor) getBaseConfidenceThreshold(speciesLowercase string) float32 {
    if config, exists := p.Settings.Realtime.Species.Config[speciesLowercase]; exists {
        return float32(config.Threshold)
    }
    return float32(p.Settings.BirdNET.Threshold)
}
```

**Matching Logic:**
- ❌ **Case-sensitive** (direct Go map key lookup)
- ✅ **Common names** work (when exact lowercase)
- ❌ **Scientific names** don't work
- ❌ **Only exact lowercase** works

**What works and doesn't work:**
```yaml
config:
  "american robin":       # ✅ Works (lowercase common name)
    threshold: 0.75
  "American Robin":       # ❌ Fails silently (capitalized)
    threshold: 0.75
  "AMERICAN ROBIN":       # ❌ Fails silently (uppercase)
    threshold: 0.75
  "Turdus migratorius":   # ❌ Fails silently (scientific name)
    threshold: 0.75
  "turdus migratorius":   # ❌ Fails silently (scientific lowercase)
    threshold: 0.75
```

### Inconsistency Table

| Feature | Include/Exclude Lists | Custom Config (Thresholds) |
|---------|----------------------|---------------------------|
| **Case sensitivity** | ❌ Case-INSENSITIVE | ✅ Case-SENSITIVE |
| **Common names** | ✅ Works (any case) | ✅ Works (lowercase only) |
| **Scientific names** | ✅ Works (any case) | ❌ Doesn't work |
| **Capitalization** | ✅ Any capitalization | ❌ Must be exact lowercase |

### User Experience Issues

1. **Silent Failures**: Users configure thresholds with capitalized names (matching the UI autocomplete), but they never get applied. No error is shown.

2. **Documentation Gap**: The [Wiki documentation](https://github.com/tphakala/birdnet-go/wiki/BirdNET%E2%80%90Go-Guide#stage-2-confidence-threshold) shows examples but never explicitly states:
   - ⚠️ You MUST use lowercase common names in `config`
   - ⚠️ Scientific names do NOT work in `config`
   - ⚠️ Capitalization matters in `config` but not in `include`/`exclude`

3. **UI Misleads Users**: The admin UI species autocomplete provides capitalized common names (e.g., `"American Robin"`), which work for include/exclude but fail for custom config.

4. **Inconsistent Behavior**: Users expect uniform behavior across all three sections:
   ```yaml
   # This seems logical but doesn't work consistently:
   include:
     - "American Robin"    # ✅ Works
   exclude:
     - "House Sparrow"     # ✅ Works
   config:
     "American Robin":     # ❌ Doesn't work!
       threshold: 0.75
   ```

### Example Scenario

A user wants to:
1. Always include "American Robin" (bypass range filter)
2. Set a lower threshold of 0.65 for "American Robin"

They configure:
```yaml
realtime:
  species:
    include:
      - "American Robin"  # ✅ Works - species is included
    config:
      "American Robin":   # ❌ Fails silently - threshold NOT applied
        threshold: 0.65
```

**Result:** The species is included (because `include` works), but the global threshold is used instead of 0.65 (because `config` lookup fails). No error is logged at INFO level.

### Why It's Confusing

- The admin UI provides "American Robin" as an autocomplete option
- `include` accepts "American Robin" 
- `exclude` accepts "American Robin"
- But `config["American Robin"]` silently fails
- Users must discover through experimentation that they need `"american robin"` (lowercase)

## Current Workaround

Users must manually lowercase all species names in the `config` section:

```yaml
config:
  "american robin":      # Must be lowercase
    threshold: 0.75
  "european blackbird":  # Must be lowercase
    threshold: 0.70
```

But they can use natural capitalization in `include`/`exclude`:
```yaml
include:
  - "American Robin"     # Can be any case
  - "European Blackbird" # Can be any case
```

## Proposed Solution

Make custom config matching **consistent** with include/exclude:

1. **Case-insensitive matching** (use `strings.EqualFold` like include/exclude)
2. **Support both common and scientific names** (like include/exclude)
3. **Preserve display capitalization** for UI (store original, lookup normalized)

### Implementation Approach

**Option A: Normalize keys on config load**
```go
// When loading YAML config, normalize all config keys
for key, value := range rawSpeciesConfig {
    normalizedKey := strings.ToLower(key)
    settings.Realtime.Species.Config[normalizedKey] = value
}
```

**Pros:**
- Simple one-line fix
- Backward compatible (lowercase keys still work)
- Users can write configs in any case
- Normalization happens once (no performance impact)

**Cons:**
- Display shows lowercase after reload (user enters "American Robin", sees "american robin")
- New config via UI not normalized until restart

**Option B: Store display name with normalized lookup**
```go
type SpeciesConfig struct {
    DisplayName string  // "American Robin" - for UI
    Threshold   float64
    Interval    int
    Actions     []SpeciesAction
}

// Map uses lowercase keys for lookup
config["american robin"] = SpeciesConfig{
    DisplayName: "American Robin",
    Threshold: 0.75,
}
```

We can add a second (fallback) O(1) lookup on scientific name for consistency with include/exclude list handling.
Additionally, we need to perform normalization when a new config setting is created via API write (otherwise it wouldn't be normalized until restart).

**Pros:**
- Case-insensitive lookup
- Preserves display capitalization
- Best user experience
- Backward compatible (old configs work, new field ignored by old code)

**Cons:**
- More complex implementation (requires struct field addition)
- Slightly larger config files (one additional field per species)

**Option C: Match both common and scientific names (like include/exclude)**

Reuse the existing `matchesSpecies` logic for config lookups:
```go
func (p *Processor) getBaseConfidenceThreshold(speciesLowercase string) float32 {
    for configKey, config := range p.Settings.Realtime.Species.Config {
        if matchesSpecies(speciesLowercase, configKey) {
            return float32(config.Threshold)
        }
    }
    return float32(p.Settings.BirdNET.Threshold)
}
```

**Pros:**
- Consistent with include/exclude behavior
- Supports both common and scientific names

**Cons:**
- O(n) lookup instead of O(1) map lookup
- Performance impact on every detection

tphakala:
## Summary
- Add `NormalizeSpeciesConfigKeys()` function to convert species config map keys to lowercase
- Apply normalization on config load and after API settings updates
- Ensures users can enter species names in any case (e.g., "American Robin", "AMERICAN ROBIN") and they resolve correctly

## Problem
Species include/exclude lists use case-insensitive matching via `matchesSpecies()`, but the species config map uses direct case-sensitive lookup. This causes silent failures when users configure thresholds with capitalized names like "American Robin" - the config is stored but never matched during detection lookup.

Fixes #1701

## Implementation
- **New file**: `internal/conf/species_config.go` - contains `NormalizeSpeciesConfigKeys()` with two-pass algorithm ensuring mixed-case keys (from updates) overwrite lowercase keys (from existing config)
- **Config load**: Normalize after `viper.Unmarshal()` in `Load()`
- **API update**: Normalize after `updateSettingsSectionWithTracking()` for realtime section

## Test plan
- [x] Unit tests for normalization function (empty, lowercase, mixed-case, uppercase, duplicates, nil)
- [x] Integration test simulating viper config load
- [x] API test for normalization on PATCH /api/v2/settings/realtime
- [x] All existing tests updated to use lowercase keys in assertions
- [x] Race detector: `go test -race ./internal/conf/... ./internal/api/v2/...`
- [x] Lint: `golangci-lint run -v`

## Summary by CodeRabbit

* **Bug Fixes**
  * Species configuration keys are now normalized to lowercase during config load and API updates, making lookups case-insensitive and preventing mismatches.

* **Tests**
  * Added and updated tests to verify lowercase normalization, preserve existing values, and ensure concurrency-safe test setup.

### DIFF
--- a/internal/api/v2/settings.go
+++ b/internal/api/v2/settings.go
@@ -524,12 +524,6 @@ func (c *Controller) UpdateSectionSettings(ctx echo.Context) error {
 		return c.HandleError(ctx, err, fmt.Sprintf("Failed to update %s settings", section), http.StatusBadRequest)
 	}
 
-	// Normalize species config keys to lowercase after realtime or species section updates
-	// This ensures case-insensitive matching regardless of how users enter species names
-	if (section == "realtime" || section == SettingsSectionSpecies) && settings.Realtime.Species.Config != nil {
-		settings.Realtime.Species.Config = conf.NormalizeSpeciesConfigKeys(settings.Realtime.Species.Config)
-	}
-
 	if err := c.handleSettingsChanges(&oldSettings, settings); err != nil {
 		*settings = oldSettings
 		return c.HandleError(ctx, err, "Failed to apply settings changes, rolled back to previous settings", http.StatusInternalServerError)
--- a/internal/conf/species_config_normalization_test.go
+++ b/internal/conf/species_config_normalization_test.go
@@ -151,7 +151,10 @@ realtime:
 	err = v.Unmarshal(&settings)
 	require.NoError(t, err)
 
-	// Apply normalization (this is what we're testing gets called)
+	// Apply normalization as called in Load()
+	// Note: Viper may already normalize map keys to lowercase during unmarshal,
+	// making this idempotent. We still call it to ensure consistent behavior
+	// regardless of viper's internal implementation.
 	settings.Realtime.Species.Config = NormalizeSpeciesConfigKeys(settings.Realtime.Species.Config)
 
 	// Verify keys are normalized to lowercase


tphakala:
## Code Review Feedback Addressed ✅

Fixed the issues identified by CodeRabbit and Gemini:

1. **PUT /api/v2/settings** - Added normalization after `updateAllowedSettingsWithTracking`
2. **PATCH /api/v2/settings/species** - Extended condition to include species section
3. **Pre-merge normalization** - Added `normalizeSpeciesConfigKeysInJSON()` to normalize incoming JSON keys BEFORE the deep merge. This was critical - without it, PATCH operations with mixed-case keys would lose existing config values (interval, actions) because the deep merge couldn't match keys with different casing.

Commit: ce9a09c8

PR #1736, opened 2026-01-09
fix(securefs): use internal logger instead of undefined log in Windows FIFO

tphakala:
## Summary
- Replace undefined `log` calls with `GetLogger()` in `fifo_windows.go`
- Fixes Windows build failure caused by using nonexistent `log` identifier
- The file imported `logger` package but incorrectly called `log.Warn`/`log.Debug` instead of `GetLogger().Warn`/`GetLogger().Debug`

## Test plan
- [ ] Verify Windows build succeeds: `task windows_amd64`

## Summary by CodeRabbit

* **Chores**
  * Improved logging infrastructure consistency across file system operations to enhance code maintainability and provide better troubleshooting support. Internal logging systems were standardized to apply uniform patterns and context handling, enabling more reliable error tracking, enhanced system monitoring capabilities, and supporting improved development efficiency and overall platform stability.

### DIFF
--- a/internal/securefs/fifo_windows.go
+++ b/internal/securefs/fifo_windows.go
@@ -46,7 +46,7 @@ func createFIFOPlatform(path string) (string, error) {
 	// Remove any existing file at the path location
 	if _, err := os.Stat(path); err == nil {
 		if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
-			log.Warn("Error removing existing file at FIFO path",
+			GetLogger().Warn("Error removing existing file at FIFO path",
 				logger.String("path", path),
 				logger.Error(err))
 		}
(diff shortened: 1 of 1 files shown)


tphakala:
## Response to Review Comments

**Re: Caching GetLogger()**

The pattern of calling `GetLogger()` each time is intentional and consistent across all securefs files (18 calls, none cached). As documented in [securefs.go:19-21](internal/securefs/securefs.go#L19-L21):

> The logger is fetched from the global logger each time to ensure it uses the current centralized logger (which may be set after package init).

This function is called during pipe creation (rare), not a hot path. Caching would break consistency with the rest of the package.

**Re: Returning errors for placeholder failures**

The warn-and-continue behavior was in the original code - this PR only fixes the compilation error (undefined `log` → `GetLogger()`), preserving existing semantics.

The placeholder is purely informational metadata for humans to track what a Windows pipe maps to. The function returns `fullPipeName` regardless - nothing depends on the placeholder file existing. Changing error handling would be a behavior change beyond this bug fix's scope.

PR #1740, opened 2026-01-10
fix(weather): store timestamps in local timezone for correct date queries

tphakala:
## Summary
- Fix UTC/local timezone mismatch when storing weather timestamps
- Weather providers return UTC, but date queries use local dates
- Early morning hours were stored with wrong date, making them unfindable

## Changes
- Convert UTC to local time before storing in `SaveWeatherData`
- Add timezone offset to log timestamps for debugging clarity
- Ensure API consumers use local time when formatting responses
- Code cleanup: remove duplicate error logging, add constant for absolute zero

## Test plan
- [ ] Verify new weather records show local timezone offset in database
- [ ] Verify hourly weather queries return correct data for local dates
- [ ] Check logs show timestamps with timezone (e.g., `2026-01-10 11:00:00+02:00`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Weather timestamps now reflect the user's local timezone.
  * Temperature validation tightened to prevent invalid values.

* **Improvements**
  * More consistent storage of daily and hourly weather data using local time.
  * Cleaner, more reliable data fetching and polling with clearer logging and timestamped operations.

### DIFF
--- a/internal/weather/weather.go
+++ b/internal/weather/weather.go
@@ -284,14 +284,13 @@ func (s *Service) fetchAndSave() error {
 			Build()
 	}
 
-	// Convert to local time once for logging and storage.
-	// Mutating data.Time is safe since data is only used within this function.
-	// SaveWeatherData's In(time.Local) call becomes a no-op.
-	data.Time = data.Time.In(time.Local)
+	// Convert to local time for logging. SaveWeatherData handles its own
+	// timezone conversion for storage.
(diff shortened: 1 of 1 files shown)


tphakala:
Addressed review feedback in 41c97c67: Using local variable for logging instead of mutating input parameter. SaveWeatherData handles its own timezone conversion.

PR #1760, opened 2026-01-12
fix(processor): allow ExecuteCommand actions without parameters

tphakala:
## Summary

- Fixes bug where `ExecuteCommand` actions would not execute when no parameters were specified
- Adds dedicated log file for action processing (`logs/actions.log`)
- Adds nil guard to prevent potential panic in `logDetectionResults`

Fixes: https://github.com/tphakala/birdnet-go/discussions/1757

## Root Cause

The condition `if len(actionConfig.Parameters) > 0` at processor.go:1116 prevented commands without parameters from being added to the action list. Commands like `/usr/local/bin/notify.sh` that don't need any parameters would silently fail.

## Changes

1. **processor.go**: Removed the parameter length check so commands execute regardless of parameters
2. **logger/config.go**: Added `analysis.processor` module with dedicated log file at `logs/actions.log`
3. **processor.go**: Added nil guard for `logDedup` to prevent panic in tests or partial initialization

## Test plan

- [x] Added `execute_command_test.go` with tests for parameterless commands
- [x] Added `execute_internal_test.go` with comprehensive unit tests for ExecuteCommandAction
- [x] All tests pass with `go test -race -v ./internal/analysis/processor/...`
- [x] Linter passes with `golangci-lint run -v`

## Summary by CodeRabbit

* **Bug Fixes**
  * ExecuteCommand actions now run even when no parameters are defined; logging guard prevents nil errors during action logging.

* **Improvements**
  * Dedicated per-module action processor log file added for clearer visibility.
  * Confidence values for command parameters are now normalized for consistent handling.

* **Tests**
  * Large new test suites validating command execution, parameter handling, sanitization, security, and edge cases.

### DIFF
--- a/internal/analysis/processor/execute.go
+++ b/internal/analysis/processor/execute.go
@@ -266,6 +266,9 @@ func buildSafeArguments(params map[string]any, note *datastore.Note) ([]string, 
 
 // isValidParamName checks if a parameter name contains only safe characters
 func isValidParamName(name string) bool {
+	if name == "" {
+		return false
+	}
 	for _, r := range name {
 		if (r < 'a' || r > 'z') && (r < 'A' || r > 'Z') &&
 			(r < '0' || r > '9') && r != '_' && r != '-' {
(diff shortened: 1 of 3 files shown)


tphakala:
Addressed review feedback:

- **gemini-code-assist**: Fixed confidence normalization. Changed `param == "confidence"` to `param == "Confidence"` to match struct field. Test now expects 95.0.

- **coderabbitai**: Fixed empty param validation. Added explicit empty string check at start of `isValidParamName`. Test now asserts `false`.

PR #1886, opened 2026-01-27
fix(v2only): implement SetSunCalcMetrics for observability

### LINKED ISSUE #1884: v2only: Implement SetSunCalcMetrics for observability
## Parent Issue
Part of #1862 (Database Normalization)

## Problem
The `SetSunCalcMetrics` method in `internal/datastore/v2only/datastore.go:137` is an empty stub:

```go
func (ds *Datastore) SetSunCalcMetrics(_ any) {}
```

This method ignores its argument entirely, resulting in missing SunCalc telemetry/observability in v2-only mode compared to the legacy implementation.

## Expected Behavior
The method should store and utilize the metrics instance for SunCalc observability, similar to how `SetMetrics` works for general datastore metrics.

## Files Affected
- `internal/datastore/v2only/datastore.go`

## Severity
Low - Missing observability, not core logic failure

tphakala:
## Summary

- Implements the `SetSunCalcMetrics` method in the v2only datastore
- Method was an empty stub that ignored its argument, resulting in missing SunCalc observability metrics when running in v2-only mode
- Now properly forwards metrics to the SunCalc service, matching legacy datastore behavior

## Changes

- Added import with alias `obmetrics` to avoid shadowing the `metrics` parameter in `SetMetrics` method
- Implemented `SetSunCalcMetrics` to type-assert and delegate to `suncalc.SetMetrics()`

## Test Plan

- [x] Build passes
- [x] Lint passes (0 issues)
- [x] All v2only tests pass
- [x] Thread safety tests pass
- [x] Gemini review approved

Fixes #1884

## Summary by CodeRabbit

* **Chores**
  * Enhanced internal observability infrastructure to improve metrics collection and monitoring capabilities.

### DIFF
--- a/internal/datastore/v2only/datastore.go
+++ b/internal/datastore/v2only/datastore.go
@@ -16,6 +16,7 @@ import (
 	"github.com/tphakala/birdnet-go/internal/detection"
 	"github.com/tphakala/birdnet-go/internal/errors"
 	"github.com/tphakala/birdnet-go/internal/logger"
+	obmetrics "github.com/tphakala/birdnet-go/internal/observability/metrics"
 	"github.com/tphakala/birdnet-go/internal/suncalc"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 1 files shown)


tphakala:
Re: [Gemini's suggestion to remove nil check](https://github.com/tphakala/birdnet-go/pull/1886#discussion_r2733348742)

The `suncalcMetrics != nil` check is intentional. Typed nils pass type assertion with `ok=true`:

```go
var x any = nil
m, ok := x.(*MyMetrics)  // ok=false ✓

var y any = (*MyMetrics)(nil)  // typed nil
m, ok := y.(*MyMetrics)  // ok=true, m=nil ✗
```

Without the explicit check, a typed nil would pass through and call `SetMetrics(nil)`, potentially clearing previously set metrics.

This also matches the legacy implementation pattern at `internal/datastore/interfaces.go:263`.

PR #1949, opened 2026-02-13
fix(WeatherBar) weather bar showing future hours due to timezone offset mismatch

ModerateWinGuy:
## Summary
- **Root cause**: `mattn/go-sqlite3` preserves timezone offsets when storing `time.Time` values (e.g., `"2024-01-15 11:00:00+13:00"`). SQLite compares these as strings, so queries using UTC boundaries (`+00:00`) against locally-stored records (`+13:00`) produced incorrect results — yesterday's afternoon/evening records leaked into today's weather bar as "future" hours.
- **Storage fix**: Weather timestamps now stored in UTC for consistent string comparisons going forward
- **Query fix**: `strftime('%s')` used in all weather queries to compare as Unix epoch seconds, correctly handling both legacy records with local offsets and new UTC records without requiring a data migration
- **DST fix**: v2 repository `GetHourlyWeatherInLocation` now uses `AddDate(0, 0, 1)` instead of `Add(24*time.Hour)` for correct 23/25-hour DST day boundaries

## Test plan
- [x] Added 5 new tests in `TestGetHourlyWeather_NonUTCStorageOffset` that store records with non-UTC timezone offsets (+13:00, -04:00) and verify correct query behavior
- [x] Verified the "yesterday leak" test **fails** without the fix and **passes** with it
- [x] Verified all 22 existing weather timezone tests still pass
- [x] Verified weather service tests pass
- [x] Verified API v2 weather tests pass
- [x] Ran linter on changed packages — no new issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Store timestamps in UTC and compare by epoch seconds to handle legacy local-offset records, DST/day-boundary edge cases, negative offsets, and mixed UTC/local storage; fixed day-boundary calculations to prevent cross-day leakage.

* **Tests**
  * Added comprehensive tests covering non-UTC storage offsets, DST boundaries, exact inclusions/exclusions, and mixed UTC/local records.

### DIFF
--- a/internal/datastore/interfaces.go
+++ b/internal/datastore/interfaces.go
@@ -865,7 +865,15 @@ func (ds *DataStore) SaveHourlyWeather(hourlyWeather *HourlyWeather) error {
 	// Use upsert to avoid duplicates for the same timestamp.
 	// Compare using Unix epoch seconds to handle legacy records with local
 	// timezone offsets alongside new UTC records.
-	result := ds.DB.Where("strftime('%s', time) = strftime('%s', ?)", hourlyWeather.Time).
+	var whereClause string
+	switch strings.ToLower(ds.Dialector().Name()) {
+	case DialectMySQL:
+		whereClause = "UNIX_TIMESTAMP(time) = UNIX_TIMESTAMP(?)"
+	default: // SQLite
+		whereClause = "strftime('%s', time) = strftime('%s', ?)"
+	}
+
+	result := ds.DB.Where(whereClause, hourlyWeather.Time).
 		Assign(*hourlyWeather).
 		FirstOrCreate(hourlyWeather)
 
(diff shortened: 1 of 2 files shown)


tphakala:
Thanks @ModerateWinGuy for this excellent fix! 🎉

The root cause analysis was spot-on - SQLite's string comparison with mixed timezone offsets was indeed causing the weather bar to show incorrect hours. Your solution using epoch-based timestamp comparisons is elegant and handles both legacy records (with local timezone offsets) and new UTC records seamlessly.

The comprehensive test coverage you added is particularly impressive:
- Timezone boundary cases (GMT+13, EST, EDT)
- DST transitions (23-hour and 25-hour days)
- Mixed storage formats (legacy local offsets + new UTC)
- Edge cases (leap years, year boundaries, midnight)

This is exactly the kind of thorough fix that prevents future regressions. The dialect-aware approach also ensures compatibility with both SQLite and MySQL deployments.

Great work and thanks for contributing to BirdNET-Go! 🐦

PR #2139, opened 2026-03-09
fix(imageprovider): guard against empty scientific names and add datastore schema tag

tphakala:
## Summary

Fixes two production issues identified via Sentry:

- **Empty scientific name validation errors** — Image cache entries with orphaned v2 label relationships return empty `ScientificName` from `GetAllImageCaches`, causing `SaveImageCache` to reject them with `[validation] invalid scientific_name: scientific name cannot be empty` during background cache refresh
- **Missing datastore schema telemetry** — No way to distinguish legacy vs v2 schema in Sentry events, making triage harder

### Fix 1: Guard against empty scientific names
- **`findStaleEntries`** — Skip entries with empty `ScientificName` with a warning log, preventing unnecessary refresh attempts on orphaned data
- **`saveToDB`** — Early return when `ScientificName` is empty, providing a catch-all guard for any code path

### Fix 2: Datastore schema Sentry tag
- Add `SchemaVersion() string` to `datastore.Interface` returning `"legacy"` or `"v2"`
- Add `telemetry.SetDatastoreSchemaTag()` to set a global `datastore_schema` Sentry tag
- Call it after datastore initialization in `realtime.go`

## Test plan
- [x] `golangci-lint run -v` — 0 issues
- [x] `go test -race ./internal/imageprovider/...` — all pass
- [x] `go test -race ./internal/datastore/v2only/...` — all pass
- [x] `go test -race ./internal/telemetry/...` — all pass
- [x] `go test -race ./internal/analysis/processor/...` — all pass
- [x] Pre-commit hooks pass
- [x] Gemini code review — no blocking issues

## Follow-up considerations
- Orphaned image cache entries with empty scientific names will be skipped but not cleaned up (potential future migration/cleanup task)

## Summary by CodeRabbit

* **New Features**
  * Datastore now exposes schema version information for better visibility.
  * Telemetry automatically tags the active datastore schema version for improved observability.

* **Bug Fixes**
  * Image provider skips and warns on entries missing a scientific name, preventing saves and stale-entry processing.

### DIFF
--- a/internal/analysis/processor/mock_datastore_test.go
+++ b/internal/analysis/processor/mock_datastore_test.go
@@ -349,7 +349,7 @@ func (m *ActionMockDatastore) GetActiveNotificationHistory(_ time.Time) ([]datas
 func (m *ActionMockDatastore) DeleteExpiredNotificationHistory(_ time.Time) (int64, error) {
 	return 0, nil
 }
-func (m *ActionMockDatastore) SchemaVersion() string { return "legacy" }
+func (m *ActionMockDatastore) SchemaVersion() string { return datastore.SchemaVersionLegacy }
 func (m *ActionMockDatastore) GetDatabaseStats() (*datastore.DatabaseStats, error) {
 	return &datastore.DatabaseStats{Type: "mock", Connected: true}, nil
 }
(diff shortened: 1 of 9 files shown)


tphakala:
Addressed CodeRabbit nitpicks in f08613ac:

1. **Schema version constants** — Added `SchemaVersionLegacy` and `SchemaVersionV2` constants in `internal/datastore`, replaced all magic strings across implementations and test stubs
2. **Telemetry guard** — Added `shouldSkipTelemetry()` + empty string guard to `SetDatastoreSchemaTag`
3. **Regression test for saveToDB** — Skipping for now. `saveToDB` is private and the upstream `findStaleEntries` guard prevents the path from being hit. Defense-in-depth catch-all, not the primary fix

PR #2167, opened 2026-03-10
fix(processor): remove redundant SSE audio file polling

### LINKED ISSUE #2151: SSE audio file wait timeout in analysis processor (BIRDNET-GO-QB)
## Summary

The SSE action's `waitForAudioFile()` is timing out after 5 seconds waiting for audio clip files to appear on disk. This is a new Sentry issue (BIRDNET-GO-QB, 8 events) first seen in the `nightly-20260309` release, running on `amd64` containers with datastore schema v2.

A pre-existing similar issue (BIRDNET-GO-B7, 77K events) has been tracked for longer with different Sentry fingerprinting, suggesting the underlying race condition between audio file writing and SSE broadcast has been a persistent problem.

## Error Details

- **Error**: `audio file <clip_path> not ready after 5s timeout`
- **Component**: `analysis.processor`
- **Category**: `timeout`
- **Sentry Issue**: BIRDNET-GO-QB (new fingerprint), related to BIRDNET-GO-B7 (77K events)

### The Race Condition

The `CompositeAction` executes `DatabaseAction` → `SSEAction` → `MqttAction` sequentially. Within `DatabaseAction.ExecuteContext()`, audio export happens synchronously — it reads PCM data from the capture buffer and writes it to disk via `SaveAudioAction`. After `DatabaseAction` completes, `SSEAction` runs and calls `waitForAudioFile()` which polls for the file on disk.

The timeout occurs when the audio file is not present or not large enough (> 1024 bytes) within 5 seconds. The possible causes are:

1. **Audio export failure without `AudioExportFailed` signal**: If `ReadSegmentFromCaptureBuffer()` returns data but `SaveAudioAction.Execute()` fails (e.g., disk full, permission error, slow I/O on SD cards), the `handleAudioExportError` function sets `DetectionContext.AudioExportFailed` to skip the wait. However, if the error occurs in a way that doesn't reach the error handler (e.g., the write succeeds but the file is still being flushed/synced by the OS), SSEAction will wait the full 5 seconds.

2. **Slow disk I/O on resource-constrained hardware**: On devices like Raspberry Pi with SD cards, the `SaveAudioAction` may complete (write call returns) but the OS may not have flushed the data to disk yet. The subsequent `os.Stat()` in `waitForAudioFile()` could see a file with `Size() <= 1024` bytes during the flush, causing the timeout. This is especially relevant for non-WAV exports using FFmpeg encoding which is CPU-intensive.

3. **Extended capture duration clips**: Extended capture mode (introduced in commit `c2921d57`) generates longer audio clips. These take longer to encode (especially with FFmpeg for non-WAV formats), which could push the total audio save time beyond the point where `SSEAction` times out.

4. **Capture buffer read failure**: If `ReadSegmentFromCaptureBuffer()` fails, the audio export is skipped entirely but `AudioExportFailed` is correctly set. However, partial failures or edge cases in buffer reads may not trigger the flag consistently.

### Why This is a New Sentry Fingerprint (QB vs B7)

The two different Sentry fingerprints (BIRDNET-GO-QB and BIRDNET-GO-B7) are caused by Sentry's server-side data scrubbing hashing different path structures differently. The clip name format is `YYYY/MM/species_name_CONFp_TIMESTAMP[_DURs].ext` (e.g., `2026/03/cardinalis_cardinalis_97p_20260309T120000Z.wav`). Different species names, confidence levels, and timestamps produce different hashed segments in Sentry, creating different fingerprints. The QB fingerprint likely represents a different species or confidence pattern being grouped separately by Sentry's data scrubbing rules.

## Affected Code Paths

- **Timeout error origin**: `internal/analysis/processor/actions_integrations.go` — `SSEAction.waitForAudioFile()` (line ~473)
- **Audio save**: `internal/analysis/processor/actions_database.go` — `SaveAudioAction.Execute()` (line ~385)
- **Audio save trigger**: `internal/analysis/processor/actions_database.go` — `DatabaseAction.ExecuteContext()` (line ~217)
- **Clip name generation**: `internal/analysis/processor/processor.go` — `buildClipPath()` (line ~1014)
- **Constants**: `internal/analysis/processor/actions_types.go` — `SSEAudioFileTimeout = 5s`, `MinAudioFileSize = 1024`

### Option A: Eliminate the polling-based wait (Recommended)

Instead of having `SSEAction` poll the filesystem for the audio file, use the existing `DetectionContext` to signal audio export completion. Add an `AudioExportDone` atomic flag (or channel) that `DatabaseAction` sets after `SaveAudioAction` succeeds. `SSEAction` would check this flag instead of polling the filesystem, eliminating the race entirely.

```go
// In DetectionContext:
AudioExportDone atomic.Bool  // Set true when audio file is confirmed written

// In DatabaseAction, after successful SaveAudioAction:
a.DetectionCtx.AudioExportDone.Store(true)

// In SSEAction.waitForAudioFile():
// Wait for signal instead of polling filesystem
```

### Option B: Increase timeout and add adaptive backoff

Increase `SSEAudioFileTimeout` from 5s to 10s for containers/slow hardware, and use exponential backoff instead of fixed 100ms polling interval. This is a simpler change but doesn't eliminate the fundamental race.

### Option C: Make SSE broadcast independent of audio file readiness

The SSE broadcast already proceeds even if the audio wait times out (the error is logged as a warning). Consider removing the audio file wait entirely from `SSEAction` — the frontend already handles audio file unavailability with retry logic (`PlayOverlay.svelte`, `AudioPlayer.svelte`). This would eliminate the 5-second delay and the Sentry noise.

## Impact

- **User-facing**: SSE broadcasts are delayed by up to 5 seconds when audio files aren't ready, causing delayed real-time dashboard updates
- **Telemetry noise**: 77K+ events for the related BIRDNET-GO-B7 issue suggest this is a high-frequency occurrence
- **No data loss**: Detections are saved to the database regardless; only the SSE real-time notification is delayed

tphakala:
## Summary

- Remove `SSEAction.waitForAudioFile()` filesystem polling that caused 77K+ Sentry events (BIRDNET-GO-B7, BIRDNET-GO-QB)
- The polling was always redundant: `DatabaseAction` and `SSEAction` run sequentially in `CompositeAction`, so audio export completes before SSE runs
- Remove `AudioExportFailed` signaling mechanism and related constants (`SSEAudioFileTimeout`, `SSEAudioCheckInterval`, `MinAudioFileSize`) that only existed to support the polling
- Net change: -328 lines, +6 lines — pure removal of dead code path

## Why this works

The `CompositeAction` executes `DatabaseAction` → `SSEAction` → `MqttAction` **sequentially**. Audio export happens synchronously within `DatabaseAction.ExecuteContext()` via `SaveAudioAction`. By the time `SSEAction.Execute()` is called, the audio file is already on disk (or the export failed). The filesystem polling was solving a problem that doesn't exist.

The frontend (`PlayOverlay.svelte`, `AudioPlayer.svelte`) already handles audio file unavailability with retry logic, covering any remaining edge cases.

## Test plan

- [x] All existing SSE, CompositeAction, and DetectionContext tests pass with `-race`
- [x] `golangci-lint` passes with zero issues
- [x] Gemini peer review confirms correctness

Fixes #2151

## Summary by CodeRabbit

* **New Features**
  * Added support for a small ("sm") spectrogram size (258px).

* **Behavior Changes**
  * Streaming/broadcasting no longer waits for audio file readiness; audio export errors are logged but do not delay or alter detection/broadcast flow.

### DIFF
--- a/internal/analysis/processor/sse_action_test.go
+++ b/internal/analysis/processor/sse_action_test.go
@@ -265,13 +265,10 @@ func TestSSEAction_Execute_NoClipNameBroadcastsQuickly(t *testing.T) {
 		SSEBroadcaster: mockBroadcaster.BroadcastFunc(),
 	}
 
-	startTime := time.Now()
 	err := action.Execute(t.Context(), nil)
-	duration := time.Since(startTime)
-
 	require.NoError(t, err)
 
-	// SSE broadcast should always be fast
-	assert.Less(t, duration, 500*time.Millisecond,
-		"SSE broadcast should complete quickly")
+	// Verify broadcast was called successfully
+	assert.Equal(t, 1, mockBroadcaster.GetBroadcastCount(), "Should broadcast once")
+	require.NotNil(t, mockBroadcaster.GetLastNote(), "Note should be broadcasted")
 }


tphakala:
CodeRabbit review addressed:

- **sse_action_test.go**: Fixed — replaced wall-clock timing assertion with behavioral check (broadcast count + note verification). Good catch.
- **media.go**: This file is not part of this PR's changes — it's from a different commit on main. The suggestion may be valid but is out of scope here.

PR #2426, opened 2026-03-20
feat/taxonomy synonyms in config.yaml

tariquesani:
This PR adds backend support for configurable taxonomy synonym overrides via a new top-level taxonomySynonyms section in config.yaml, while preserving built-in synonym mappings for backward compatibility. Lookup behavior remains bidirectional and case-insensitive, but now merges built-ins with config values so user-defined entries can override or extend defaults. Configuration schema and defaults were updated in conf, and tests were expanded to cover built-in completeness, override precedence, custom additions, and invalid/blank entry handling. This enables taxonomy updates without code changes.

## Release Notes

* **New Features**
  * Added optional configuration option to define custom taxonomy synonym mappings for scientific name overrides.
  * Implemented Windows-compatible secure filesystem operations with sandboxed path containment and validation.

* **Tests**
  * Added unit tests for taxonomy synonym configuration merging and override behavior.

### DIFF
--- a/internal/conf/config.go
+++ b/internal/conf/config.go
@@ -1415,6 +1415,8 @@ type Settings struct {
 
 	BirdNET BirdNETConfig `json:"birdnet"` // BirdNET configuration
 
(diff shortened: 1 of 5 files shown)


tariquesani:
Sorry that was a messed up branch

PR #2427, opened 2026-03-20
feat: support config-based taxonomy synonym overrides

tariquesani:
This PR adds backend support for configurable taxonomy synonym overrides via a new top-level taxonomySynonyms section in config.yaml, while preserving built-in synonym mappings for backward compatibility. Lookup behavior remains bidirectional and case-insensitive, but now merges built-ins with config values so user-defined entries can override or extend defaults. Configuration schema and defaults were updated in conf, and tests were expanded to cover built-in completeness, override precedence, custom additions, and invalid/blank entry handling. This enables taxonomy updates without code changes.

## Summary by CodeRabbit

* **New Features**
  * Added optional taxonomy synonym setting to configure custom scientific-name mappings (empty by default).
  * Custom synonym overrides let users replace or supplement built-in mappings for improved species lookup and display.
  * Overrides are applied when the application starts and whenever settings are updated, so changes take effect without code changes.

### DIFF
--- a/internal/api/v2/settings.go
+++ b/internal/api/v2/settings.go
@@ -219,6 +219,9 @@ func (c *Controller) UpdateSettings(ctx echo.Context) error {
 	// Update the cached telemetry state after settings change
 	telemetry.UpdateTelemetryEnabled()
 
+	// Rebuild taxonomy synonym lookup cache if overrides changed
+	imageprovider.SetCustomSynonyms(settings.TaxonomySynonyms)
+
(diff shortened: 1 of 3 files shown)


tphakala:
Thanks for the contribution! This is a good idea — users shouldn't need to submit PRs every time a species gets reclassified.

I'm going to polish the implementation a bit before merging. The main things I want to address:

- Cache the synonym lookup maps instead of rebuilding them on every call (performance)
- Fix the YAML template formatting so uncommenting examples actually works
- Avoid coupling the imageprovider package to the config singleton
- Clean up stale reverse-map entries when overrides replace built-ins

tariquesani:
Thank you for working on this PR and making it up to the mark

PR #2589, opened 2026-03-26
fix: use Playwright baseURL in alert-rules E2E availability probe

tphakala:
## Summary

The availability probe built its own URL from `BASE_URL`/`localhost:8080` instead of using Playwright's configured `baseURL`. In CI this could hit the wrong origin and convert real failures into skipped tests.

Changed to use Playwright's `request` fixture with a relative path. Connection errors now propagate as test failures instead of being silently masked. Only explicit 404/409 responses (feature genuinely unavailable) cause skips.

## Test plan

- [x] Prettier formatted
- [x] Probe now fails loudly on connection errors instead of skipping

## Summary by CodeRabbit

* **Tests**
  * Refined alert-schema availability probe in end-to-end alert-rules tests (removed extra error swallowing and warning logs; non-OK skip logic preserved).
  * Updated end-to-end alert tests and helpers to call alert APIs via relative endpoints, ensuring consistent request paths for CRUD/export checks.

### DIFF
--- a/frontend/tests/e2e/settings/alert-rules.spec.ts
+++ b/frontend/tests/e2e/settings/alert-rules.spec.ts
@@ -3,8 +3,6 @@ import { test, expect, type Page, type APIRequestContext } from '@playwright/tes
 test.describe('Alert Rules Settings Page', () => {
   test.setTimeout(30000);
 
-  const baseUrl = process.env['BASE_URL'] ?? 'http://localhost:8080';
-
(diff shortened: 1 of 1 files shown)


tphakala:
Fixed in 440f3c266. Removed the `baseUrl` variable entirely and switched all API calls in the file to relative paths so everything goes through Playwright's configured `baseURL`.

PR #2675, opened 2026-04-04
fix: clean up model display names and filter by enabled config

tphakala:
## Summary
- Clean up model display names in audio settings dropdown: remove verbose "GLOBAL"/"6K" prefixes, show backend type (TFLite/ONNX)
- Filter model list API to only return models that are enabled in configuration
- Add named constants for model names and backend types as single source of truth

**Before:** Dropdown showed all models with names like "BirdNET GLOBAL 6K V2.4" regardless of config  
**After:** Only enabled models shown, with clean names like "BirdNET v2.4 (TFLite)"

## Changes
- `internal/classifier/model_registry.go`: Add `Backend` field, `DisplayName()` method, name/backend constants
- `internal/api/v2/models.go`: Filter by `Settings.Models.Enabled`, use `DisplayName()` 
- Tests and references updated to use constants

## Test Plan
- [x] All classifier tests pass (`go test -race -v ./internal/classifier/`)
- [x] Buffer manager tests pass (`go test -race -v ./internal/analysis/`)
- [x] golangci-lint clean (zero issues)
- [ ] Verify dropdown on running instance shows only enabled models with clean names

## Summary by CodeRabbit

* **New Features**
  * Model list endpoint now respects configuration-driven enablement, showing only enabled models.
  * Model names shown to users now include backend info when applicable (e.g., "BirdNET v2.4 (TFLite)").

* **Refactor**
  * Standardized and centralized user-facing model naming for consistent display.

* **Tests**
  * Added and updated tests to validate model metadata, backend labeling, and display-name behavior.

### DIFF
--- a/internal/api/v2/models.go
+++ b/internal/api/v2/models.go
@@ -37,6 +37,7 @@ func (c *Controller) ListModels(ctx echo.Context) error {
 					ID:   alias,
 					Name: info.DisplayName(),
 				})
+				break // one entry per model
 			}
 		}
 	}


tphakala:
Fixed the duplicate-alias nitpick from CodeRabbit in 3d4a3a41 — added `break` after first matching alias per model.

PR #2733, opened 2026-04-11
fix(frontend): show "No audio sources" instead of infinite "Loading"

rexxars:
Minor: when no audio sources are configured, the "live audio" page shows "Loading……" indefinitely in the audio source dropdown. This changes it to say "No audio sources" instead if that is the case.

## Summary by CodeRabbit

* **Improvements**
  * Audio source dropdown now reflects discovery completion: shows "Loading..." while scanning and switches to "No Audio Sources" once initial audio-level data confirms no sources.
  * Start button empty-state displays a spinner during discovery and, once discovery completes, switches to a microphone icon and a "No Audio Sources" message.


tphakala:
Thanks for the fix! Merging.

PR #2740, opened 2026-04-12
feat: Implement Species Guide provider with caching and API endpoints

iamrans0m00:
### Overview
This branch adds a new Species Guide Provider feature that enriches bird detections with contextual guide information from Wikipedia and eBird. It includes intelligent caching, fallback handling, and frontend UI support for species guides, similar-species comparison, and user notes.

### What it does
- Adds guideprovider package for:
  - Wikipedia guide fetching
  - eBird provider support
  - two-tier caching: in-memory + persistent database
  - stale-while-revalidate refresh behavior
  - negative caching for not-found species
  - fallback policy support
- Adds API v2 endpoints for:
  - `/api/v2/species/:scientific_name/guide`
  - `/api/v2/species/:scientific_name/similar`
  - species notes CRUD
- Adds frontend settings and UI components for:
  - enabling/disabling species guide display
  - choosing provider and fallback behavior
  - toggling enrichments, similar species comparison, and notes
- Adds operational documentation and rollout guidance

### Key implementation details
- Feature is disabled by default
- Config keys use lowercase YAML naming:
  - `speciesguide.enabled`
  - `speciesguide.provider`
  - `speciesguide.fallbackpolicy`
  - `speciesguide.warmtopn`
  - `speciesguide.prefetchenabled`
  - `speciesguide.shownotes`
  - `speciesguide.showenrichments`
  - `speciesguide.showsimilarspecies`
- Wikipedia access is throttled to 1 request/sec per instance by design
- The API README and rollout docs have been updated to match implementation

### Testing
- Unit and integration tests added for guide provider, caching behavior, API endpoints, and error cases
- Documentation updated to reflect current config and API behavior

### Why this is useful
This provides users with richer species context directly in the dashboard, improves detection UX, and allows gradual rollout with safe defaults and monitoring.

## Summary by CodeRabbit

* **New Features**
  * Species Guide UI: Wikipedia (optional eBird) guides with locale-aware sections, season highlighting, similar-species side-by-side comparison, per-species notes (create/edit/delete), and an in-app Species Comparison component and enlarged detail modal; detection rows/cards now open species details.
* **Bug Fixes**
  * Various race-condition and frontend test/TypeScript fixes related to species guide and comparison.
* **Documentation**
  * Rollout guide, operations runbook, API docs, and developer docs added/updated.
* **Chores**
  * Build/run, ignore rules, deps, metrics, tests, and tooling updates.


iamrans0m00:
@coderabbitai resume

iamrans0m00:
@coderabbitai full review

tphakala:
Thanks for putting this together, this is a substantial piece of work and the overall architecture is really solid. The two-tier cache with singleflight deduplication, circuit breaker on Wikipedia, negative caching, and stale-while-revalidate refresh are all well thought out. I can tell a lot of design effort went into this.

That said, after doing a thorough review (including running the backend and frontend through separate deep reviews), I've found enough issues that I don't think this is mergeable in its current state. Let me walk through the main concerns so you have a clear picture.

On the backend side, the most pressing issue is in `PreFetch` in `guideprovider.go` around line 285. The goroutine captures the caller's request context, so when the HTTP request completes the background fetch gets cancelled. It should derive from `c.rootCtx` instead, the same way `Get()` already does. The eBird provider also downloads the entire taxonomy (~16K entries) on every `Fetch()` call and then does a linear scan to find one species. Even with the client-side cache this is wasteful; it should index by scientific name after the first fetch, or use a species-specific endpoint. The `sync.Map` in-memory cache has no eviction or size limit. With 16K species across 14 locales, each entry potentially holding 10KB of text, this could grow to ~2GB in theory. Some form of LRU or max-size cap would be good. There's also no locale validation at the API boundary in `species.go` around line 939; arbitrary strings get stored in the DB and used as map keys. A simple format check (2-3 lowercase letters) would close that off.

On the frontend, the biggest concern is the code duplication between `SpeciesDetailModal.svelte` and `DetectionDetail.svelte`. The entire species notes CRUD logic (fetch, create, edit, delete) is copy-pasted between them, and they've already diverged: the detail view has byte-length validation and `maxlength` on textareas, but the modal doesn't. This should be extracted into a shared composable. Related to that, `DetectionDetail.svelte` fires `deleteSpeciesNote()` on a single click with no confirmation, while the modal uses `window.confirm()`. The `SpeciesComparison.svelte` Props interface uses `[key: string]: any` which goes against the project's TypeScript rules.

There are also some housekeeping issues. The branch has 14 binary `.gotmp/` test database files committed, and an `AGENTS.md` file that doesn't belong in the codebase. The 100-commit history includes WIP commits, merge commits from internal fork PRs, and many "fix review findings" iterations that make the actual changes hard to follow.

Given the scope of what needs to change, I think the cleanest path forward would be for me to reimplement this feature using your PR as the reference design. The architecture and provider abstraction are solid, so the core approach stays the same, but I'd fix the issues above and keep the commit history clean. I'd credit you as co-author on the commits since this is fundamentally your design. Let me know if you're okay with that approach, or if you'd prefer to tackle the fixes yourself. Either way works for me.

iamrans0m00:
> Thanks for putting this together, this is a substantial piece of work and the overall architecture is really solid. The two-tier cache with singleflight deduplication, circuit breaker on Wikipedia, negative caching, and stale-while-revalidate refresh are all well thought out. I can tell a lot of design effort went into this.
> 
> That said, after doing a thorough review (including running the backend and frontend through separate deep reviews), I've found enough issues that I don't think this is mergeable in its current state. Let me walk through the main concerns so you have a clear picture.
> 
> On the backend side, the most pressing issue is in `PreFetch` in `guideprovider.go` around line 285. The goroutine captures the caller's request context, so when the HTTP request completes the background fetch gets cancelled. It should derive from `c.rootCtx` instead, the same way `Get()` already does. The eBird provider also downloads the entire taxonomy (~16K entries) on every `Fetch()` call and then does a linear scan to find one species. Even with the client-side cache this is wasteful; it should index by scientific name after the first fetch, or use a species-specific endpoint. The `sync.Map` in-memory cache has no eviction or size limit. With 16K species across 14 locales, each entry potentially holding 10KB of text, this could grow to ~2GB in theory. Some form of LRU or max-size cap would be good. There's also no locale validation at the API boundary in `species.go` around line 939; arbitrary strings get stored in the DB and used as map keys. A simple format check (2-3 lowercase letters) would close that off.
> 
> On the frontend, the biggest concern is the code duplication between `SpeciesDetailModal.svelte` and `DetectionDetail.svelte`. The entire species notes CRUD logic (fetch, create, edit, delete) is copy-pasted between them, and they've already diverged: the detail view has byte-length validation and `maxlength` on textareas, but the modal doesn't. This should be extracted into a shared composable. Related to that, `DetectionDetail.svelte` fires `deleteSpeciesNote()` on a single click with no confirmation, while the modal uses `window.confirm()`. The `SpeciesComparison.svelte` Props interface uses `[key: string]: any` which goes against the project's TypeScript rules.
> 
> There are also some housekeeping issues. The branch has 14 binary `.gotmp/` test database files committed, and an `AGENTS.md` file that doesn't belong in the codebase. The 100-commit history includes WIP commits, merge commits from internal fork PRs, and many "fix review findings" iterations that make the actual changes hard to follow.
> 
> Given the scope of what needs to change, I think the cleanest path forward would be for me to reimplement this feature using your PR as the reference design. The architecture and provider abstraction are solid, so the core approach stays the same, but I'd fix the issues above and keep the commit history clean. I'd credit you as co-author on the commits since this is fundamentally your design. Let me know if you're okay with that approach, or if you'd prefer to tackle the fixes yourself. Either way works for me.


Thank you, I am happy to have a hand with this.  If you want me to tackle specific issues just let me know and I will be happy to help.

iamrans0m00:
Closing in favor of #3646, which is a complete remake of this feature (reworked guide-provider package, two-tier caching, selectable similar-species comparison, notes, and full UI/i18n). Please review there instead.

PR #2774, opened 2026-04-15
fix(analysis): silence MQTT flood and floor jobqueue stop timeout

tphakala:
## Summary

Two surgical fixes for analysis-pipeline lifecycle edges:

### MQTT "client not available" flood
A live Sentry group surfaced 290 events/month with the pattern "MQTT client not available". Root cause: when `initializeMQTT()` fails at startup (broker unreachable, auth/TLS failure) or the client is between `DisconnectMQTTClient()` and `SetMQTTClient()` during reconfiguration, `p.MqttClient` is nil. Streaming publishers (sound level, publishing at 10-100/sec) called `PublishMQTT` which built a `CategoryMQTTPublish`-tagged error every time, forwarding each one to Sentry.

Fix: `PublishMQTT` now returns new sentinel `ErrMQTTClientNotReady` (plain `errors.New` — no telemetry category). The streaming publisher in `sound_level.go` checks the sentinel with `errors.Is` and silently drops (records a Prometheus metric `client_not_ready`, debug log only). A `sync.Once` emits a single warn on first occurrence so operators still learn MQTT is configured but unreachable; further attempts are silent by design.

The detection pipeline already gates `MqttAction` creation on `mqttClient != nil`, so this sentinel is only needed for streaming (non-detection) publishers.

### Job queue stop timeout floor
A live Sentry group surfaced the error "timed out waiting for jobs to complete after 0s" when `StopWithTimeout` was called with a zero or negative `time.Duration`. This happens when a caller derives the timeout from `context.Deadline()-time.Now()` and the parent context has already expired by the time `Stop` runs.

The canonical caller (processor shutdown) already floors the timeout at `MinJobQueueGracePeriod`. This PR adds a defensive `MinStopTimeout = 500ms` floor inside `StopWithTimeout` itself so any future caller gets the same safety net.

## Details

- `internal/analysis/processor/mqtt.go`: exported `ErrMQTTClientNotReady` sentinel, `PublishMQTT` returns it when client is nil with a once-per-process warn via new `Processor.mqttNotReadyWarnOnce sync.Once` field.
- `internal/analysis/sound_level.go`: sentinel detection via `errors.Is`, silent drop with Prometheus `publishing_errors{stage="client_not_ready"}` metric, debug log only.
- `internal/analysis/jobqueue/queue.go`: `StopWithTimeout(timeout)` floors `timeout <= 0` to `MinStopTimeout` (500ms).
- `internal/analysis/processor/processor.go`: one new `sync.Once` field on `Processor` — most of the diff is gofmt column realignment after the field insert.

## Test Plan

- [x] `go test -race ./internal/analysis/... ./internal/mqtt/...` — PASS
- [x] `TestPublishMQTT_NilClient_WarnsOnlyOnce` — 32 goroutines × 50 calls under `-race` verifies `sync.Once` contract
- [x] `TestPublishSoundLevelToMQTT_ClientNotReady_Idempotent` — 500 iterations stay silent
- [x] `TestStopWithTimeout_{Zero,Negative,Positive}Timeout` — floor contract
- [x] `golangci-lint run -v ./internal/analysis/... ./internal/mqtt/...` — 0 issues

## Summary by CodeRabbit

* **Bug Fixes**
  * MQTT publishing now degrades gracefully when the client is unavailable: emits a single warning and returns a non-fatal no-op instead of repeated errors.
  * Shutdown behavior improved with a minimum timeout floor to avoid edge-case blocking on immediate or non-positive timeouts.

* **Tests**
  * Added tests covering MQTT unavailability, one-time warning behavior, idempotent silent publishing, and shutdown timeout flooring.

### DIFF
--- a/internal/analysis/jobqueue/queue.go
+++ b/internal/analysis/jobqueue/queue.go
@@ -27,6 +27,15 @@ const (
 	// ActionStatsTargetSize is the target size after cleanup (with hysteresis margin)
 	// Set to 80% of max to avoid repeated cleanup triggers
 	ActionStatsTargetSize = int(MaxActionStatsEntries * 0.8)
+
+	// MinStopTimeout is the minimum grace period enforced inside
+	// StopWithTimeout when a caller passes a zero or negative timeout. The
+	// canonical caller (processor shutdown) already floors the timeout at
+	// its own MinJobQueueGracePeriod, but this constant provides a safety
+	// net for any future caller and prevents a 0s select-case fallthrough
+	// that would report "timed out waiting for jobs to complete after 0s"
+	// to telemetry even though the shutdown is behaving correctly.
+	MinStopTimeout = 500 * time.Millisecond
 )
 
 // JobQueue manages a queue of jobs that can be retried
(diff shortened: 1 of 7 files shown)


tphakala:
@coderabbitai All 3 nitpicks fixed in 5878bd50:

1. **jobqueue/stop_timeout_floor_test.go** — consolidated the three separate `TestStopWithTimeout_*` functions into a single table-driven `TestStopWithTimeout_FloorBehavior` with subtests, and replaced the hardcoded `500*time.Millisecond` elapsed assertion with the `MinStopTimeout` constant so the test stays tied to the production contract.
2. **processor/mqtt.go** — tightened `ErrMQTTClientNotReady` and `PublishMQTT` docs to state the sentinel is returned whenever the client reference is nil, regardless of the enabled flag. Matches actual behavior (no gated check on enabled/disabled exists in the function).
3. **sound_level_not_ready_test.go** — added `t.Cleanup(func() { conf.SetTestSettings(prev) })` in all three test functions that mutate the package-global settings, snapshotting the prior value via `conf.GetSettings()` before the override.

PR #2788, opened 2026-04-16
refactor(analysis): consolidate Float32Pool through audiocore buffer.Manager

tphakala:
## Summary

- Delete the duplicate `Float32Pool` implementation in `internal/analysis/float32_pool.go` so there is a single pool module owned by `internal/audiocore/buffer`.
- Thread `*buffer.Manager` through `ProcessData` so the 16-bit conversion hot path reaches pools via `buffer.Manager.Float32Pool(size)`, the existing lazy per-size pool map. This eliminates the silent fallback to `make()` for non-standard chunk sizes.
- Drop the package-level `float32Pool` global, `InitFloat32Pool()` / `ReturnFloat32Buffer()` helpers, and the startup call in `birdnet_service.go`.
- Add `TestConvert16BitToFloat32WithPool_ZeroAllocs` to lock in the reduced allocation profile across standard and non-standard sample sizes.

Part of the ongoing audiocore memory-regression work. This PR only touches the analysis side; router, `AnalysisBuffer.Read`, and new pool types remain for a follow-up.

## Test Plan

- [x] `go vet ./...` clean
- [x] `golangci-lint run -v ./internal/analysis/...` — 0 issues
- [x] `go test -race ./internal/analysis/...` green
- [x] `TestConvert16BitToFloat32WithPool_ZeroAllocs` passes with allocs <= 1 at both the standard `Float32BufferSize` and a non-standard size
- [x] End-to-end sound card validation via ALSA loopback: 7 MQTT detections received over a 3-minute window
- [x] Local CodeRabbit review: no findings

## Summary by CodeRabbit

* **Refactor**
  * Updated audio processing to use an injected buffer manager with per-size pooling, removing a legacy global float32 pool for more efficient, size-aware buffer reuse.

* **Tests**
  * Added a performance test ensuring float32 conversion reuses pooled buffers with minimal allocations.


tphakala:
Thanks for the nitpicks. Addressed both in a709b617:

- **`process_alloc_test.go` nil guard on Put:** applied. Test now mirrors the production guard in `convert16BitToFloat32WithPool`. In practice the test sizes are always positive so nil is unreachable, but the symmetry makes failure modes assert-style rather than panic-style.
- **`_ = sink`:** keeping it. Disagree that it is redundant — `testing.AllocsPerRun` is opaque to the compiler, but `sink` in the enclosing function scope is only ever written and never read without the trailing `_ = sink`, which permits standard dead-store DCE of the closure's assignment. It is 9 bytes of belt-and-suspenders.
- **`process.go` INVARIANT marker:** applied. Added an explicit INVARIANT note near the pool `Put` explaining the Predict-must-copy contract and the use-after-free risk if a future backend retains `sampleData`.

PR #2792, opened 2026-04-17
fix(audio): copy frame data in hlsConsumer.Write before channel send

tphakala:
## Summary

`hlsConsumer.Write` previously handed the caller's `frame.Data` slice header directly to a buffered channel read asynchronously by FFmpeg. Any upstream buffer reuse (for example, when the audiocore router eventually pools its output slices) would race with the FFmpeg reader and corrupt the live audio stream.

This PR copies `frame.Data` once at the top of `Write` so the consumer owns its payload for the lifetime of the channel enqueue. No behaviour change for the live HLS path (still drops oldest frame on channel saturation); cost is one owned copy per frame, orders of magnitude cheaper than the upstream allocations this unblocks pooling for.

Preparation for wiring the existing `buffer.BytePool` / new `Float64Pool` into the audiocore router hot path, which accounts for ~438 GB of 2-day allocation churn in production profiling.

## Test Plan

- [x] New `TestHLSConsumer_WriteDoesNotRetainFrameData` asserts post-Write mutation of the caller's slice is not visible on the channel (fails on main, passes with the fix).
- [x] `go test -race ./internal/api/v2/...` passes.
- [x] `go test -race ./internal/audiocore/... ./internal/analysis/...` passes.
- [x] `golangci-lint run -v` clean.
- [x] Local `coderabbit review --plain -t committed` reports no findings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented a potential race in HLS audio streaming by ensuring audio frame data is copied before asynchronous transmission, so buffers can be safely reused after sending.

* **Tests**
  * Added a unit test confirming transmitted audio frames are isolated from subsequent mutations of the original buffer.

### DIFF
--- a/internal/api/v2/audio_hls.go
+++ b/internal/api/v2/audio_hls.go
@@ -1478,8 +1478,7 @@ func (h *hlsConsumer) Write(frame audiocore.AudioFrame) error { //nolint:gocriti
 	}
 
 	// Copy once up front. Both select arms below need an owned slice.
-	buf := make([]byte, len(frame.Data))
-	copy(buf, frame.Data)
+	buf := slices.Clone(frame.Data)
 
 	select {
 	case h.ch <- buf:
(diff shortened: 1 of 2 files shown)


tphakala:
@coderabbitai addressed the `slices.Clone` nitpick in 52ef2638 — matches the 59 existing `slices.Clone` uses in the project. No Gemini re-review requested: first pass returned no feedback and this is a one-line idiomatic swap.

PR #2798, opened 2026-04-17
fix(analysis): defer SaveAudioAction read for Extended Capture tails

### LINKED ISSUE #2789: bug: Failed to read capture buffer for audio export component

nightly-20260414 Build Date: 2026-04-14T08:38:02Z

### Install method

Docker run / docker-compose (manual, ghcr.io/tphakala/birdnet-go)

### Device type

PC / Server (x86_64)

Debian 13

### Hardware details

- Ram: 16 GB
- Storage: NVME
-  HTTP Stream

### Description

On versions past 20260322, if Extended Capture is enabled, I get the errors like the following:

 ```ERROR [analysis.processor] Failed to read capture buffer for audio export component=analysis.processor.actions detection_id=TUFT_1616_c71d error="requested segment exceeds written data" source=http://xxx begin_time=2026-04-16T16:15:50-04:00 duration_seconds=44 operation=capture_buffer_read_for_export```

By disabling this option, it operates as before.  

### Steps to reproduce

1. Enable "Extended Capture", under the Recordings settings. I set Max Duration to 30s.
2. Restart if asked and then wait for a few captures. Once the first error shows up, it seems like all subsequent captures will fail. The detection is logged, but you are unable to review the audio or spectrogram.

### Expected behaviour

Audio and spectrograms will generate without error when the `Extended Capture` option is selected.

### Actual behaviour

 ```ERROR [analysis.processor] Failed to read capture buffer for audio export component=analysis.processor.actions detection_id=TUFT_1616_c71d error="requested segment exceeds written data" source=http://xxx begin_time=2026-04-16T16:15:50-04:00 duration_seconds=44 operation=capture_buffer_read_for_export```

### Relevant log output

```shell

```

_No response_

### Pre-submission checks

- [x] I am running the latest release or have tested with the `:dev` container image.

### Additional context

I tested with both 20260414 and the latest :dev container. Reverting back to 20260322 restores full functionality.

tphakala:
## Summary

Fixes #2789. When Extended Capture is enabled, the audio export path silently drops every clip with `error="requested segment exceeds written data"`.

`buildSaveAudioAction` computes `captureLength` from `EndTime - BeginTime + PreCapture` and then eagerly calls `readCaptureSegment`. With an extended clip whose tail has not yet been written to the ring buffer, `ReadSegment` correctly returns `ErrInsufficientData` because `endByteOffset > writtenBytes`. The resulting `SaveAudioAction` has nil `pcmData`, its `Execute()` becomes a no-op, and the user loses both audio and spectrogram for the affected detections.

## Fix

Defer the read when `captureEndTime` is in the future:

- `buildSaveAudioAction` returns a `SaveAudioAction` populated with `bufferMgr`, `sourceID`, `beginTime`, `duration`, and `readyAt` instead of `pcmData`.
- `SaveAudioAction.Execute()` checks these fields: if `readyAt` is still future it returns a sentinel `errAudioExportDeferred`; once `readyAt` has passed it reads from the capture buffer and falls through to the existing encode path.
- `getJobQueueRetryConfig` enables retry for deferred `SaveAudioAction` only. `MaxRetries` is computed from `duration` so short clips fail fast (6 retries, ~61s) while a `MaxExtendedCaptureDuration` (1200s / 20 min) tail still has enough schedule to complete. Backoff parameters (1s initial, 30s cap, 2.0x) are constant.
- If encoding fails after the buffer read, the populated `pcmData` lets the next retry skip straight to encode without re-reading a buffer that may have rolled over.

## Memory impact

Strictly better for Extended Capture. The pre-fix path eagerly copied up to ~5.8 MB of PCM per queued action. Deferred actions hold only a `buffer.Manager` pointer and a few timestamps until `readyAt`, then allocate `pcmData` at execute time.

## Test plan

- [x] Unit tests in `internal/analysis/processor/save_audio_action_test.go`:
  - defers with `errAudioExportDeferred` when `readyAt` is in the future
  - reads the capture buffer and writes a WAV when `readyAt` has passed
  - propagates errors from a missing capture source (final retry failure)
  - retry config is enabled with the backoff constants
  - eager `SaveAudioAction` (non-deferred) has retry disabled
  - `MaxRetries` schedule covers capture duration + margin across the supported range (10s / 60s / 300s / 1200s)
- [x] `golangci-lint run -v` clean
- [x] `go test -race` passes across `internal/analysis/processor/`
- [x] `coderabbit review --plain -t uncommitted` reports no findings
- [ ] Manual: enable Extended Capture, trigger several detections, confirm audio + spectrogram files appear on disk

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio export reliability: exports now defer and automatically retry when capture data isn’t yet available, reducing failed/empty exports. Retry timing adapts to capture length so longer recordings get longer retry windows.
* **Tests**
  * Added tests covering deferred export behavior, retry scheduling, and success/failure cases to ensure robust behavior.

### DIFF
--- a/internal/analysis/processor/actions_database.go
+++ b/internal/analysis/processor/actions_database.go
@@ -14,9 +14,16 @@ import (
 	"github.com/tphakala/birdnet-go/internal/conf"
 	"github.com/tphakala/birdnet-go/internal/datastore"
 	"github.com/tphakala/birdnet-go/internal/detection"
+	"github.com/tphakala/birdnet-go/internal/errors"
 	"github.com/tphakala/birdnet-go/internal/events"
 	"github.com/tphakala/birdnet-go/internal/logger"
 )
+
+// errAudioExportDeferred signals that SaveAudioAction cannot yet read the
+// requested capture segment because the tail of the clip is still being
+// written (Extended Capture). The job queue's retry mechanism re-runs the
+// action after a backoff delay until the buffer has caught up.
+var errAudioExportDeferred = errors.NewStd("audio export deferred until capture tail is available")
 
 // Execute logs the note to the log file.
 // Note: File logging errors are logged but not returned. This is intentional because:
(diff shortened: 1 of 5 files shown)


tphakala:
@coderabbitai Thanks for the review.

- **Tautological sentinel test**: fixed in cf5317b1. The test now wraps `errAudioExportDeferred` with `fmt.Errorf %w` and asserts `errors.Is` on the wrapped form, so the unwrap path is actually exercised.
- **Transient vs permanent buffer errors**: agreed this is a real improvement, but keeping it out of scope for this PR. The behavior is bounded (45 retries for the 1200s worst case) and you flagged it as non-blocking. I'll track it as a follow-up; the clean approach is a jobqueue-recognized non-retryable error wrapper applied when `CaptureBuffer` returns a "source not registered" error or `ReadSegment` returns the "endTime is before buffer's start time" case. That lands best as its own PR alongside the matching jobqueue change.

tphakala:
@coderabbitai No need, I'll track that one internally. Thanks.

PR #2821, opened 2026-04-20
[codex] Update Estonian BirdNET label translations

alarmatwork:
## What changed

This updates `internal/classifier/data/labels/V2.4/BirdNET_GLOBAL_6K_V2.4_Labels_et_ee.txt` by replacing English common names with Estonian names where an exact Latin-name match was available from official EOÜ naming sources.

## Why

The Estonian labels file contained many entries that were still left in English after the `_` separator. This makes the locale file inconsistent and less useful for Estonian users.

## Impact

The Estonian label set now uses EOÜ-backed Estonian bird names for matched taxa, while preserving the original Latin taxon keys.

## Validation

I cross-checked the replacements against the linked EOÜ Estonian bird list and the EOÜ world bird names page, matching by Latin name and preferring the Estonia-list wording where both sources covered the same species.

## Sources

- https://www.eoy.ee/ET/list-of-estonian-birds/
- https://www.eoy.ee/ET/16/31/birdnames/


alarmatwork:
Superseded by #2822, which renames the branch and updates the commit/PR metadata to match CONTRIBUTING.md while keeping the same translations update and source references.

PR #2841, opened 2026-04-21
fix: pick up config changes from UI

bo0tzz:
I noticed that weather api requests on my instance were still using old coordinates from before I changed the setting. I saw that there's a few other places with potentially similar behaviour, so figured I'd fix it a bit more broadly.

## Summary by CodeRabbit

* **New Features**
  * Spectrogram generation now samples runtime configuration once per job so export path, size/raw, style, dynamic range, and audio-tool paths take immediate effect.

* **Bug Fixes / Behavior**
  * Weather service pins the chosen provider at startup; provider selection won’t change on hot-reload.

* **Tests**
  * Added tests covering hot-reload of coordinates/API key and provider-pinning behavior.

### DIFF
--- a/internal/conf/config.go
+++ b/internal/conf/config.go
@@ -1883,6 +1883,21 @@ func GetSettings() *Settings {
 	return settingsInstance.Load()
 }
 
+// CurrentOrFallback returns the latest published settings snapshot, or the
+// supplied fallback when none has been published. It exists so long-lived
+// services can pick up UI edits without a restart: capture the initial
+// *Settings in a field, then call conf.CurrentOrFallback(s.settings) whenever
+// the value is actually read. In production the fallback is effectively
+// unreachable (settings are always loaded before services start); it's there
+// so unit tests that inject a custom *Settings into a struct literal — without
+// touching the package-global atomic pointer — keep working.
+func CurrentOrFallback(fallback *Settings) *Settings {
+	if latest := settingsInstance.Load(); latest != nil {
+		return latest
+	}
+	return fallback
+}
+
 // StoreSettings publishes a new *Settings snapshot atomically. Callers must
 // not mutate the pointee after calling StoreSettings; readers may observe the
 // snapshot concurrently through GetSettings. The canonical writer pattern is:
--- a/internal/spectrogram/generator.go
+++ b/internal/spectrogram/generator.go
@@ -227,8 +227,10 @@ func (g *Generator) currentSettings() *conf.Settings {
 
 // getDynamicRange returns the configured dynamic range value for Sox -z parameter.
 // Returns the default value ("100") if not configured or if an invalid value is set.
-func (g *Generator) getDynamicRange() string {
-	dr := g.currentSettings().Realtime.Dashboard.Spectrogram.DynamicRange
+// The settings snapshot is threaded in from the public entry point so the
+// whole generation sees a consistent view.
+func (g *Generator) getDynamicRange(settings *conf.Settings) string {
+	dr := settings.Realtime.Dashboard.Spectrogram.DynamicRange
 	if dr == "" {
 		return defaultDynamicRange
 	}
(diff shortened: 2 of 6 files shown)


tphakala:
Thanks for tracking this down and fixing it! The stale-coordinates bug is real and annoying, and I appreciate you looking at the spectrogram paths too, not just the weather service.

I went through the diff and have a couple of things I'd like to address before merging:

**Snapshot consistency in `processJob` and generator methods**

In `fetchAndSave` you correctly capture `currentSettings` once at the top and reuse it throughout the function. But in `processJob` (and the generator methods like `generateWithSoxFile`, `getSoxSpectrogramArgs`, etc.), `currentSettings()` is called multiple times independently. Each call does an atomic load, so if someone saves settings mid-job, `Size` could come from one snapshot and `Raw` from another.

Could you capture the settings once at the top of `processJob` and pass it through? Same idea as what you already did in the weather service. For the generator, the cleanest approach would be to capture it once per public entry point (`GenerateFromFile`, `GenerateFromPCM`) and thread it through the private methods.

**Provider name drift in weather metrics**

In `fetchAndSave`, `providerName` now reads from the fresh settings snapshot. But `s.provider` is still the one created at startup (as your comment correctly notes). If someone changes the provider in settings without restarting, the metrics and logs would be tagged with the new provider name while the actual requests still go through the old one.

I think the safest fix is to pin the provider name at construction time, maybe as a field on the Service struct, so it always matches the actual provider in use.

**Minor things (not blocking)**

- The `CurrentOrFallback` helper works, but you could also just call `conf.GetSettings()` directly since the test already calls `StoreSettings()`. Up to you, either way is fine.
- The test could use a short comment explaining why it's not `t.Parallel()` (we have that pattern in `config_race_test.go` for the same reason).

Let me know if you have questions about any of this. Happy to help if anything's unclear about how the settings atomic pointer works in the rest of the codebase.

bo0tzz:
Addressed all three review items in 350910a9:

**Provider name drift** — added a `providerName string` field on `Service` pinned at construction based on the actual provider implementation selected. All subsequent logs, metrics, and error-context use `s.providerName` instead of re-reading from the settings snapshot, so switching provider via the UI would never mislabel an in-flight request. Added `TestNewService_PinsProviderName` to document this.

**Snapshot consistency in `processJob` and generator methods** — both components now capture one snapshot per unit of work:
- `PreRenderer.processJob` captures `settings := pr.currentSettings()` at the top and threads it into `normalizeSpectrogramPath` and the subsequent `Size`/`Raw`/log reads. `normalizeSpectrogramPath` now takes a `*conf.Settings` so both callers (Submit path + processJob) control consistency.
- `Generator.GenerateFromFile` and `GenerateFromPCM` capture once and thread the snapshot through every internal method (`generateWithSoxFile`, `generateWithSoxDirect`, `generateWithFFmpegSoxPipeline`, `generateWithSoxPCM`, `generateWithFFmpeg`, `getSoxArgs`, `getSoxSpectrogramArgs`, `getDynamicRange`). `GetSoxSpectrogramArgsForTest` captures once internally.
- Generator-internal calls to `g.currentSettings()` are gone; only the three public entry points (and `GetSoxSpectrogramArgsForTest`) read from the global. The PreRenderer→Generator boundary still takes two snapshots by design — `Size`/`Raw` cross as per-call parameters, and render-engine config (`Style`, paths, etc.) belongs to the Generator's side.

**Test isolation (CodeRabbit + your note on `t.Parallel`)** — `TestFetchAndSave_HotReloadCoordinates` now captures `conf.GetSettings()` and restores it via `t.Cleanup`, and has a comment explaining why it skips `t.Parallel()` (matches the pattern in `config_race_test.go`).

On `CurrentOrFallback` — I kept it because the existing `&Service{settings: ...}` struct-literal tests in `weather_test.go` don't touch global conf state, so the fallback path is what lets those tests keep working without having to `StoreSettings` in each. Happy to simplify if you'd prefer the test churn instead.

tphakala:
Addressed the remaining CodeRabbit feedback:

**Test constants nitpick** - Fixed in 7830f111. The `TestNewService_PinsProviderName` test table now uses `yrNoProviderName`, `openWeatherProviderName`, and `wundergroundProviderName` constants instead of string literals. The previous commit (1acb6c61) already fixed the same issue in `NewService` itself.

**sunCalc stale coordinates (outside-diff)** - Valid observation, but out of scope for this PR. The `sunCalc` is pinned at startup with the initial lat/lon; after a coordinate change, sunrise/sunset in `DailyEvents` still uses the old location. This is a separate fix that should be tracked independently rather than expanding the scope of this community contribution. Filed as a follow-up.

Issue #2709, opened 2026-04-10
Audio served with gzip compression breaks playback in Chrome 147


rodrigodavid:
## Issue Type
- [x] Bug report

## Description
Audio files served via `/api/v2/audio/{id}` are compressed with gzip 
when the client sends `Accept-Encoding: gzip`. Chrome 147 (released 
April 2026) no longer accepts compressed audio streams, causing 
`net::ERR_CONTENT_DECODING_FAILED` and silent playback failure.

Binary media files (WAV/FLAC) should never be gzip-compressed — 
they don't benefit from compression and browsers don't expect 
compressed audio responses.

## Steps to Reproduce (for bugs)
1. Access BirdNET-Go through Cloudflare Tunnel (HTTPS)
2. Open any detection page in Chrome 147
3. Click play on the audio player

## Expected behavior
Audio plays normally. WAV/FLAC files are served without 
`Content-Encoding: gzip`.

## Actual behavior
Audio fails silently with `net::ERR_CONTENT_DECODING_FAILED` in 
Chrome 147. Safari is unaffected (more lenient with compressed audio).

Verified with:
```bash
curl -v http://localhost:8080/api/v2/audio/{id} \
  -H "Accept-Encoding: gzip" \
  --output /dev/null 2>&1 | grep -i "content-encoding"
# Returns: Content-Encoding: gzip
```

### Version Information
nightly-20260322

### System Information
- Device:
  - [x] Raspberry Pi (Raspberry Pi 5 Model B)
- OS: Raspberry Pi OS Debian Trixie (Lite 64-bit)
- Hardware specs:
  - CPU: Broadcom BCM2712, quad-core Cortex-A76 64-bit @ 2.4GHz
  - RAM: 16GB
  - Storage: 1TB NVMe (Lexar NQ790)
- Additional hardware:
  - USB Audio device: Behringer UMC202HD

## Possible Solution
The web server should skip gzip compression for audio MIME types 
(`audio/wav`, `audio/flac`, `audio/mpeg`, etc.), or set 
`Content-Encoding: identity` explicitly for `/api/v2/audio/*` routes.

<img width="861" height="575" alt="Image" src="https://github.com/user-attachments/assets/d87afc84-3e8a-4251-85e8-62ec50474009" />

birdwing:
Thank you!
I have been fighting this for 2 days. I thought it was a cloudflare issue.

I tried disabling all cache, compression and protections on cloudflare to try and fix the issue.
[...]

PR #2858, opened 2026-04-26
fix(securefs): add ETag header to prevent stale cache 416 errors

### LINKED ISSUE #2846: bug: Audio playback intermittent after gzip fix - HTTP 416 Range Not Satisfiable

nightly-20260418

### Install method

### Device type

Raspberry Pi 5

Raspberry Pi OS Debian Trixie (Lite 64-bit)

### Hardware details

- RAM: 16GB
- Storage: 1TB NVMe (Lexar NQ790)
- Reverse proxy: Cloudflare Tunnel

### Description

Following the fix for gzip compression of audio files (see #2709), audio playback is now intermittent. Some detections play correctly, others fail with HTTP 416 (Range Not Satisfiable). Reloading the page sometimes resolves the failure, suggesting a race condition or inconsistent range request handling.

Image

Image

### Steps to reproduce

1. Open any detection page via Cloudflare Tunnel in Chrome
2. Some detections fail to load audio with HTTP 416
3. Reloading the page sometimes fixes it, sometimes doesn't

### Expected behaviour

Audio loads consistently for all detections.

### Actual behaviour

HTTP 416 (Range Not Satisfiable) on `/api/v2/audio/{id}` intermittently.
Previously was ERR_CONTENT_DECODING_FAILED (gzip issue, now fixed).

### Relevant log output

```shell

```

_No response_

### Pre-submission checks

- [x] I am running the latest release or have tested with the `:dev` container image.

### Additional context

The intermittent nature (sometimes works on reload) suggests a possible race condition between audio file writing and range request handling, or an edge case in the new range-aware compression bypass.

tphakala:
## Summary

- Adds ETag header (mtime + file size) to all SecureFS-served files, fixing intermittent HTTP 416 Range Not Satisfiable errors on audio playback
- Root cause: after the gzip compression fix (#2709), browsers with cached gzipped audio receive 304 Not Modified (because only Last-Modified was checked), then use stale cached Content-Length for subsequent range requests, which fails with 416
- The ETag incorporates file size, so stale cache entries from the gzipped era get invalidated automatically

## Root Cause Analysis

Support dump analysis showed zero 416 responses server-side across 26K+ audio requests, confirming the 416 originates from browser cache state, not from the server. Validated via Cloudflare Tunnel QA test system: API audio routes pass through as `cf-cache-status: DYNAMIC`, confirming Cloudflare doesn't cache them.

The actual failure chain:
1. Pre-#2709: Chrome cached gzipped audio with compressed Content-Length
2. Post-#2709: `http.ServeContent` returns 304 when `If-Modified-Since` matches, ignoring Range headers entirely
3. Chrome uses its stale cached (gzipped) version with wrong Content-Length
4. Subsequent range requests based on the stale size produce 416

## Test plan

- [x] `TestAudioETagInvalidatesStaleCache` validates exact #2846 scenario: stale ETag gets 200/206 (fresh), not 304
- [x] `TestAudioRangeWithGzipMiddleware_EndToEnd` validates gzip+range through full middleware stack
- [x] `TestAudioRangeConsistencyAcrossSequentialRequests` simulates Chrome's caching pattern
- [x] `TestAudioConcurrentRangeRequests` stress test with 20 goroutines
- [x] All existing media/securefs/middleware tests pass with `-race`
- [x] Validated ETag behavior through Cloudflare Tunnel QA system
- [x] Production range request testing confirms correct behavior

Closes #2846

## Summary by CodeRabbit

* **New Features**
  * Added ETag support to improve HTTP caching for file downloads.

* **Bug Fixes**
  * Improved HTTP Range request handling to ensure correct partial-content responses with appropriate status codes and headers.

* **Tests**
  * Added regression tests covering range requests, caching validation, and audio streaming scenarios.

### DIFF
--- a/internal/api/v2/media_test.go
+++ b/internal/api/v2/media_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/labstack/echo/v4"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/tphakala/birdnet-go/internal/api/middleware"
 	"github.com/tphakala/birdnet-go/internal/audiocore/ffmpeg"
 	"github.com/tphakala/birdnet-go/internal/datastore/mocks"
 	"github.com/tphakala/birdnet-go/internal/imageprovider"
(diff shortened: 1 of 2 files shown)


tphakala:
Fixed the redundant `initMediaRoutes()` call in b3c4661c. Echo middleware applies at request time to all routes regardless of registration order, so the re-registration was unnecessary.

PR #2863, opened 2026-04-26
fix: use light foreground color on live stream page for spectrogram

rexxars:
On the live stream page, if you're using light mode, it is very hard to see the play button/"Audio Source" text. See below for comparison.

| light mode | dark mode |
| ---- | ----- |
| image | image |

This PR standardizes on light foreground color for this spectrogram, since it's overlayed on a black background

## Summary by CodeRabbit

* **Style**
  * Enhanced the start button appearance on the live stream page with improved text color contrast and clearer hover state visibility for better user interaction feedback.


tphakala:
Nice catch, thanks for fixing this. The white text on the black spectrogram background is the right call here.

PR #2867, opened 2026-04-27
refactor(conf): split config.go into focused files

tphakala:
## Summary

Split the 2,802-line `internal/conf/config.go` into four files organized by responsibility:

- **config.go** (1,643 lines): type/struct definitions, constants, `//go:embed`, pure query methods
- **migrations.go** (470 lines): all `Migrate*` functions, `persistMigration`, model validation helpers
- **storage.go** (562 lines): `Load()`, `initViper()`, `GetSettings()`, `CurrentOrFallback()`, `StoreSettings()`, `Setting()`, `SaveSettings()`, `SaveYAMLConfig()`, global state
- **test_helpers.go** (144 lines): `SetTestSettings()`, `GetTestSettings()`, `SettingsBuilder`

Zero behavior change: same package, same exports, same function signatures. Only file boundaries moved.

## Test Plan
- [x] `go build ./...` passes
- [x] `go test -race ./internal/conf/...` passes
- [x] `golangci-lint run -v ./internal/conf/...` - 0 issues
- [x] `git diff --stat` confirms changes only in `internal/conf/`

## Summary by CodeRabbit

* **Refactor**
  * Reorganized configuration management into separate storage, migration, and test modules for clearer behavior and reliability.

* **New Features**
  * Automatic migration of legacy config formats, improved validation with non-fatal warnings, and safer session-secret handling to preserve existing setups.

* **Tests**
  * New test helpers and a builder to easily construct and inject test configurations.


tphakala:
All three CodeRabbit nitpicks addressed in ebdf8c1e:

1. **storage.go**: Replaced four hand-rolled migration-save blocks with `persistMigration()` calls for TLS, OAuth, RTSP, and dashboard layout.
2. **migrations.go**: Changed `MigrateLocationConfigured()` to return `bool` like sibling `Migrate*` methods, with persistence delegated to the caller via `persistMigration()`.
3. **test_helpers.go**: Updated misleading `SetTestSettings` doc comment to accurately describe atomic publish behavior.

tphakala:
@coderabbitai The hardcoded known-IDs fallback is a valid concern, but it's out of scope for this PR which is strictly a file reorganization with zero behavior change. The map was moved as-is from the original config.go. I'll track extracting model ID constants as part of the broader Forgejo #492 conf package refactor.

PR #2873, opened 2026-04-27
fix(frontend): apply base paths to missing images, spectrograms etc

rexxars:
(Supersedes #2862)

My previous attempt at fixing basepaths was not exhaustive, it seems. Now that I've got an actual instance up and running behind a basepath, I found more URLs not working:
- Spectrograms
- Audio URLs
- Species images + placeholders

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved image and audio loading issues when the app operates with a custom base path, ensuring proper functionality in reverse proxy and non-root deployment scenarios.

* **Tests**
  * Added test coverage for species page rendering, detection components, and spectrogram URL resolution across various base path configurations.

### DIFF
--- a/frontend/src/lib/desktop/components/modals/ReviewModal.svelte
+++ b/frontend/src/lib/desktop/components/modals/ReviewModal.svelte
@@ -7,6 +7,7 @@
   import SpeciesThumbnail from './SpeciesThumbnail.svelte';
   import type { Detection } from '$lib/types/detection.types';
   import { fetchWithCSRF } from '$lib/utils/api';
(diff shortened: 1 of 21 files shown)


tphakala:
Thanks for this, great catch on all the missing basepath calls. The fix is clean and consistent, and the regression tests you added are solid.

I noticed one more spot that CodeRabbit flagged: `spectrogramUrl` in `DetectionCardMobile.svelte` (line 29) is still root-relative. I'll push a follow-up commit after merging to wrap that one in `buildAppUrl()` too.

PR #2899, opened 2026-04-29
fix(settings): eliminate stale *conf.Settings pointer reads across codebase

tphakala:
## Summary

- Add `currentSettings()` helpers to BirdNET and Controller structs (matching existing Processor pattern)
- Fix all runtime settings reads to use fresh atomic snapshots instead of stale startup pointers
- Covers 8 files across classifier, API v2, processor, and analysis packages

### HIGH severity fixes (user-visible settings not taking effect)
- **Predict method**: sensitivity and labels now read from fresh settings on every inference call
- **Species cache key**: lat/lon/model used for cache key now reflect UI changes
- **LocationConfigured**: range filter activation responds to location changes without restart

### MEDIUM severity fixes
- Dynamic threshold API stats return current ValidHours/MinThreshold
- MQTT/BirdWeather/Weather integration endpoints use current broker/coordinates
- Extended capture and daylight filter read labels from fresh settings
- Control monitor reads labels from fresh settings after model reload
- NumSpecies/Labels accessors return current label set

### Deferred
- OAuth2Server (8 stale read sites) requires test infrastructure updates; tracked separately

## Test plan
- [x] `go build ./...` passes
- [x] `golangci-lint run -v` reports zero issues
- [x] Pre-commit hooks pass
- [x] `go test -race ./internal/analysis/...` passes
- [x] `go test -race ./internal/api/v2/...` passes
- [x] `go test -race ./internal/security/...` passes
- [x] Classifier test failure confirmed pre-existing (TFLite checkptr issue on main)

## Summary by CodeRabbit

* **Improvements**
  * Runtime configuration now applies immediately across the system without restarts.
  * BirdNET labels, detection thresholds, species filters, and location-based behavior use the latest settings snapshot.
  * MQTT, BirdWeather/Weather integrations, dynamic thresholds, and inference logic read current configuration so monitoring and connections reflect up-to-date settings.

### DIFF
--- a/internal/analysis/control_monitor.go
+++ b/internal/analysis/control_monitor.go
@@ -303,8 +303,8 @@ func (cm *ControlMonitor) handleReloadBirdnet() {
 		cm.notifySuccess("Range filter rebuilt successfully")
 	}
 
-	// Rebuild name maps with new locale labels
-	labels := cm.bn.Settings.BirdNET.Labels
+	// Rebuild name maps with new locale labels (use fresh settings, not stale pointer)
+	labels := conf.CurrentOrFallback(cm.bn.Settings).BirdNET.Labels
 	if cm.proc != nil && cm.proc.Ds != nil {
 		cm.proc.Ds.UpdateNameMaps(labels)
 		GetLogger().Info("Datastore name maps updated with new labels")
--- a/internal/analysis/processor/daylight_filter.go
+++ b/internal/analysis/processor/daylight_filter.go
@@ -41,10 +41,10 @@ func (p *Processor) initDaylightFilter() {
 		return
 	}
 
-	// Get BirdNET labels for common name resolution
+	// Get BirdNET labels from fresh settings for common name resolution
 	var labels []string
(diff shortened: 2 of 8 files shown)


tphakala:
Addressed CodeRabbit findings in 442eff4a:

1. **MQTT snapshot consistency** (Major): Fixed. `checkMQTTConnectionStatus` now accepts the settings snapshot as a parameter instead of re-reading independently. The entire MQTT probe runs on one consistent snapshot.

2. **BirdWeather formatClientError** (Minor): Fixed. Changed `c.Settings` to `currentCfg` on the error path after `birdweather.New()`.

3. **Cache key vs GetProbableSpecies snapshot** (Major): Acknowledged but out of scope. `GetProbableSpecies` already uses `conf.CurrentOrFallback(bn.Settings)` internally (from PR #2897). The theoretical inconsistency window is tiny (atomic pointer swap between two calls in the same goroutine), and the consequence is a cache miss/re-fetch, not incorrect data served. Fixing this properly requires refactoring `GetProbableSpecies` to accept a settings parameter, which touches callers outside this PR's scope.

Issue #2813, opened 2026-04-18
feat: Include details about which RTSP stream for detections


oldjohngalt:
### What problem does this solve?

I have multiple exterior security cameras and feed the different streams to Birdnet.  It would be great to have the detection associated to a specific stream.

### Proposed solution

This would allow additional analytics and reporting. For example, was a detection in the front yard or the back yard?  There are great metrics and analytics.  Being able to sort/filter by stream would be nice.

### Pre-submission checks

- [x] I have searched existing open and closed issues and this feature has not already been requested.

uSlackr:
https://github.com/tphakala/birdnet-go/discussions/1889

oldjohngalt:
Thanks! Sorry for the duplicate suggestion.

tphakala:
Have you updated to latest release, because it already contains this https://github.com/tphakala/birdnet-go/releases/tag/nightly-20260418

**Audio Source Labels in the Detection UI**

Detection views now display the human-readable source name (e.g. "Garden feeder") instead of a raw RTSP URL or device ID. The desktop detections table gains a new Source column at lg and above, mobile detection cards show the source name inline after the timestamp, and the Search view recognises a new source: filter that maps to the existing location API parameter. The API's DetectionResponse.source field is now a structured SourceInfo object (id, type, displayName) instead of a plain string, aligning with the realtime/SSE path that already used the structured form. Detections whose source cannot be resolved from the live AudioSource table now fall back to the configured source name in settings instead of showing an ID (https://github.com/tphakala/birdnet-go/pull/1681 by @jasonmadigan, https://github.com/tphakala/birdnet-go/pull/2757, addresses https://github.com/tphakala/birdnet-go/issues/1494).

mgsb:
Is there a config setting for this new feature? I am running 20260418 and do not see audio source labels in detections on either desktop or mobile.

jvilhuber:
Ah interesting. I see it in some places:

<img width="1282" height="523" alt="Image" src="https://github.com/user-attachments/assets/26d8c433-a0ce-4505-852b-1625f0594cfd" />

but not others:

<img width="1266" height="620" alt="Image" src="https://github.com/user-attachments/assets/0703a86d-cc78-4668-997e-462fd14a9cf4" />

also not here: 

<img width="1333" height="603" alt="Image" src="https://github.com/user-attachments/assets/dce3e534-43e1-47b7-aebc-e0e5ef664029" />

PR #3020, opened 2026-05-11
feat(analytics): per-source filter on /ui/analytics

jdevalk:
## Summary

Adds an optional `source_id` query parameter to every `/api/v2/analytics/*` endpoint and a new multi-select source picker on the `/ui/analytics` page, so users with multiple audio sources (e.g. several cameras) can scope every chart on the analytics page to one or more sources.

Addresses the analytics half of #2813. The detections half — Source column / `source:` filter in the search view — already shipped; this PR follows that pattern to the analytics page.

## What it does

**Backend** (v2 schema only — legacy `notes` accepts the param and ignores it, documented in the API README):

- New `GET /api/v2/analytics/sources` listing historical audio sources with detection counts, sorted descending. Distinct from `/system/audio/sources` (live engine sources) — this returns the rows that actually have detections.
- Variadic `sourceIDs ...uint` parameter added to:
  - `GetSpeciesSummary`, `GetHourlyDistribution`, `GetDailyAnalytics`, `GetDetectionTrends`, `GetNewSpecies`, `GetSpeciesFirstDetectionInPeriod` (v2 detection repo)
  - All matching wrappers in `v2only.Datastore` (incl. `GetSpeciesDiversityData`)
  - `datastore.Interface` (legacy `DataStore` accepts and ignores)
- Central `buildAnalyticsBaseQuery` helper applies the filter as `WHERE detections.source_id IN (...)`. Raw-SQL methods (`GetNewSpecies`, `GetSpeciesFirstDetectionInPeriod`) use a new `buildSourceFilterClauses` helper that splices the IN list into both the derived table and the outer join so per-source first-seen semantics are correct.
- API handler `parseOptionalSourceIDs` parses `?source_id=1,3,5` — comma-separated, deduped, rejects zero/negative/non-numeric, capped at 64 IDs.

**Frontend** (Svelte 5):

- `Analytics.svelte` adds `sourceGroups` to filter state, fetches `/analytics/sources` on mount, resolves selected display names back to underlying `audio_sources.id` values via a `$derived` map, and threads `source_id` into all 5 fetch URLs.
- `FilterForm.svelte` gets a multi-select dropdown that's hidden when there are fewer than 2 sources (single-source installs see no change).
- Source rows are grouped by `display_name` in the picker so duplicate `audio_sources` rows (which can happen when `node_name` or `source_uri` changes) collapse to one entry. The API request expands the selected name back to the full ID list.
- i18n keys `analytics.filters.audioSource{,All,Unknown}` added to all 14 locales.

## How `source_id` semantics work

`?source_id=1,3` is an IN-clause filter. For `GetNewSpecies` specifically, "first ever" is scoped to the selected sources — so a species can show up as new-for-this-source even if it was detected earlier on a different source. This matches how a user would intuit a per-source view.

The legacy `notes` schema has no `audio_sources` FK; on that path the parameter is silently ignored (documented in `internal/api/v2/README.md`).

## Test plan

- [x] `internal/datastore/v2/repository/detection_analytics_source_filter_test.go` — 4 funcs / 11 subtests, run with `-race`:
  - no filter returns all sources combined
  - single source restricts to that source
  - multi-source IN unions results
  - nonexistent source returns empty (not error)
  - per-source `GetNewSpecies` excludes a species that was first-seen on a different source within the window
- [x] `internal/api/v2/analytics_source_filter_test.go`:
  - `parseOptionalSourceIDs` table-driven across 11 inputs (empty / single / list / whitespace / dedupe / zero / negative / non-numeric / all-invalid / whitespace-only)
  - Truncation at `maxSourceIDsPerRequest` (64)
  - Handler threads `source_id` into the datastore call (captured via mockery)
  - `ListAnalyticsSources` happy path / empty list / datastore-error
- [x] Frontend `npm run check:all` (prettier, eslint, stylelint, svelte-check, ast-grep) is clean
- [x] All 1918 existing frontend tests still pass, including `FilterForm.test.ts`
- [x] Mocks regenerated via `mockery` and committed
- [x] Live-tested against a real install with ~5,800 detections across 3 active sources: filter counts match expected sub-totals; nonexistent source returns empty cleanly

## Notes for review

- I picked `audio_sources.id` (integer PK) as the filter token rather than `source_uri` or `display_name`. Reason: it's stable, it's the actual FK, and it keeps the IN-clause precise even when `display_name` has duplicates from `node_name` changes. The frontend hides this by grouping the picker by display name and expanding to the underlying IDs in the request.
- The variadic `sourceIDs ...uint` signature was chosen over a slice or options struct to keep existing callers (especially the species tracker in `internal/analysis/species/database.go`) source-compatible — they pass no source IDs.
- `.dockerignore` extended with `**/node_modules` because a stale local `frontend/node_modules` (host-platform binaries) can leak into the build context and break `svelte-check` inside the container. Bit me during local testing; figured I'd save the next contributor the hour.
- I only wired the picker into `Analytics.svelte` for this PR. `AdvancedAnalytics.svelte` and `Species.svelte` hit additional endpoints (`/time/hourly/batch`, `/time/daily/batch`, `/species/diversity`, `/species/thumbnails`) — those endpoints already accept `source_id` on the backend, so a follow-up to add the picker there should be small.

## Summary by CodeRabbit

* **New Features**
  * Audio-source filtering in Analytics: multi-select picker (shown when multiple sources), supports grouped sources, clears on reset; analytics queries scoped to selected sources and charts refresh on selection.

* **Documentation**
  * API docs updated to describe the optional source_id query parameter and behavior.

* **Localization**
  * Added audio-source filter translations for multiple locales.

* **Tests**
  * Added parsing, API, repository, and UI fixture tests for source filtering.

* **Chores**
  * .dockerignore updated to exclude dependency and build output directories.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3020)


jdevalk:
Reverted the anonymization for `/api/v2/analytics/sources` in 0addb5cd.

The bots (CodeRabbit, Sentry, Gemini) were right that anonymizing source metadata for unauthenticated clients is the safe default. But on closer inspection, the same fields are already fully exposed by `GET /api/v2/detections`, which is registered on the same public group and returns `source.id` (= SourceURI, often containing RTSPS stream tokens), `source.displayName`, and `source.type` for every detection record. Example from a default install:

```json
{
  "data": [
    {
      "id": 5822,
      "source": {
        "id": "rtsps://192.168.1.1:7441/dCzTleUCySg9ApbX?enableSrtp",
        "type": "unknown",
        "displayName": "Kabelbaan"
      },
      ...
    }
  ]
}
```

So anonymizing only on `/analytics/sources` while leaving `/detections` fully public:

- Doesn't actually protect the operator (any public viewer can pull the detection list and learn the same names + URIs).
- Breaks the picker UX for unauthenticated operators viewing their own dashboard without configuring auth.
- Introduces an inconsistency that a future contributor will either have to fix on both endpoints or remove from this one.

If upstream wants to tighten this exposure, the right scope is a coordinated change covering `/detections`, the SSE source streams, and this endpoint together — likely with a `PublicAccess.SourceMetadata` setting mirroring `PublicAccess.LiveAudio`. That feels out of scope for an analytics-filter PR; happy to file a follow-up issue.

The non-auth findings from those earlier commits (strconv.Itoa cleanup, prefix-aware sources query via the audio source repo, and the `GetSpeciesFirstDetectionInPeriod` WHERE-splice SQL bug fix) are kept.

jdevalk:
Addressed three of the four findings in 44603b83:

- **CodeRabbit (Major)** — wrapped `GetTimeOfDayDistribution`, `processHourlyBatchSpecies`, `processDailyBatchSpecies` in `context.WithTimeout(..., analyticsQueryTimeout)` to match neighbor handlers. The two batch processors share one budget across all N per-species queries, same as the rest of this file.
- **Sentry (Medium)** — added explicit `gorm:"column:display_name"` tag on the anonymous row struct in `GetAllWithDetectionCount`. Every other column already had explicit tags.
- **CodeRabbit (Minor)** — added Set dedup + `Array.slice(0, 64)` clamp on `selectedSourceIdsParam` in `Analytics.svelte`, `Species.svelte`, and `AdvancedAnalytics.svelte`. Each page now has a local `MAX_SOURCE_IDS_PER_REQUEST = 64` constant cross-referencing `parseOptionalSourceIDs` in `internal/api/v2/utils.go`.

**Sentry (Low) — SQLite `ROW_NUMBER()` in `GetSpeciesFirstDetectionInPeriod`**: declining this one because it's pre-existing. The function and its `ROW_NUMBER() OVER (PARTITION BY ...)` window query shipped on `main` long before this PR — I only added a variadic `sourceIDs` filter to the existing SQL. SQLite 3.25 was released in 2018 and is what every distribution currently ships, so the practical risk is low; documenting a minimum SQLite version is a sensible cleanup but feels like a separate concern from per-source analytics filtering. Happy to file a follow-up issue if it's something maintainers want pinned down.

tphakala (changes requested):
Two things need fixing before this can go in.

The Species.test.ts tests don't mock /api/v2/analytics/sources. When fetchAudioSources fires on mount and the fetch rejects (the mockFetchSequence returns Promise.reject for unhandled URLs), the catch block calls logger.error() but logger is undefined in the test environment because loggers.analytics isn't set up there. That produces two unhandled TypeErrors. Adding '/api/v2/analytics/sources': () => ({ sources: [] }) to the mockFetchSequence call in both tests should fix it.

The Go lint also fails. Six gocritic issues in the new test files: five places in detection_analytics_source_filter_test.go use context.Background() where the project linter wants t.Context(), and analytics_source_filter_test.go:113 has append([]uint(nil), src...) which should be slices.Clone(src).

Everything else looks solid. The SQL filter building is safe (manual placeholder strings, but args are properly parameterized through GORM), the legacy path ignoring sourceIDs is correct and documented, and the API design is clean. The duplicated AudioSourceOption interface and fetchAudioSources logic across three page components is worth a follow-up issue but not a blocker for this PR.

jdevalk:
Re: the Sentry comment on `internal/datastore/v2only/datastore.go:2590` (hardcoded `"detections"` table breaks on v2_-prefixed MySQL) — the hardcoded name there is pre-existing (`git blame` points at e1dd4b1e, Feb 14), so it's outside the scope of this PR. The same bug also affects three sibling queries in the same file (`GetTopBirdsData`, `GetBatchHourlyOccurrences`, `DeleteClipPathsBatch`).

Opened #3054 off main to fix all four uniformly with a prefix-aware pattern. Doesn't touch anything else.

tphakala:
Hi @jdevalk,

Thank you for submitting this PR. The features are useful and the core code implementation is solid.

However, I cannot merge this PR in its current state due to a few issues and translation errors:

1. PR Scope splitting
To keep reviews manageable, minimize merge conflicts, and allow safe, independent rollbacks if a regression occurs, I have a policy of keeping contribution PRs focused on exactly one feature or concern.

This PR batches three major, distinct features:
1. **Analytics per-source filtering** (the filter drop-down + backend query additions).
2. **Saved clip reanalysis** (the decoder + `POST /api/v2/detections/:id/reanalyze` endpoint).
3. **Operator-driven species correction** (`POST /api/v2/detections/:id/correct-species` + frontend modals).

Could you please split these into three separate branches/PRs?
- PR 1: Analytics per-source filter
- PR 2: Saved clip reanalysis
- PR 3: Operator-driven species correction (dependent on PR 2)

2. Translation Validation / ICU Syntax Regression
The PR modified the `notifications.content.buffer.overloadMessage` translation in 7 locale files (`de.json`, `es.json`, `fi.json`, `fr.json`, `nl.json`, `pl.json`, `pt.json`), replacing double single-quotes (`''{sourceName}''`) with single single-quotes (`'{sourceName}'`).

In the ICU MessageFormat parser used by the frontend i18n framework, single quotes escape braces. Changing this to `'{sourceName}'` escapes the `{sourceName}` variable, causing it to be treated as literal text. This breaks the parameter interpolation and fails `npm run i18n:validate`.

To fix this, we should revert those changes or restore the double single-quotes (`''{sourceName}''`).

3. Translation Quality Regression
In multiple translation files (e.g. `de.json`, `pt.json`), several previously localized strings were accidentally reverted/replaced with English strings (e.g., replacing `"Einstellungen"` with `"Settings"`, `"Vogelerkennung"` with `"Bird Detection"`, etc.). Please make sure to preserve the existing localized translations.

tphakala:
This PR still bundles many unrealated changes, I am closing this as PR history is too messy at this point. Please file a clean PR if you still want this feature reviewed.

PR #3024, opened 2026-05-12
fix(Perch): add location filtering for perch model

ModerateWinGuy:
## Problem

When running both BirdNET and Google Perch classifiers simultaneously, Perch
detections completely bypassed the geographic range filter. Birds outside the
configured location appeared in the detection list.

The range filter check in shouldFilterDetection only applied to models whose ID
starts with "BirdNET". Since Perch's model ID is "Perch_V2", the condition was
always false for Perch — the range filter was silently skipped for every Perch
detection.

## Fix

The range filter now applies to all bird classifiers when location-based
filtering is active, with two guards:

- Bat is exempt: bat species are not in the bird range filter list — applying
  it would block all bat detections.

- Non-BirdNET models only filter when location is configured: when no location
  is set, the range filter species list is built from BirdNET's ~6,400 labels
  only. Perch has ~14,795 species; Perch-exclusive species absent from BirdNET's
  label set would be incorrectly blocked. When a location is configured, the
  list is geographically filtered and the check is correct for Perch too.

BirdNET behaviour is unchanged.

## Tests

Added range_filter_test.go with 8 sub-tests covering:

- Perch out-of-range species filtered when location configured (the bug)
- Perch in-range species passes when location configured
- Perch species not filtered when no location configured (Perch-exclusive regression)
- BirdNET behaviour unchanged across all location states
- Bat always exempt from the bird range filter

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined species range filtering logic to correctly apply across different bird detection classifiers while properly exempting certain classifier types from range-based filtering.

* **Tests**
  * Added unit tests to verify range filter behavior across multiple detector types and location configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3024)

### DIFF
--- a/internal/analysis/processor/processor.go
+++ b/internal/analysis/processor/processor.go
@@ -886,12 +886,18 @@ func (p *Processor) shouldFilterDetection(settings *conf.Settings, result datast
 		return true, confidenceThreshold
 	}
 
-	// Range filter applies only to BirdNET; Bat/Perch have independent species sets.
-	if strings.HasPrefix(modelID, detection.DefaultModelName) && !settings.IsSpeciesIncluded(result.Species) {
+	// Apply range filter to all bird classifiers when location-based filtering is active.
+	// Bat is excluded: bat species are not in the bird range filter species list.
+	// For non-BirdNET models (Perch, BSG, etc.), only apply when location is configured:
+	// the species list is built from BirdNET labels, so model-exclusive species absent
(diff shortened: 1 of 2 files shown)


tphakala:
I added BirdNET v3.0 geomodel support which is applied to Perch v2 now, this legacy range filter support is now obsolete.

ModerateWinGuy:
Sounds good 👍

Issue #3044, opened 2026-05-12
bug: per-stream/source Models edits don't take effect until restart (hot-reload broken)


jdevalk:
# bug: Per-stream model assignment doesn't take effect without container restart

## Summary

After installing a model from the gallery (e.g. Perch v2) and adding it to a stream's `models` list via Settings → Audio Capture → save, the running orchestrator continues to classify with the *original* model set. The save persists to disk and the API reports the new state, but the actual analysis pipeline never picks it up. Two distinct bugs combine to produce this experience.

Repro on my install (4 RTSP streams, freshly merged main as of `e89cba2a`):

1. Install Perch v2 from model gallery → success, geomodel companion auto-downloads
2. Settings → Audio Capture → tick "Perch v2" on each stream → Save
3. Toast says "saved successfully"
4. Reload page → checkboxes are back to BirdNET only

Even after working around step 4 (see bug #1 below) by saving via direct PUT to `/api/v2/settings`, every detection in the log still shows `model_id=BirdNET_V2.4`, `model_count=1`. Perch v2 is installed, files on disk, but never gets handed audio frames.

## Bug 1: stale `c.Settings` after `applyConfigForInstall`

`internal/classifier/model_manager.go:935` publishes the updated settings via `conf.StoreSettings(updated)` and persists with `conf.SaveSettings()`, but never touches `Controller.Settings` on the API side. After installing a model:

```
On disk (correct):            models.enabled = [birdnet, perch_v2]
conf.GetSettings() (correct): models.enabled = [birdnet, perch_v2]
c.Settings (stale):           models.enabled = [birdnet]
```

`GetAllSettings` (`internal/api/v2/settings.go:95`) reads `c.Settings` directly, so the frontend GETs a pre-install view. Worse, `UpdateSettings` clones from `c.Settings` as its baseline (line 230) — so when the user saves changes to streams, the PUT goes through merge logic with stale `models.enabled`, and the validator that flags "stream X references model not in models.enabled" trips on the merged-but-stale clone.

Fix sketch: have `applyConfigForInstall` (and the symmetric `applyConfigForUninstall`) notify the controller, or have the controller observe `conf.StoreSettings` events. Simplest is probably to make `GetAllSettings` / `UpdateSettings` go through `conf.GetSettings()` like the global publisher does, and treat `c.Settings` as a test-injection point only.

## Bug 2: `streamsSettingsChanged` ignores `stream.Models`

[...]

PR #3045, opened 2026-05-12
fix(api/v2): hot-reload when per-stream/source Models list changes

jdevalk:
## Summary

- `streamsSettingsChanged` and `audioDeviceSettingChanged` now compare the per-stream / per-source `Models` list, so adding or removing a classifier (e.g., ticking Perch v2 alongside BirdNET in the Audio Capture settings) actually triggers `reconfigure_rtsp_sources` / `reconfigure_audio_sources`.
- Before this fix, the PUT persisted to disk and the API reported the new state, but the running orchestrator never rebuilt its stream → model bindings — audio kept flowing only to the original model set until a container restart. That breaks the hot-reload contract in `CLAUDE.md`.

Reproduced on a 4-RTSP-stream install: ticked Perch v2 on each stream, save returned 200, on-disk YAML showed `[birdnet, perch_v2]` per stream, but every detection in the log stayed `model_id=BirdNET_V2.4, model_count=1` until the container was restarted. After restart, `model_count=2` consensus appeared immediately.

## Changes

- `internal/api/v2/settings.go` — extend `streamsSettingsChanged` to compare `stream.Models` via `slices.Equal`.
- `internal/api/v2/settings_audio.go` — extend `audioDeviceSettingChanged` to compare `source.Models` (in addition to the deprecated singular `source.Model`).
- `internal/api/v2/settings_update_test.go` — three new tests:
  - `TestStreamsSettingsChanged_DetectsModelEdits` (table: identical / added / removed / reordered / replaced / both empty / nil vs `[]`)
  - `TestAudioDeviceSettingChanged_DetectsModelsEdits` (table: identical / added / removed / replaced)
  - `TestStreamsSettingsChanged_BackwardCompatibility` (sweep over Name / URL / Enabled / Type / Transport / count to confirm the new check is additive)

`slices.Equal` is order-sensitive, which is what we want: downstream inference iterates the list in order, so a reorder is a real change.

This addresses the second of the two bugs in #3044. The first bug (stale `Controller.Settings` snapshot after `applyConfigForInstall`) has more surface area and I've left it for maintainer guidance per the discussion in the issue.

## Test plan

- [x] `go test -race -run TestStreamsSettingsChanged ./internal/api/v2/...` — covered by CI; my local host doesn't have the TFLite C headers but the modified package is otherwise self-contained.
- [x] `go test -race -run TestAudioDeviceSettingChanged ./internal/api/v2/...` — same.
- [x] Manual repro: 4 RTSP streams, install Perch v2 from gallery, tick on each stream, save. Before fix: no `model_count=2` until restart. After fix (verified by building the patched image and saving via the UI): `reconfigure_rtsp_sources` fires, classifiers reload, `model_count=2` consensus on Roodborst / Merel across all four streams within seconds.

Closes the per-stream-model half of #3044.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of changes to assigned models/classifiers for streams and audio devices so configuration hot-reloads reliably trigger when model assignments change.

* **Tests**
  * Added tests verifying model-list edits (add/remove/reorder/replace) and backward compatibility of change detection for streams and audio sources.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3045)

### DIFF
--- a/internal/api/v2/settings.go
+++ b/internal/api/v2/settings.go
@@ -2225,7 +2225,14 @@ func mqttSettingsChanged(oldSettings, currentSettings *conf.Settings) bool {
 		oldMQTT.TLS.ClientKey != newMQTT.TLS.ClientKey
 }
 
-// streamsSettingsChanged checks if stream settings have changed
+// streamsSettingsChanged checks if stream settings have changed in a way that
+// requires the audio engine to reconfigure RTSP sources.
+//
+// Per-stream Models is included: when a user adds or removes a classifier on
+// a stream (e.g., enables Perch v2 alongside BirdNET) the orchestrator must
+// rebind the stream's analysis pipeline. Without this check, the save
+// persists to disk but the running pipeline keeps using the previous model
+// set until a restart — silently breaking the hot-reload contract.
 func streamsSettingsChanged(oldSettings, currentSettings *conf.Settings) bool {
 	oldRTSP := oldSettings.Realtime.RTSP
 	newRTSP := currentSettings.Realtime.RTSP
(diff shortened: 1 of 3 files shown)


tphakala:
Thank you for contributing!

PR #3047, opened 2026-05-12
feat(detections): reanalyze a clip with every loaded model + apply correction in one click

jdevalk:
## Summary

Adds a "second opinion + correction" loop on the detection detail page. Operators can:

1. **Reanalyze a saved clip** against every currently-loaded classifier in parallel and see a side-by-side confidence grid (BirdNET v2.4 vs Perch v2 vs anything else they've installed). No persistence — purely read-only inference.
2. **Apply a correction**: click "Use this" on the species they think is right, confirm, and the detection's `label_id` / `model_id` / `confidence` are updated in place and the row is marked `verified='correct'` in a single transaction.

The motivation: with PR #3032 + #3035 making Perch v2 a one-click install alongside BirdNET v2.4, users now have two strong classifiers but no way to ask "what does Perch think of this BirdNET call?" without losing the original or running models manually. This closes that loop.

Reproducible example on my own install: BirdNET-Go called a clip 99% **Aegithalos caudatus** (Long-tailed Tit). Reanalyze shows BirdNET 99.6% LTT but **99.8% Pied Flycatcher** in the same clip, and Perch v2 says Pied Flycatcher 86.9%, LTT 69.8%. Both classifiers agreeing on PF turns a contested call into a clear correction.

## Endpoints

- `POST /api/v2/detections/:id/reanalyze` — body `{modelIds?: string[]}`. Empty list means "every loaded compatible model"; explicit list runs only those. Returns top-10 species with per-model confidence: `{predictions: [{scientificName, commonName, byModel: {modelId: conf}}, ...], modelsRun: [...], clipDurationSec, ...}`. Auth-protected via `detectionGroup` middleware.
- `POST /api/v2/detections/:id/correct-species` — body `{scientificName, modelId, confidence}`. Updates the detection in place + upserts the review row to `verified='correct'` inside one DB transaction. Rejects unknown species/model and locked detections.

## Implementation notes

- **Multi-model decode is grouped by sample rate** so we only spawn one ffmpeg per unique rate, not one per model. BirdNET v2.4 (48k) + Perch v2 (32k) = two decodes regardless of how many models share a rate. Future bat models with `RawSampleRate=256000` are excluded by default since RTSP audio is not ultrasonic.
- **Window walking matches the realtime pipeline**: 50% overlap, max-per-species across windows. Per-model results are passed through `classifier.SplitSpeciesName` + the orchestrator's resolver chain so locale-specific common names (Dutch in my case, 24 locales via the v3 geomodel taxonomy companion from #3042) end up in the response.
- **Correction goes through raw SQL on `c.DS.Transaction(...)`** rather than the `datastore.Interface` — `v2only.Datastore` doesn't expose a generic `UpdateNote` surface and adding one would have rippled through mockery mocks for a single transactional write path. The schema (`ai_models` / `labels` / `detections` / `detection_locks` / `detection_reviews`) is stable.
- **Correction calls `c.invalidateDetectionCache()`** so the 5-minute species-list cache reflects the new label immediately — same pattern as `DeleteDetection` / `ReviewDetection` / `LockDetection`. Without it the parent list would show the pre-correction species for up to 5 minutes after the user "confirmed".
- **`fetchWithCSRF` timeout is bumped to 120s for reanalyze only**. The default 30s aborts under contention with the realtime pipeline (4 streams × 2 models holding the per-model orchestrator mutexes). The endpoint completes in ~5s on an idle box and ≤30s under heavy realtime load; 120s is comfortable headroom.
- **Modal auto-run is wrapped in `untrack`** to break a Svelte 5 effect dependency loop: `runReanalysis()` transitively reads `isRunning`, so the original effect dep set included it; flipping `isRunning` back to false at the end of the fetch re-fired the effect, firing another fetch, looping forever (12 identical POSTs visible in DevTools before the fix).

## Bug also fixed: `clipName` missing on detection responses

While building this I noticed the Download button on the detection detail view was silently absent on every detection that actually had a clip — `DetectionResponse` never serialized `note.ClipName` even though the frontend type declared the field and the datastore had it. The Reanalyze button is gated on the same condition, which is how I caught it. Single-field fix included in commit `fix(api/v2): include clipName on detection responses`.

## Bonus: clickable date/time + status on `/ui/detections`

Each row's species name + thumbnail already linked to the detail page; the date/time and status cells didn't. Both now navigate to the detail view via the same `handleDetailsClick`, using real `<a>` tags so middle-click / right-click context menu work normally. Small change, on-topic for the same overall workflow.

## Tests

- 5 backend unit tests covering window aggregation (BirdNET-shape labels split into scientific+common), Perch-style bare scientific names (CommonName left empty so the handler can resolve via locale chain), short-clip padding, predict error propagation, and per-model no-truncation contract (top-N is the multi-model aggregator's job, not per-model).
- Decoder roundtrip test on a hand-rolled WAV fixture (skipped when ffmpeg is not in PATH).
- ffmpeg-path-empty guard test.
- Manual end-to-end on a real install: ran reanalyze on a 45s clip, confirmed per-model grid renders, applied correction, verified DB ground truth + API response + detection list parent view all updated correctly.

## i18n

New translation keys across all 8 supported locales (en/de/es/fi/fr/nl/pl/pt):

- `detections.reanalyze.title`, `descriptionMulti`, `multiModelSummary`, `colSpecies`, `running`, `noPredictions`, `useThis`, `useThisAriaLabel`, `noModelForCorrection`, `confirmTitle`, `confirmBody`, `confirmApply`, `applying`, `correctionApplied`
- `detections.reanalyze.button` / `buttonAriaLabel` (the Sparkles button on the metadata bar)
- `detections.row.openDetailFromStatus` (aria-label for the new clickable status badge)

`generate:i18n-types` was re-run; `types.generated.ts` is part of the diff.

## Test plan

- [x] `gofmt -l ./internal/...` — clean
- [x] `npm run format:check` / `npm run lint` — clean (the only ESLint warnings present are pre-existing on `MetricStrip.svelte` which I didn't touch)
- [x] Manual: reanalyze + correct on three real detections — BirdNET-only, Perch-only-vocab species (resolves via taxonomy CSV), and a low-confidence original
- [ ] Full `golangci-lint run -v` and `npm run check:all` — CI will run these (my local host can't link TFLite C headers)

Happy to split into smaller PRs if you'd prefer — natural seams would be `clipName fix` → `single-model reanalyze` → `multi-model + correction` → `row click UX`. Kept it as one to keep the change story coherent and because the smaller pieces don't really stand alone (multi-model needs the per-model schema decisions, correction needs reanalyze to feed it, the button is gated on clipName).

## Summary by CodeRabbit

* **New Features**
  * On-demand reanalysis from detection detail: a modal that runs reanalysis when opened, shows per-model species confidence grid, and offers an inline two-step "Use this" correction flow with confirmation, success feedback, and automatic refresh.
  * Date/time and status cells now deep-link into the detection details.

* **Localization**
  * Added reanalysis and detail-row translations in multiple languages (en, de, es, fi, fr, nl, pl, pt).

* **Tests**
  * New tests covering reanalysis aggregation and audio decoding.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3047)


jdevalk:
Addressed all four review comments. Pushed:

- **gemini (reanalyze.go:116)** — `ctx.Bind` error now returns 400 instead of being swallowed.
- **gemini (reanalyze.go:230)** — Models within each sample-rate group now run concurrently via `errgroup`. Distinct model IDs grab distinct orchestrator per-model mutexes, so parallelism is real; calls into the same model stay serialized by the orchestrator. Wall time roughly halves on a 2-model install and scales with model count. `gctx` propagation cancels in-flight work on request cancellation.
- **gemini (ReanalyzeModal.svelte:147)** — Replaced `window.location.reload()` with an `onCorrected` callback. The parent (`DetectionDetail`) wires it to its existing `fetchDetection()` — same pattern as `ReviewCard.onSaveComplete`.
- **sentry (reanalyze.go:371)** — Switched `clipLen` math to nanoseconds-based so fractional-second windows don't silently truncate. No-op today (all loaded models are integer-second), but future-proofs.

/gemini review

jdevalk:
Second-round review feedback addressed in 844dd633:

**Backend:**
- `clip_decode.go`: cancel ffmpeg before `cmd.Wait()` when the stdout byte cap is exceeded. Without this, hitting the cap left ffmpeg blocked on a full stdout pipe and Wait hung until the outer context timed out.
- `correct_species.go`: distinguish 404 / 409 / 400 / 500 instead of mapping every user-caused failure to 400. `is locked` → 409, missing rows → 404, vocab/loaded checks → 400. 5xx responses now use a generic message; the wrapped Go error stays in structured logs.
- `correct_species.go`: dropped the custom `containsCaseInsensitive` helper. Inlined as `strings.ToLower(err.Error()) + strings.Contains` against the marker list — stdlib, no ASCII-fastpath quirks.
- `reanalyze.go`: deduplicate `requestedIDs` after registry-ID resolution so e.g. `{"modelIds":["birdnet","BirdNET_V2.4"]}` doesn't schedule the same model twice.

**Frontend:**
- `ReanalyzeModal.svelte`: added a monotonic `requestSeq` guard. With the 120s reanalyze timeout, a user can open the modal, close it, and reopen for a different detection mid-fetch — the older fetch's result is now discarded instead of overwriting the newer state.
- `DetectionRow.svelte`: added `aria-label` to the date/time anchor + new `detections.row.openDetailFromDate` i18n key across all 8 locales.

**i18n:**
- `de.json`: switched my recently-added `detections.reanalyze.*` strings from informal "du/Klicke" to formal "Sie/Klicken Sie" to match the rest of the German locale. Other 7 locales were already idiomatic.

All 15 review threads resolved.

/gemini review

jdevalk:
Cleaned up the remaining review feedback in `e16d4418`. Quick summary for anyone scanning:

### What changed in this commit

**1. Atomic species + review write — sentry's `correction.go:72` finding.**
The v2 path used to do `detRepo.Update(...)` followed by `detRepo.SaveReview(...)` as two separate operations. A transient failure on `SaveReview` (DB timeout, connection drop) would leave the detection corrected to the new species but without the matching `verified='correct'` review.

Added `DetectionRepository.CorrectAndVerify(ctx, id, updates, review)` that wraps both writes in a single `r.db.Transaction(...)` via a tx-bound sibling repository. The atomic `NOT EXISTS` lock guard inside `Update` and the `OnConflict` upsert inside `SaveReview` stay in exactly one place each — the new method is just the transaction envelope. Both call sites (v2only `CorrectNoteSpecies` and `DualWriteRepository`'s v2 mirror) get atomicity through the shared `WriteSpeciesCorrection` helper without any change to their own code.

**2. Discriminate Get errors in the API handler — coderabbit's `correct_species.go:113` finding.**
`c.DS.Get(idStr)` was treating every error as 404. A transient DB error (connection drop, schema mismatch) was being reported to operators as "Detection not found". Now: `repository.ErrDetectionNotFound` or any `bnerrors.IsNotFound(err)` classification stays 404; anything else maps to 500 with a generic client message (full error stays in the structured log).

**3. Four atomicity regression tests** in `detection_impl_test.go`:
- `TestCorrectAndVerify_HappyPath` — both writes commit together.
- `TestCorrectAndVerify_LockedRollsBackReview` — when the lock guard inside `Update` fires, the review row never lands. This is the regression test for the bug sentry flagged.
- `TestCorrectAndVerify_NotFoundLeavesReviewAbsent` — same guarantee for the not-found case; no orphan reviews on missing detections.
- `TestCorrectAndVerify_IdempotentOnRerun` — rerun upserts cleanly, exactly one review row remains.

All four pass locally; the full `internal/datastore/v2/...` and `internal/datastore/v2only/` suites stay green.

### Threads I resolved

The six earlier sentry "Resolved in `<sha>`" threads (clip_decode.go:131, correct_species.go:208 / :162, interfaces.go:1376, v2only/datastore.go:1490, detection_repository.go) plus the two threads addressed in this commit (correction.go atomicity, correct_species.go sentinel). Pure UI hygiene on the sentry ones — the commits were already in.

### Threads I'm leaving open for your call

Three gemini suggestions I didn't apply — happy to if you'd prefer:

- **`reanalyze.go:111` / `correct_species.go:79` — "refresh config from global source at start of method"**: these are read-only endpoints that snapshot `c.Settings.BirdNET.Locale` once per request. The locale doesn't change mid-request, and re-fetching from the global source on every reanalyze/correct call adds a sync round-trip for no behavioural difference. Happy to wire in a per-method config refresh if you want strict adherence to the project's general rule — just let me know.

- **`ReanalyzeModal.svelte:212` — defensive `?? []` in `$derived.by`**: the `if (!result) return ''` guard at the top of the deriver already short-circuits when `result` is null/undefined, and `result.modelsRun` is non-nullable in the typed response. The `?? []` would only help if our backend type contract changed; I'd rather catch that with a runtime guard at the fetch boundary than scatter nullish-coalescing through derivers. Open to applying it if you'd prefer the defensive style.

### Where this leaves the PR

CI is re-running on `e16d4418`. All previous CI was green; this commit doesn't touch any of the orchestrator/decoder code paths the heavy tests cover, so I don't expect surprises.

@tphakala — at this point the only outstanding feedback is the three gemini stylistic suggestions above (which I documented my reasoning on rather than applying blind). Would love your eyes on this whenever you have a window; the feature description in the PR body has a reproducible example on my own install if you want to spin it up. Happy to split into the four-step stack I mentioned (clipName fix → single-model → multi-model + correction → row UX) if that helps the review, though my preference is still to keep it as one coherent change because the pieces don't really stand alone.

jdevalk:
Also pulled `main` into the branch in `f7e78e4b` — the PR went `CONFLICTING` while it sat for those 12 days, since this branch touches the i18n locale files and `DetectionRow.svelte` and both got significant new commits upstream (notably the row bulk-selection Checkbox + `StatusBadges` → `VerificationBadges` rename).

Resolution was clean: locale JSONs deep-merged so every key from both sides survives (verified the eight `messages/*.json` files still parse and that all the `reanalyze.*` keys plus the new upstream `row.play`/`row.playAudio` are present); `DetectionRow.svelte` kept the new Checkbox cell + the new component name while keeping this PR's anchor wrappers for the date/time and status cells; `mock_Interface.go` regenerated via the configured mockery setup (no manual edits). PR is back to `MERGEABLE` and CI is re-running.

tphakala:
Hi @jdevalk, a quick licensing note before this gets merged.

tphakala:

alexbelgium:
This is a great tool indeed to verify with multiple models the acceptability of a detection

PR #3054, opened 2026-05-13
fix(datastore/v2only): prefix-aware table names in 4 raw-SQL queries

jdevalk:
## Summary

Four queries in `internal/datastore/v2only/datastore.go` build raw SQL with hardcoded `"detections"` / `"labels"` / `"detection_reviews"` table names. On MySQL deployments that are still in the v1→v2 migration window the v2 schema lives behind a `v2_` prefix (see `TablePrefix()` on `MySQLManager`), so every one of these queries hits *table not found* — silently breaking:

- `GetTopBirdsData` — top-of-period species aggregator (~L1045)
- `GetBatchHourlyOccurrences` — batched hourly-occurrence fetch (~L1211)
- `DeleteClipPathsBatch` — clip-path cleanup batch (~L2121)
- `GetSpeciesDiversityData` — species diversity chart (~L2566)

The last one is what Sentry [flagged on #3020](https://github.com/tphakala/birdnet-go/pull/3020#discussion_r2549500000). The hardcoded name there predates #3020 and the same bug exists in the three siblings, so this PR fixes all four uniformly rather than patching just the one the bot caught.

## Fix shape

- Read `ds.manager.TablePrefix()` once at the top of each query.
- Concatenate the prefix onto every raw table reference: the `Table(...)` anchor and the `Joins(...)` clauses. Returns `""` on SQLite and on fresh-install MySQL where the v2 tables already use unprefixed names, so the existing happy path keeps working.

To avoid type-asserting the `Manager` interface at every call site, also expose `TablePrefix()` on `v2.Manager` itself. Both `SQLiteManager` and `MySQLManager` already implement it on their concrete types; this just lifts it onto the interface.

## Test plan

- [ ] CI: `go test ./...` (no behavior change on SQLite / fresh-install MySQL — prefix is `""`)
- [ ] Manual: on a v1→v2 migrated MySQL DB, hit `/api/v2/analytics/species/diversity`, `/api/v2/analytics/top-birds`, and the batched hourly-occurrences endpoint — confirm they return data instead of failing with *Table 'detections' doesn't exist*

## Summary by CodeRabbit

* **Refactor**
  * Updated internal datastore handling to consistently apply table prefixes so database queries behave correctly across different database setups and migration modes.
* **Tests**
  * Aligned test doubles with the updated datastore interface to preserve test accuracy.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3054)

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3054)

### DIFF
--- a/internal/api/v2/app_wizard_test.go
+++ b/internal/api/v2/app_wizard_test.go
@@ -65,6 +65,7 @@ func (f *fakeV2Manager) CheckpointWAL() error { return nil }
 func (f *fakeV2Manager) Delete() error        { return nil }
 func (f *fakeV2Manager) Exists() bool         { return true }
 func (f *fakeV2Manager) IsMySQL() bool        { return false }
+func (f *fakeV2Manager) TablePrefix() string  { return "" }
 
 // newFakeV2ManagerWithDetections creates a fakeV2Manager backed by an
 // in-memory SQLite database.  If count > 0, it inserts that many dummy
--- a/internal/datastore/v2/manager.go
+++ b/internal/datastore/v2/manager.go
@@ -55,6 +55,15 @@ type Manager interface {
 	Exists() bool
 	// IsMySQL returns true if this is a MySQL manager.
 	IsMySQL() bool
+	// TablePrefix returns the prefix applied to every v2 table name. It is
+	// non-empty (currently "v2_") only on MySQL deployments that are still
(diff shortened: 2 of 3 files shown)


tphakala:
Thank you for contributing!

PR #2066, opened 2026-02-26
feat(audio): add per-source quiet hours scheduling

iamrans0m00:
## Summary

Add configurable quiet hours that stop/start individual audio streams and the sound card during time windows when birds are inactive, reducing CPU usage. Supports two modes:

- **Fixed**: Clock-based HH:MM start/end times (handles overnight windows like 22:00→06:00)
- **Solar**: Relative to sunrise/sunset with configurable minute offsets (±3 hours)

### Backend
- `QuietHoursConfig` struct added to `StreamConfig` and `AudioSettings` in config
- `QuietHoursScheduler` evaluates every minute, stops/starts streams and sound card via existing manager/control channel infrastructure
- Validation for modes, time formats, solar events, and offset ranges
- `GET /api/v2/streams/quiet-hours/status` endpoint with privacy-sanitized stream URLs
- Settings change detection triggers immediate scheduler re-evaluation
- Named constants for all mode/event strings (no magic strings)

### Frontend
- Reusable `QuietHoursEditor` component with fixed time inputs and solar event selectors
- Per-stream quiet hours configuration in StreamCard edit mode
- Sound card quiet hours in AudioSettingsPage Sound Card tab
- `QuietHoursIndicator` moon icon in header when any source is suppressed
- Stream cards show blue "Quiet Hours" status pill instead of "unknown" when suppressed
- i18n support across all 10 languages (en, de, es, fi, fr, it, nl, pl, pt, sk)

### Design Decisions
- Scheduler reads config each tick (1 min) — no explicit reload needed for most changes
- Tracks suppressed state internally — only restarts streams it stopped, never touches user-managed streams
- Sound card stop/start via existing control monitor channel pattern
- Backward compatible — `QuietHoursConfig` defaults to `{enabled: false}`, no migration needed

## Testing

- [x] Go unit tests for quiet hours scheduler (fixed mode, solar mode, overnight windows, edge cases)
- [x] Go tests pass: `go test -race -v ./internal/myaudio/...` and `./internal/conf/...`
- [x] Go packages build clean (conf, myaudio, api/v2, analysis)
- [x] ESLint: 0 errors, 0 warnings
- [x] TypeScript: 0 errors, 0 warnings (svelte-check, 4553 files)
- [x] Prettier: all files formatted
- [x] Manual testing: streams stop/restart during quiet hours, status API returns correct state, UI shows suppressed status pill, header indicator appears

## Screenshots

[...]

PR #3127, opened 2026-05-17
fix(audiocore): wire stream manager and fix source ID mismatch in quiet hours

iamrans0m00:
## Summary

Fixes a regression introduced by the audiocore refactor (PR #2489) that silently broke quiet hours enforcement for RTSP streams. The original PR #2066 was working when it was opened, but the refactor introduced two separate wiring bugs that together caused the scheduler to have no effect.

### Root cause

1. **`SetScheduler` never wired the stream manager**: The `SetStreamManager` call was missing, so the scheduler's `getManager` func was always nil. At evaluation time the scheduler bailed out early with "stream manager not available" and did nothing.

2. **`New()` bypassed `SetScheduler`**: When a scheduler was passed to `New()`, it called `e.scheduler.Store(scheduler)` directly, skipping `SetScheduler` entirely, so the wiring fix in `SetScheduler` would not apply to schedulers set at construction time.

3. **Source ID mismatch**: The scheduler used `stream.URL` as the source ID to call `StopStream`/`StartStream`, but the engine's `SourceRegistry` generates hashed IDs (`rtsp_<hex>`, e.g. `rtsp_a1b2c3d4`). The stop/start calls were targeting IDs that never existed.

### Fix

- **`engine.go`**: `SetScheduler` now calls `s.SetStreamManager(...)` after storing the scheduler. `New()` now delegates to `SetScheduler` instead of calling `e.scheduler.Store` directly.
- **`quiet_hours.go`**: Added `GetActiveStreamURLs() map[string]string` to the `StreamManager` interface (returns URL→sourceID map). The scheduler now builds a reverse lookup at evaluation time and uses the real runtime source IDs when calling `StopStream`/`StartStream`. A `suppressedURLs` map tracks which URLs were stopped (and their runtime IDs) so they can be correctly restarted.
- **`quiet_hours_test.go`**: Updated `mockManager` to implement the new interface and all test cases use realistic hashed source IDs.
- **`engine.go`**: Replaced magic number `1` with named constant `defaultChannels` per project conventions.

## Test plan

- [ ] Existing unit tests pass: `go test ./internal/audiocore/...`
- [ ] Quiet hours stop/start verified manually with an RTSP stream configured
- [ ] `golangci-lint run -v` clean on changed files

## Summary by CodeRabbit

* **New Features**
  * Start and stop individual streams at runtime from the service.
  * Query active stream IDs and their connection URLs.

* **Improvements**
  * Quiet-hours management now tracks suppression by runtime stream identity and reliably restarts the same runtime stream.
  * GetSuppressedStreams() reporting improved to return suppression keyed by stream URL (with legacy fallback), and a new query lets callers check suppression by runtime stream ID.

* **Tests**
  * Quiet-hours behavior covered by updated tests reflecting runtime IDs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3127?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/analysis/audio_pipeline_service.go
+++ b/internal/analysis/audio_pipeline_service.go
@@ -300,8 +300,7 @@ func (p *AudioPipelineService) Start(_ context.Context) error {
 			if ok && src.Type == audiocore.SourceTypeAudioCard {
 				return p.quietHoursScheduler.IsSoundCardSuppressed()
 			}
-			suppressed := p.quietHoursScheduler.GetSuppressedStreams()
-			return suppressed[sourceID]
+			return p.quietHoursScheduler.IsStreamSuppressed(sourceID)
 		},
 	}
 	p.watchdog = audiocore.NewLivenessWatchdog(
--- a/internal/audiocore/engine/engine.go
+++ b/internal/audiocore/engine/engine.go
@@ -232,6 +232,76 @@ func (e *AudioEngine) Scheduler() *schedule.QuietHoursScheduler {
 // resources (SunCalc, ControlChan) only available after service startup.
 func (e *AudioEngine) SetScheduler(s *schedule.QuietHoursScheduler) {
 	e.scheduler.Store(s)
+	if s != nil {
+		s.SetStreamManager(func() schedule.StreamManager { return e })
+	}
+}
+
+// GetActiveStreamIDs returns the runtime source IDs currently tracked by FFmpeg.
+func (e *AudioEngine) GetActiveStreamIDs() []string {
+	return e.ffmpegMgr.GetActiveStreamIDs()
+}
+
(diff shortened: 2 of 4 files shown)


tphakala:
Thanks for this fix! The quiet hours sourceID mismatch was a real bug that completely prevented stream suppression from working. Good catch on the `SetScheduler` bypass in `New()` too.

I rebased on main and added one extra commit: `GetSuppressedStreams()` now returns URL-keyed maps (for the API), but the liveness watchdog (merged after your branch diverged in #3106) was indexing by runtime sourceID. Added `IsStreamSuppressed(sourceID)` to handle that lookup correctly.

PR #3196, opened 2026-05-20
fix(SettingSave): fix exclude list settings save failure

ModerateWinGuy:
## Summary

- Fix ignored species silently re-appearing in detections after the ONNX geomodel and Perch model were introduced
- Fix `GetExcludedSpecies` reading from a stale settings pointer under an incorrect write lock
- Remove now-unused `speciesExcludeMutex` field from the API controller

## Root Cause

Every time the range filter rebuilds its species list (on startup and daily), `UpdateIncludedSpecies` publishes a brand-new settings snapshot via `conf.StoreSettings(newSnapshot)`. After that call, the API controller's `c.Settings` field is **stale** — it still points to the old settings object, not the live one.

`toggleSpeciesInIgnoredList` and `addSpeciesToIgnoredList` were reading `settings := c.Settings` (the stale pointer), mutating the exclude list **in-place** on that stale object, then calling `conf.SaveSettings()`. Because `conf.SaveSettings()` reads from the **live** atomic pointer (not the mutated stale one), it wrote the old exclude list to disk. The processor also reads from the live atomic pointer, so it never saw the in-place change either. The ignore was silently discarded both in-memory and on disk — every single time.

This regression was introduced alongside the ONNX geomodel because that was the first time `UpdateIncludedSpecies` ran regularly (on startup + daily), causing `c.Settings` to diverge from `conf.GetSettings()` shortly after boot. Before the geomodel, the atomic pointer was rarely replaced, so `c.Settings` happened to stay in sync and the bug was latent.

The main settings save path (`UpdateSettings`) already handled this correctly — its `getSettingsOrFallback()` helper explicitly reads from `conf.GetSettings()` to avoid "a stale `c.Settings` pointer being overwritten by out-of-band `StoreSettings` calls (range filter, etc.)". The ignore-species functions were never updated to follow the same pattern.

## Fix

Replaced the in-place mutation pattern in `toggleSpeciesInIgnoredList` and `addSpeciesToIgnoredList` with the same **clone-mutate-publish** approach used by the main settings save path:

1. Acquire `c.settingsMutex.Lock()` (same mutex as `UpdateSettings`) to serialise against concurrent saves
2. Read the live snapshot via `c.getSettingsOrFallback()` instead of `c.Settings`
3. Deep-clone via `conf.CloneSettings(current)`
4. Modify the exclude list on the clone
5. Publish via `conf.StoreSettings(updated)` so the atomic pointer is updated immediately
6. Keep `c.Settings = updated` in sync
7. Persist to disk via `conf.SaveSettings()` (which now reads the correctly updated snapshot)
8. Call `c.handleSettingsChanges(current, updated)` to trigger an immediate range filter rebuild, so the species include list also reflects the change without waiting for the next daily cycle

Also fixed `GetExcludedSpecies` to read from `c.getSettingsOrFallback()` under `c.settingsMutex.RLock()` instead of the stale `c.Settings` under a write lock, and removed the now-unused `speciesExcludeMutex` field from the controller struct.

## Summary by CodeRabbit

* **Refactor**
  * Unified settings update flow to use consistent snapshots and cloned updates for safer applies.
  * Consolidated concurrency handling for excluded-species management under a single settings lock.

* **Bug Fixes**
  * Settings changes now reliably trigger dependent operations; persistence failures roll back in-memory state.
  * TLS and MQTT certificate operations (upload/delete/generate) are handled more safely and consistently.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3196?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/api/v2/api.go
+++ b/internal/api/v2/api.go
@@ -58,8 +58,7 @@ type Controller struct {
 	TaxonomyDB          *classifier.TaxonomyDatabase
 	controlChan         chan string
 	shutdownRequester   ShutdownRequester // programmatic shutdown trigger (e.g., for restart)
-	shutdownMu          sync.RWMutex      // protects shutdownRequester
-	speciesExcludeMutex sync.RWMutex      // Mutex for species exclude list operations
+	shutdownMu sync.RWMutex // protects shutdownRequester
 	// DisableSaveSettings prevents persisting settings changes to disk.
 	// When set to true, all settings modifications remain in memory only.
 	// This is primarily used in testing but can be used in production for read-only mode.
--- a/internal/api/v2/detections.go
+++ b/internal/api/v2/detections.go
@@ -1543,13 +1543,9 @@ func (c *Controller) IgnoreSpecies(ctx echo.Context) error {
 
 // GetExcludedSpecies returns the list of excluded species
 func (c *Controller) GetExcludedSpecies(ctx echo.Context) error {
-	settings := c.Settings
-
-	// Create a copy of the slice to avoid race conditions
-	c.speciesExcludeMutex.Lock()
-	species := make([]string, len(settings.Realtime.Species.Exclude))
-	copy(species, settings.Realtime.Species.Exclude)
-	c.speciesExcludeMutex.Unlock()
+	c.settingsMutex.RLock()
+	species := slices.Clone(c.getSettingsOrFallback().Realtime.Species.Exclude)
+	c.settingsMutex.RUnlock()
 
 	return ctx.JSON(http.StatusOK, ExcludedSpeciesResponse{
 		Species: species,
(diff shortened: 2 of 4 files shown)


tphakala:
Hi @ModerateWinGuy, a licensing note about your contribution here.

I'm planning to relicense BirdNET-Go from CC BY-NC-SA 4.0 to AGPL-3.0, an OSI-approved open source license. It would keep the copyleft, share-alike spirit but remove the non-commercial restriction and make the project properly open source. There are no plans to commercialize the project; this is to align better with the open source community and make packaging easier. Full context is in #4243, and consent is tracked in #4244.

Since your work is part of the codebase, I'd like your agreement before relicensing. Could you reply confirming:

Thanks a lot for this contribution.

ModerateWinGuy:
Sure. 👍 
I agree that my contributions to birdnet-go, past and future, may be licensed under AGPL-3.0.

PR #3276, opened 2026-05-26
fix(classifier): apply range filter to Perch detections

keithkml:
## Summary

Perch detections were bypassing the geographic range filter unless the range filter model was explicitly set to `v3`. Live config can use the default `latest` range model while Perch is enabled, which lets out-of-range Perch labels pass through to notifications.

This changes range-filter eligibility so Perch uses the same included-species scientific-name lookup as BirdNET whenever location filtering is configured. Bat/BSG remain excluded from the bird range filter path.

## Fix

- Apply the range filter to `Perch_V2` detections when `LocationConfigured` is true, independent of the configured range model name.
- Keep range filtering disabled when no location is configured.
- Add regression coverage for:
  - Perch with `latest` range model filtering an out-of-range scientific-name label.
  - Perch allowing an in-range scientific-name label.
  - Bat/BSG staying outside the range-filter path.

## Live Smoke Test

I checked a live Raspberry Pi deployment without replacing the running service:

- Confirmed Perch and BirdNET were both enabled with location filtering configured and `rangefilter.model: latest`.
- Confirmed the live range command, using the deployed config, produced a small local included-species set.
- Compared recent high-confidence Perch detections against that live included-species set and found out-of-range Perch labels that the current deployed build had accepted. This reproduces the reported notification behavior and matches the fixed code path.

## Preflight Status

Preflight review passes completed:

- Scope: one bug fix only; two Go files changed.
- Reuse/efficiency: uses existing `DetectionModelInfoForID` and `Settings.IsSpeciesIncluded` fast scientific-name lookup.
- Correctness: reviewer pass caught the need to guard on `LocationConfigured`; fixed before final checks.
- Quality/maintainability: no new abstraction, no config/API changes.
- i18n: not applicable, no user-facing text changed.
- Integration wiring: no endpoint/event/config schema changes.

Verification commands run and passed:

- `git diff --check`
- `golangci-lint run -v` with local CGO TFLite headers/libs configured
- `npm run check:all` under Node 24.14.0 from the Codex bundled runtime; exited 0 with three existing unrelated ESLint warnings
- `npm test` under Node 24.14.0
- `npm run build` under Node 24.14.0 to populate embedded frontend assets for Go checks
- `go test -tags noembed,skipfrontend ./internal/analysis/processor -run 'TestShouldApplyRangeFilter|TestShouldFilterDetection_AppliesRangeFilterToPerch'`
- `go test -race ./...` with local test environment settings: `GOFLAGS=-parallel=1`, `TMPDIR=/private/tmp`, `FFMPEG_PATH=/opt/homebrew/bin/ffmpeg`, plus local TFLite CGO headers/libs

Notes on local environment:

- The first exact full race run exposed local environment issues unrelated to this patch: sandbox-blocked `httptest` ports, missing absolute FFmpeg path, macOS `/var` symlink behavior in a mount test, and classifier memory-test noise under parallel `-race` execution.
- After setting the local environment values above, the full race suite passed.

## Breaking Changes

None. This only restores expected range-filter behavior for Perch detections when location filtering is configured.

## Bug Fixes
- Improved range filtering logic to properly account for location configuration settings
- Enhanced how different detection models interact with range-based filtering for more accurate results

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3276?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/analysis/processor/processor.go
+++ b/internal/analysis/processor/processor.go
@@ -950,17 +950,18 @@ func (p *Processor) parseAndValidateSpecies(settings *conf.Settings, result data
 
 // shouldApplyRangeFilter returns true if the given model should have its
 // detections filtered by the geographic range filter.
-// BirdNET (any version) and unknown models: always filtered (DetectionModelInfoForID
-// returns DefaultModelInfo with Name="BirdNET" for unrecognized IDs).
-// Perch: filtered only when the v3.0 geomodel is active (it covers Perch species).
+// BirdNET (any version), Perch, and unknown models are filtered. Perch returns
+// scientific-name labels, and the included-species set stores scientific names
+// for O(1) lookup, so the normal range list applies even when the active range
+// model is the embedded BirdNET geomodel rather than v3.
 // Bat/BSG: never filtered (independent species sets, no geomodel coverage).
 func shouldApplyRangeFilter(modelID string, settings *conf.Settings) bool {
-	mInfo := classifier.DetectionModelInfoForID(modelID)
-	if mInfo.Name == detection.DefaultModelName {
-		return true
+	if settings == nil || !settings.BirdNET.LocationConfigured {
+		return false
 	}
-	if mInfo.Name == classifier.DetectionNamePerch {
-		return settings.BirdNET.RangeFilter.Model == "v3"
+	mInfo := classifier.DetectionModelInfoForID(modelID)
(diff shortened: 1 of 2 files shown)


tphakala:
Thank you

PR #3280, opened 2026-05-27
feat(ui): respect global threshold in recent detections

keithkml:
## Summary

- Keep `/api/v2/detections/recent` backwards-compatible by returning raw recent detections by default.
- Add `aboveThreshold=true` as an opt-in query parameter that filters recent detections by the configured global BirdNET confidence threshold.
- Update the dashboard's recent detections request to use `aboveThreshold=true`, so the dashboard excludes detections below the global confidence threshold.
- Add legacy datastore and v2-only datastore/repository paths for retrieving recent detections above a minimum confidence.
- Cover the default and threshold-filtered recent-detection paths with API, legacy datastore, and v2-only datastore tests, and update endpoint docs.

## Root Cause

The dashboard's recent detections widget used the raw `/api/v2/detections/recent` endpoint, so it could show detections below the global threshold even though alerting ignores detections below that threshold. The first version changed the endpoint default; this version keeps the API default stable and makes the dashboard opt into threshold filtering.

## Behavior Change

`GET /api/v2/detections/recent` keeps its existing default behavior.

Clients that want global-threshold filtering can call:

```text
/api/v2/detections/recent?aboveThreshold=true
```

When `aboveThreshold=true`, results include only detections with `confidence >= BirdNET.Threshold`. If settings are unavailable, the threshold floor falls back to `0`.

## Review Feedback Addressed

- Pin the in-memory SQLite datastore test to a single connection and close the DB in `t.Cleanup`.
- Log the actual retrieved detection count and requested limit separately.
- Replace threshold/count literals in the new datastore tests with named constants.
- Tighten dashboard endpoint docs so they describe the global-threshold behavior without overstating notification semantics.
- Add doc comments for the new threshold helper/test paths flagged by CodeRabbit's docstring coverage warning.

## Preflight Status

- [x] Single concern: one user-facing dashboard behavior change, with API/datastore support kept to that behavior.
- [x] Preflight Phase 1-3 review completed; no additional in-scope findings remain.
- [x] No unrelated PR diff: final PR diff is 16 files, all tied to dashboard recent-detections threshold behavior or required interface/test plumbing.
- [x] Scope complete: no TODO/FIXME added for core functionality.
- [x] Breaking changes documented: no backwards-incompatible API default change; opt-in behavior is documented here and in endpoint docs.
- [x] i18n complete: no user-facing frontend strings were added.
- [x] No secrets or PII in the diff.
- [x] `git diff --check` passed.
- [x] `golangci-lint run -v` passed with local TensorFlow Lite CGO paths: 0 issues.
- [x] `npm run check:all` passed; it still reports 3 pre-existing ESLint security warnings in unchanged frontend files.
- [x] `npm test` passed: 107 files, 1968 tests.
- [x] Focused Go tests passed outside the sandbox:
  - `go test -tags noembed,skipfrontend ./internal/datastore ./internal/datastore/v2only ./internal/datastore/v2/repository ./internal/datastore/v2/migration/testutil`
  - `go test -tags noembed,skipfrontend ./internal/api/v2 -run 'TestGetRecentDetections|TestGetRecentDetections_DefaultParams' -count=1 -v`
- [x] In-scope Go race tests passed outside the sandbox before the doc-comment-only follow-up:
  - `go test -race -tags noembed,skipfrontend ./internal/datastore ./internal/datastore/v2only ./internal/datastore/v2/repository ./internal/datastore/v2/migration/testutil ./internal/api/v2`
- [ ] `go test -tags noembed,skipfrontend ./internal/api/v2 -count=1` is not green locally due to unrelated date-sensitive weather test behavior:
  - `internal/api/v2`: `TestGetHourlyWeatherForDayFutureDate` expected `No weather data available for future date` but received `No weather data found for the specified date`.
- [ ] `go test -race ./...` was run outside the sandbox with TensorFlow Lite CGO paths and is not green locally due to unrelated macOS/environment-sensitive failures:
  - `internal/birdweather`: `TestEncodeFlacUsingFFmpeg` expects `ffmpeg` to be accepted, but local validation requires an absolute FFmpeg path.
  - `internal/monitor`: `TestGroupPathsWithPartitions_MixedRootAndVolumePaths` expected Linux mount data (`/dev/sda1`) that is unavailable on this macOS runner.

## Summary by CodeRabbit

* **New Features**
  * Optional filtering for recent detections by the global confidence threshold; dashboard shows detections meeting or exceeding the configured BirdNET threshold.

* **Documentation**
  * API docs updated to describe the new above-threshold query parameter and example usage.

* **Tests**
  * Added and updated tests to cover threshold-filtered recent detections and default limit behavior.

* **Refactor**
  * Detection retrieval logic consolidated to support confidence-based filtering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3280?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/lib/desktop/features/dashboard/API_ENDPOINTS.md
+++ b/frontend/src/lib/desktop/features/dashboard/API_ENDPOINTS.md
@@ -42,6 +42,9 @@ const dailySummary = await response.json();
 **Query Parameters:**
 
 - `limit` (optional): Number of recent detections to return, default: 10
+- `aboveThreshold` (optional): When `true`, only return detections at or above the global BirdNET confidence threshold
+
+By default, this endpoint returns the latest detections without threshold filtering. The dashboard uses `aboveThreshold=true` so recent detections match notification-worthy detections.
 
 **Response:**
 
@@ -68,7 +71,7 @@ const dailySummary = await response.json();
 **Usage in Svelte:**
 
 ```typescript
-const response = await fetch('/api/v2/detections/recent?limit=10');
+const response = await fetch('/api/v2/detections/recent?limit=10&aboveThreshold=true');
 const recentDetections = await response.json();
 ```
 
--- a/frontend/src/lib/desktop/features/dashboard/API_ENDPOINTS.md
+++ b/frontend/src/lib/desktop/features/dashboard/API_ENDPOINTS.md
@@ -44,7 +44,7 @@ const dailySummary = await response.json();
 - `limit` (optional): Number of recent detections to return, default: 10
 - `aboveThreshold` (optional): When `true`, only return detections at or above the global BirdNET confidence threshold
 
-By default, this endpoint returns the latest detections without threshold filtering. The dashboard uses `aboveThreshold=true` so recent detections match notification-worthy detections.
+By default, this endpoint returns the latest detections without threshold filtering. The dashboard uses `aboveThreshold=true` to exclude detections below the global confidence threshold.
 
 **Response:**
 
--- a/frontend/src/lib/desktop/features/dashboard/pages/DashboardPage.svelte
+++ b/frontend/src/lib/desktop/features/dashboard/pages/DashboardPage.svelte
@@ -456,7 +456,7 @@ Performance Optimizations:
     try {
       const response = await fetch(
         buildAppUrl(
-          `/api/v2/detections/recent?limit=${Math.max(detectionLimit, MIN_FETCH_LIMIT)}&includeWeather=true`
+          `/api/v2/detections/recent?limit=${Math.max(detectionLimit, MIN_FETCH_LIMIT)}&includeWeather=true&aboveThreshold=true`
         )
       );
       if (!response.ok) {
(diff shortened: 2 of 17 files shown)


tphakala:
Thanks for putting this together, but I'm going to pass on it.

This changes behavior that's working as intended. BirdNET-Go has a dynamic threshold feature that deliberately lowers the confidence threshold for species it has already seen with high confidence, so it keeps picking them up at lower confidence afterwards. Those lower-confidence detections are saved on purpose. Filtering recent detections by the global threshold would drop exactly those detections from the list, which defeats the whole point of dynamic thresholds.

So this isn't something I want to merge. Appreciate the effort regardless.

keithkml:
Thanks for considering it! I do not want to push on this, and am happy to drop it. I'll just ask some advice before I let it go :) 

I find "recent detections" not very useful because it's mainly low-confidence false positives. Any other ideas for how this tile could be work better for me, if not this PR?

PR #3282, opened 2026-05-27
fix(audio): boost quiet normalized exports

keithkml:
## Summary

Saved detection clips can remain almost inaudible when export normalization is enabled and the input audio is below FFmpeg `loudnorm`'s EBU R128 gate. In that case `loudnorm` reports invalid measured loudness such as `input_i=-inf` and the exported clip is not boosted to an audible level. Spectrograms rendered from those saved clips can then appear nearly black compared with clips that were loud enough for normal `loudnorm` processing.

This PR keeps the existing FFmpeg-based normalization path, but adds a bounded fallback for the gated-quiet case.

## Changes

- Run the existing PCM loudness analysis before normalized exports.
- Keep valid measurements on the existing single-pass `loudnorm` export path.
- When `loudnorm` returns invalid measured stats but a finite true peak, export with `loudnorm=...:offset=...` using a clamped offset derived from the true-peak target. This keeps the fallback inside `loudnorm` and its limiter instead of using a separate gain filter.
- Write the loudness-analysis temp WAV with the export PCM channel count, so stereo/interleaved inputs are analyzed with the same layout used by the actual export.
- Pin normalized export output back to the source sample rate so `loudnorm` does not leave saved FLAC clips at 192 kHz.
- Add regression coverage for very quiet normalized exports staying audible, configured sample-rate retention, per-phase export timeouts, and channel-aware WAV analysis input.

## Background

The issue was observed on a Raspberry Pi 5 BirdNET-Go deployment running both BirdNET and Perch on the same audio source. The quiet exported clips were not specific to either model; the common failure mode was FFmpeg `loudnorm` treating very low-level audio as below-gate silence.

This supersedes #3279, which focused on spectrogram brightness. The spectrogram symptom is downstream of the saved-clip audio level: fixing export normalization addresses the underlying audio issue instead of only changing image rendering.

## Before / After

Before/after audio export normalization

Image layout: left is the current saved quiet clip, right is the same sample after the `loudnorm offset` fallback; top is spectrogram, bottom is waveform.

Measured on the quiet sample used for the image:

- Before: mean `-80.3 dB`, max `-65.2 dB`
- After: mean `-20.3 dB`, max `-5.7 dB`, sample rate `48000 Hz`

## Preflight Status

- Scope check: one bug fix, limited to normalized audio export behavior plus one before/after image asset.
- Manual preflight review: completed reuse, correctness/safety, quality/patterns, i18n, and integration wiring passes. The current update addresses review feedback that the loudness-analysis temp WAV must preserve the export channel count; no in-scope findings remain.
- Linters: `golangci-lint run -v --allow-parallel-runners` passed with `0 issues`; `npm run check:all` passed. `npm run check:all` still reports three pre-existing frontend warnings in unrelated files (`VerificationBadges.svelte`, `MetricStrip.svelte`).
- Tests passed: `go test ./internal/audiocore/convert ./internal/audiocore/ffmpeg`; `go test ./internal/audiocore/ffmpeg -run 'TestExportAudio_Normalization|TestExportAudio_NormalizationBoostsGatedQuietAudio|TestExportPhaseTimeout|TestLoudnormGateFallbackOffset_NilStats' -count=1 -v`; `go test -race ./internal/audiocore/ffmpeg`; `npm test` (`107` files, `1968` tests). `git diff --check` also passed.
- Required aggregate race gate: `go test -race ./...` was run in the sandbox and rerun outside the sandbox with the TensorFlow Lite rpath and `FFMPEG_PATH=/opt/homebrew/bin/ffmpeg` set for this macOS environment. The final rerun fails only in unrelated `internal/monitor TestGroupPathsWithPartitions_MixedRootAndVolumePaths`, which expects `/dev/sda1` but gets an empty device on this macOS host. The touched audio packages pass under the race run.
- Diff hygiene: no unrelated files, no TODO/FIXME for core functionality, no secrets or PII, and no API/config breaking changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio normalization to boost quiet or near-silent audio during export.
  * More consistent output sample rates when normalization is enabled.
  * More reliable exports with bounded processing time to reduce hangs/failures.

* **Tests**
  * Added tests verifying quiet-audio normalization and export timing/timeout behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3282?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tphakala:
Thanks for digging into this, Keith. Your diagnosis is spot on: when the input sits below loudnorm's EBU R128 gate, loudnorm reports `input_i=-inf` and the clip never gets boosted, so the saved file stays inaudible and the spectrogram renders nearly black.

The gated-quiet fix has already landed on `main` in #3367, which was built on the same approach as this PR. Current `main` carries the same pieces:

- the offset fallback for the gated case (`loudnormGateFallbackOffset` / `buildLoudnormOffsetFilter`, exporting with `loudnorm=...:offset=...`),
- pinning the normalized output back to the source sample rate so FLAC clips don't get left at 192 kHz,
- the same audibility regression test (`TestExportAudio_NormalizationBoostsGatedQuietAudio`).

#3367 went a bit further and switched the normal valid-measurement path to two-pass linear `loudnorm` (`linear=true` with the measured stats), which also fixes the tail-pumping artifacts from #3366.

Because of that I'm closing this as superseded by #3367. Two practical reasons beyond the duplication: the branch is about a week behind `main` now and conflicts on these files, and merging it would swap the two-pass linear path back to single-pass dynamic loudnorm, which would reintroduce the pumping. I'm also keeping export mono-only, so the per-channel WAV analysis here isn't something I want to carry.

Really appreciate the writeup and the before/after measurements; they made it easy to confirm `main` now handles this case. If you build from a current `main` and still hit quiet clips staying inaudible, drop a note here and I'll take another look.

PR #3305, opened 2026-05-29
fix: stop imageprovider retrying negative cache entries

keithkml:
## Summary

- Limit v2-only image cache warm-up species to labels that are actually referenced by detections, deduped by scientific name.
- Skip `__NOT_FOUND__` negative cache entries during the hourly imageprovider background refresh.
- Add focused regressions for orphan/prediction-only labels and stale negative refresh selection.

## Root Cause

The v2-only `GetAllDetectedSpecies()` implementation read every species label row, including labels created by image-cache saves for names that never had detections. Separately, the hourly refresh path treated expired negative cache rows as refresh candidates, so background refreshes could keep retrying known misses against Wikipedia/Wikimedia and trigger repeated 429 errors.

Direct user-triggered lookup behavior is preserved: negative entries can still expire and retry through the existing direct cache lookup path. The change only prevents the hourly background job from hammering providers for known misses.

## Preflight Status

- [x] Single concern: one bug fix for imageprovider background 429 spam.
- [x] Preflight passed: manual five-pass preflight review completed; one in-scope linter finding fixed (`sort.Strings` -> `slices.Sort` in new code); no tracking issues filed.
- [x] Linters pass: `/Users/keith/go/bin/golangci-lint run -v` passed with project CGO/TFLite env; `npm run check:all` exited 0. Note: frontend ESLint still reports 3 pre-existing warnings in files untouched by this PR.
- [x] Tests pass: `go test -race ./...` passed after building `frontend/dist`; `CI=true npm test` passed.
- [x] No unrelated changes: diff is limited to v2-only detected-species selection, imageprovider stale refresh selection, and matching tests.
- [x] Scope complete: no TODO/FIXME added for core functionality.
- [x] Breaking changes documented: no API, config, or behavior-breaking changes.
- [x] i18n complete: no user-facing strings added.
- [x] No secrets or PII: diff contains no credentials or personal data.

## Validation

- `git diff --check`
- `npm ci`
- `npm run check:all`
- `CI=true npm test`
- `npm run build`
- `/Users/keith/go/bin/golangci-lint run -v`
- `go test -race ./...`

The initial sandboxed `go test -race ./...` attempt failed because `frontend/dist` was not present and the sandbox blocked `httptest` listener setup. After building `frontend/dist`, the final full race suite passed outside the sandbox with the same project CGO/TFLite environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Species detection now returns only species found in actual detections, preventing unrelated labels from appearing in results
  * Image provider no longer unnecessarily re-queries for previously unfound species during background refresh cycles

* **Tests**
  * Added regression test verifying species filtering accuracy
  * Added test for negative cache entry handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3305?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/datastore/v2only/datastore.go
+++ b/internal/datastore/v2only/datastore.go
@@ -1352,26 +1352,44 @@ func (ds *Datastore) GetLastDetections(numDetections int) ([]datastore.Note, err
 // GetAllDetectedSpecies retrieves all detected species.
 func (ds *Datastore) GetAllDetectedSpecies() ([]datastore.Note, error) {
 	ctx := context.Background()
-	labels, err := ds.label.GetAllByLabelType(ctx, ds.speciesLabelTypeID)
+	labelIDs, err := ds.detection.GetAllDetectedLabels(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if len(labelIDs) == 0 {
+		return []datastore.Note{}, nil
+	}
+
+	labels, err := ds.label.GetByIDs(ctx, labelIDs)
 	if err != nil {
 		return nil, err
 	}
 
-	// Use a map to deduplicate by scientific name (since labels are now per-model).
+	// Use a map to deduplicate by scientific name (since labels are per-model).
 	// Labels may contain legacy concatenated "ScientificName_CommonName" format,
 	// so extract only the scientific name portion.
 	seen := make(map[string]struct{}, len(labels))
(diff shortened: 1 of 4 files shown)


tphakala:
Thank you!

PR #3313, opened 2026-05-30
chore(frontend): replace svelte-portal with the local portal action

tphakala:
## What

Removes the `svelte-portal` runtime dependency, continuing the frontend supply-chain reduction effort.

A local `portal` action already lives in `src/lib/utils/portal.ts` (re-parents a node to a target, default `document.body`, on mount; removes it on destroy) and is already used by `SelectDropdown.svelte`. The only `svelte-portal` consumer, `BirdThumbnailPopup.svelte`, now uses that same local action, so no new action code was needed.

- **`BirdThumbnailPopup.svelte`**: swap `import Portal from 'svelte-portal'` for `import { portal } from '$lib/utils/portal'`, and apply `use:portal` to the `#bird-popup` element instead of wrapping it in `<Portal>`. The popup is `position: fixed` with viewport coordinates and was already portaled to `document.body`, so DOM topology and visible behavior are unchanged; this only drops svelte-portal's redundant wrapper `<div>`.
- **`portal.test.ts`** (new): the action had no test coverage and is now used in two places. Covers mount re-parenting, destroy removal, explicit targets, and the "only remove if still attached to the original target" guard.
- **`package.json`**: remove `svelte-portal`.

## Why

Fewer external npm runtime dependencies means less supply-chain attack surface. This replaces a third-party component with an existing, already-vendored local action.

## Note on the diff size

The large `BirdThumbnailPopup.svelte` diff is almost entirely Prettier reindenting the popup block after one level of nesting (`<Portal>`) was removed. The functional change is three lines: the import, the `use:portal` directive, and dropping the wrapper.

## Verification

- `npm run check:all` passes (typecheck 0 errors / 0 warnings).
- `npm run build` succeeds.
- `npm run test:ci` passes: 1972 tests across 108 files, including the 4 new portal-action tests.
- No `svelte-portal` references remain in `src/` or `package.json`.

## Summary by CodeRabbit

* **Refactor**
  * Removed an external portal package and replaced it with a local portal action, preserving popup and tooltip behavior.
* **Tests**
  * Added tests verifying the new portal action’s mounting, re-parenting, and teardown behaviors, including edge cases.
* **Chores**
  * Updated test tooling configuration and dependency list to reflect the removal of the external portal package.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tphakala/birdnet-go/pull/3313?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -24,8 +24,7 @@
         "hls.js": "^1.6.16",
         "maplibre-gl": "^5.23.0",
         "reconnecting-eventsource": "^1.6.5",
-        "svelte-dnd-action": "^0.9.69",
-        "svelte-portal": "^2.2.1"
+        "svelte-dnd-action": "^0.9.69"
       },
       "devDependencies": {
         "@ast-grep/cli": "^0.42.1",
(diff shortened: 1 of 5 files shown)


tphakala:
Good catch on the stale `optimizeDeps.include` entry. Removed `svelte-portal` from `vitest.integration.config.ts` in cbcec1a62 so Vite pre-bundling no longer tries to resolve the deleted package during integration tests. I also grepped the rest of the frontend config files (vite/vitest) and that was the only stale reference left from dropping the dependency.

tphakala:
@coderabbitai review

PR #3417, opened 2026-06-08
Updated Hungarian translation

cinadr:
Corrected Hungarian translation and translated missing strings for Health Diagnostics and new functions.

## Summary by CodeRabbit

* **Documentation**
  * Hungarian translations updated across the app: navigation, dashboard, analytics, audio/spectrogram controls, capture/clip settings, security, OAuth UI, alerting and System Health.
  * Numerous labels, placeholders and help texts reworded for clarity and consistent "észlelés" terminology; "Hangszűrő kártya" renamed to "Hangkártya", clip/recording wording standardized, and retention/privacy texts clarified.
  * Time-window abbreviations localized (e.g., "15m" → "15p").


cinadr:
## ⚠️ Workflow Permission Issue

Hi @tphakala, I've identified an issue with the i18n validation workflow that's preventing it from posting validation results as PR comments.

### Problem

The `i18n Validation` workflow is failing with a **403 Forbidden** error when attempting to comment on this PR:

```code
RequestError [HttpError]: Resource not accessible by integration POST /repos/tphakala/birdnet-go/issues/3417/comments - 403
```

### Root Cause

The workflow in `.github/workflows/i18n-validation.yml` is missing the required permissions to post comments on pull requests. The `actions/github-script` action that comments on PRs needs explicit permission to write to pull requests.

### Solution

The workflow file needs to include a `permissions` block. This should be added after the `on:` section in `.github/workflows/i18n-validation.yml`:

```yaml
name: i18n Validation

on:
  pull_request:
    paths:
      - 'frontend/static/messages/*.json'
      - 'frontend/src/**/*.svelte'
      - 'frontend/src/**/*.ts'
      - 'frontend/src/lib/i18n/**'
  push:
    branches:
      - main
    paths:
      - 'frontend/static/messages/*.json'
      - 'frontend/src/lib/i18n/**'

# Add this permissions block:
permissions:
  pull-requests: write
  contents: read

jobs:
  validate-translations:
    # ... rest of workflow
```
### Impact
Current Status: The validation runs but fails silently when trying to post results as a comment
After Fix: Validation results will be posted directly as PR comments, making it easy to see translation errors at a glance

### Notes
This is a security best practice in GitHub Actions—workflows need explicit permission grants
The pull-requests: write permission is only needed for PR comment functionality
The contents: read maintains read-only access to the repository
Once applied, the workflow will post detailed validation feedback on future PRs

Thank you.

tphakala:
Thanks for the Hungarian translation update, the bulk of it looks great and is much appreciated.

There's one change I need to adjust before this can merge. Two strings replaced the i18n link placeholders with hardcoded HTML anchors:

- `settings.support.supportReport.githubRequired.description`
- `settings.support.supportReport.githubIssue.helper`

Those keys use `{createIssueLink}` and `{viewIssuesLink}` placeholders that `SupportDumpCard.svelte` fills in at runtime. The component builds each link from the configured project URLs, runs them through an href sanitizer, and adds `rel="noopener noreferrer"`. Hardcoding the `<a href="...">` tags breaks that in three ways:

1. It fails the `validate-translations` check (the red CI here), because the required parameters are now missing from the Hungarian strings.
2. It pins the URLs to `github.com/tphakala/birdnet-go`, so the links ignore the project's configured issue URLs.
3. It drops `rel="noopener noreferrer"` on the `target="_blank"` links, which is the tabnabbing concern CodeRabbit also flagged.

I'll push a small fix to your branch that restores the two placeholders and keeps your improved wording plus the translated link labels (`createGithubIssue`, `viewExistingIssues`, `createNewIssue`), so nothing visible changes for Hungarian users. The rest of the translation is good to go. I left the couple of minor spelling notes from the automated review for you to take or leave.

tphakala:
Thanks for digging into this. You're right on the facts: the workflow has no `permissions:` block and the comment step does return a 403. But that 403 isn't what made the check red. The job fails on a separate final step because the Hungarian strings were genuinely invalid (the `{createIssueLink}` and `{viewIssuesLink}` placeholders had been replaced with hardcoded `<a href>` links). That comment step only posts a cosmetic summary, it doesn't decide pass/fail. I pushed a fix for the translation, so validation now passes and the comment step gets skipped entirely (it only runs on failure), which means CI goes green with no workflow change.

Also worth knowing: adding `permissions: pull-requests: write` wouldn't actually fix commenting on this PR. For `pull_request` runs from a fork, GitHub keeps the token read-only as a security measure, and a `permissions:` block can't elevate it. That advice works for same-repo branches but not fork PRs, so reverting it was the right call. Making the workflow comment cleanly on fork PRs needs a `pull_request_target`/`workflow_run` split, which is a separate hardening task from your translation work.

Thanks again for the contribution.

PR #3444, opened 2026-06-10
feat(analytics): add opt-in eBird species-page links

alexbelgium:
## Summary

Adds an optional, client-side eBird species-page link to every species view in the Analytics section. When enabled, each species shows a link that opens its eBird page in a new tab. The URL is built entirely client-side from the detection's existing species code plus an optional user-configured region and the current UI language — **no eBird API key required**.

**Surfaces:**
- **Desktop grid cards** — compact `eBird ↗` text pill next to the species name
- **Desktop table** — same `eBird ↗` text pill (consistent with grid)
- **Mobile cards** (card / compact / list variants) — `ExternalLink` icon as a positioned corner overlay (rendered as a sibling of the card `<button>` to keep valid HTML)

**Settings** (Integrations → eBird):
- `Show eBird species page links` toggle (independent of the API-key-based eBird features)
- Optional `eBird Region` field (e.g. `BE-WAL`) with client-side validation; empty = global page

## Implementation notes

- New `frontend/src/lib/utils/ebird.ts` helper: `buildEbirdSpeciesUrl`, `isValidEbirdSpeciesCode`, `isValidEbirdRegionCode`, `toEbirdSiteLanguage` — fully unit-tested (`ebird.test.ts`)
- Placeholder species codes (uppercase-prefixed, generated for species missing from the eBird taxonomy) are intentionally excluded, so the link only appears for real eBird codes
- Backend adds two purely-additive config fields: `realtime.ebird.showspeciespagelinks` (default `false`) and `realtime.ebird.speciespageregion` (default `""`), with viper defaults and documentation in the example `config.yaml`
- Settings hot-reload: the link reactively appears/updates as soon as the toggle/region change, no restart needed

## Backward compatibility

Purely additive. Existing `config.yaml` files without the new keys fall back to viper defaults (feature off). Existing eBird-integration users see no change until they explicitly enable the toggle. The two new JSON fields in the settings API response are additive and ignored by existing clients.

## Test plan

- [ ] Enable eBird integration, toggle "Show species page links" → links appear in grid, table, and mobile views
- [ ] Set a region (e.g. `BE-WAL`) → links resolve to the region-scoped eBird URL; clear it → links resolve to the global page
- [ ] Enter an invalid region (e.g. `BEWAL`) → inline validation message shown
- [ ] Toggle off → all links disappear without a page reload
- [ ] Species with placeholder/missing eBird code → no link shown
- [ ] `npm run check:all` and `npm test` pass

## Preflight Status

Ran the 6-pass preflight quality gate. All in-scope findings resolved:

- ✅ Aligned desktop-table link to the `eBird ↗` text pill (was a leftover `Binoculars` icon)
- ✅ Removed unused `ebird-link` CSS class and stale `Binoculars` import
- ✅ Simplified redundant `Boolean()` wraps
- ✅ Documented new fields in the embedded `config.yaml`

Verified clean: `{@html}` usage (static i18n content, established pattern), URL encoding (regex-guarded + `encodeURIComponent`), settings reactivity, i18n integrity (all 9 new keys translated across all 16 locales, `{species}` placeholder preserved), HTML validity of the overlay link.

Certification:
- [x] Single concern · [x] Linters clean · [x] Tests pass · [x] No regression / backward-compat break · [x] i18n complete · [x] No secrets or PII

image

image

image

### DIFF
--- a/frontend/src/lib/desktop/features/analytics/components/ui/SpeciesCard.svelte
+++ b/frontend/src/lib/desktop/features/analytics/components/ui/SpeciesCard.svelte
@@ -2,7 +2,6 @@
   import { cn } from '$lib/utils/cn';
   import { t } from '$lib/i18n';
   import { formatDate } from '$lib/utils/formatters';
-  import { Binoculars } from '@lucide/svelte';
 
   interface SpeciesData {
     common_name: string;
@@ -56,11 +55,11 @@
           href={ebirdUrl}
           target="_blank"
           rel="noopener noreferrer"
-          class="inline-flex shrink-0 items-center justify-center rounded-md p-1 text-[var(--color-base-content)] opacity-50 transition-colors hover:bg-[var(--color-base-300)] hover:text-[var(--color-primary)] hover:opacity-100"
+          class="inline-flex shrink-0 items-center rounded px-1.5 py-0.5 text-xs font-bold tracking-wide text-[var(--color-primary)] opacity-60 transition-opacity hover:opacity-100"
           title={t('analytics.species.viewOnEbird')}
           aria-label={t('analytics.species.viewOnEbirdAria', { species: species.common_name })}
         >
-          <Binoculars class="h-4 w-4" />
+          eBird&nbsp;↗
         </a>
       {/if}
     </div>
--- a/frontend/src/lib/desktop/features/analytics/components/ui/SpeciesCard.svelte
+++ b/frontend/src/lib/desktop/features/analytics/components/ui/SpeciesCard.svelte
@@ -2,6 +2,7 @@
   import { cn } from '$lib/utils/cn';
   import { t } from '$lib/i18n';
   import { formatDate } from '$lib/utils/formatters';
+  import { Binoculars } from '@lucide/svelte';
 
   interface SpeciesData {
     common_name: string;
@@ -17,9 +18,11 @@
   interface Props {
     species: SpeciesData;
     className?: string;
+    /** eBird species-page URL; when set, a link icon is shown next to the name. */
+    ebirdUrl?: string | null;
   }
 
-  let { species, className = '' }: Props = $props();
+  let { species, className = '', ebirdUrl = null }: Props = $props();
 
   function formatPercentage(value: number): string {
     return (value * 100).toFixed(1) + '%';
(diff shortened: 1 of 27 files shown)


tphakala:
@alexbelgium thanks for this PR, however I'm not going to merge it as-is though, and I want to lay out where I'm taking this so you don't sink more time into the eBird-specific branch before the direction is settled. It's still just an idea, so treat the below as the plan rather than a fixed spec.

The short version: I want external species links to be driven by OpenFauna, and I want BirdNET-Go to know nothing about any specific provider. OpenFauna already carries Wikipedia and iNaturalist data per species, and I'm reworking its metadata so each species gets a structured set of link entries rather than flat per-site URL columns. Each entry will carry what the UI needs to render it: a stable identifier, a display name, an optional title, ordering, and an icon hint. The point of storing identifiers instead of baked URLs is that the localization you did for eBird (deriving the URL from a code plus the UI language) generalizes to every source, so Wikipedia can resolve to the user's language at render time and so on.

The consequence I care most about is that adding a new source becomes an OpenFauna change plus a recompile, with zero frontend edits and no new birdnet-go release. So instead of an eBird component, a Wikipedia component and so on, birdnet-go gets one source-agnostic renderer that takes a list of link objects and draws a card for each. If OpenFauna later ships GBIF or Xeno-canto, the cards appear automatically. A brand-new source whose icon birdnet-go doesn't recognize still renders correctly with a generic external-link glyph, and a nicer icon can be added later as non-blocking polish.

eBird stays a special case, by design. OpenFauna deliberately does not ingest eBird or Clements-derived data because of the non-commercial licensing, so the eBird link can't live in the dictionary the way the others do. But because the renderer is just iterating a list, birdnet-go can construct an eBird link entry at runtime from the model's own species_code and append it to whatever OpenFauna provided. eBird never enters the compiled data, the dictionary stays CC-clean, and to the card renderer an eBird entry looks identical to every other one. So your buildEbirdSpeciesUrl, the validation, and the opt-in plus region settings are very likely to survive as the eBird path once the generic mechanism exists. It's the eBird-only framing of the whole feature that I want to avoid baking in.

There's one open question still in the air on the OpenFauna side: whether the per-source presentation metadata (display name, ordering, icon) ships inline on every link row or in a small separate sources table that gets joined on ingest. That choice affects the shape of what birdnet-go consumes, so I'd like to pin it before there's frontend work to redo. I'm currently leaning toward the separate sources table so that a source and how it presents is one authoritative record.

If you're interested in carrying this forward, the most useful direction would be reworking the analytics UI around that generic OpenFauna-backed card list, keeping your settings, region and translation work, with eBird wired in as the runtime special case. Happy to share the metadata shape once it firms up. Either way, appreciate the effort here.

alexbelgium:
Thanks very much for the time you took to lay out your vision. I always like long term coherent module-based approaches. I'll wait for your feature to merge then and look to embed the ebird page in it ! What I like about the ebird page is that it provides region-specific localized observations maps which can be useful in understanding how credible an observation species is.

I'll close this to keep the PR list tidy but will keep an eye out then

PR #3450, opened 2026-06-11
fix(classifier): canonicalize localized common-name species overrides (+2 adjacent fixes)

tphakala:
## Summary

Fixes a bug on the v3 geomodel range-filter path where user species overrides written as **bare localized common names** (for example the Finnish `sinitiainen`) never resolved. Overrides were matched only against the geomodel's English labels, so a localized name matched nothing and the raw string was appended verbatim to the inclusion working set. That caused two problems:

1. **Silent force-include failure (functional).** The inclusion gate keys on the scientific name (the substring before `_`). A bare common name produced a useless key, so a species the user explicitly added to `realtime.species.include` was dropped at the detection gate whenever it was out of range.
2. **Spurious log noise (cosmetic).** The OpenFauna name resolver received the bare word as a scientific name, could not localize it, and logged a `could not localize some working-set species` WARN. That WARN exists to surface genuine upstream taxonomy gaps in support dumps, so the false positives diluted the diagnostic.

`addUserOverrideSpecies` and `addUserOverrideSpeciesScores` now resolve each override via a shared `canonicalOverrideLabels` helper: it matches the geomodel labels first (so English and scientific overrides behave exactly as before) and falls back to the active classifier's localized labels, appending the canonical `Scientific_Common` label. Entries that match neither label set (non-bird classes such as `drone`/`heatpump`) are still appended verbatim and legitimately reported as unresolved.

Two adjacent latent bugs found while reviewing this code path are fixed in separate commits:

- **Rune-safe placeholder code.** `GeneratePlaceholderCode` built its prefix from raw bytes of the scientific name, which cut multibyte runes mid-sequence (corrupting the prefix into U+FFFD) or read the wrong byte. It now decodes runes. ASCII names are unchanged.
- **Privacy filter same-chunk edge.** The privacy filter discarded a bird detection only when a human voice was detected *strictly after* it started. When a human and a bird landed in the exact same audio chunk (equal timestamps), the detection leaked. The comparison now uses `>=` so an equal timestamp still trips the filter; an earlier human voice still keeps the later bird (no over-discard).

## Commits

- `fix(classifier): resolve localized common-name overrides against classifier labels`
- `fix(classifier): use rune-safe prefix in placeholder species code`
- `fix(processor): discard same-chunk human-and-bird detections in privacy filter`

## Test Plan

- [x] New failing-first tests for each fix:
  - `TestBuildRangeFilter_BareLocalizedCommonNameOverride_ForceIncludesViaGate`
  - `TestBuildRangeFilter_BareLocalizedCommonNameOverride_DoesNotPolluteNameResolver`
  - `TestGetProbableSpecies_BareLocalizedCommonNameOverride_CanonicalizesLabel`
  - `TestGeneratePlaceholderCode_MultibyteSingleWordScientificName` / `_MultibyteBinomialScientificName`
  - `TestShouldDiscardDetection_PrivacyFilter_SameChunkHumanVoice` / `_EarlierHumanVoiceKept`
- [x] `go test -race` green for `internal/classifier`, `internal/conf`, `internal/openfauna`, `internal/analysis/processor`
- [x] `golangci-lint run` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed placeholder-code generation for scientific names containing non-ASCII characters.
  * Privacy filter now discards bird detections when a human detection has the same or later timestamp.

* **Enhancements**
  * Canonicalization of user-provided species names so localized/common overrides resolve to canonical labels where possible.

* **Tests**
  * Added tests for privacy-filter boundary cases, localized-name override canonicalization, and multibyte scientific-name handling.

### DIFF
--- a/internal/analysis/processor/privacy_filter_test.go
+++ b/internal/analysis/processor/privacy_filter_test.go
@@ -27,52 +27,64 @@ func newPrivacyFilterDetection(source string, firstDetected time.Time) *PendingD
 	}
 }
 
-// TestShouldDiscardDetection_PrivacyFilter_SameChunkHumanVoice covers the
-// equal-timestamp edge: when a human voice and a bird are detected in the exact
-// same audio chunk, both carry the same StartTime, so the privacy filter's strict
-// "human after bird" comparison let the bird through. A simultaneous human voice
-// must still discard the detection.
-func TestShouldDiscardDetection_PrivacyFilter_SameChunkHumanVoice(t *testing.T) {
+// TestShouldDiscardDetection_PrivacyFilterBoundary exercises the privacy filter
+// across the human-voice-vs-bird timestamp boundary. The key case is equal
+// timestamps: a human voice and a bird detected in the exact same audio chunk
+// share the same back-dated start time, and that must still discard the bird
+// (the prior strict "human after bird" comparison leaked it). A human voice from
+// an earlier chunk (strictly before) must not discard a later bird.
+func TestShouldDiscardDetection_PrivacyFilterBoundary(t *testing.T) {
 	t.Parallel()
 
 	const source = "src1"
-	chunkTime := time.Date(2026, 6, 11, 8, 0, 0, 0, time.UTC)
+	base := time.Date(2026, 6, 11, 8, 0, 0, 0, time.UTC)
 
-	settings := &conf.Settings{}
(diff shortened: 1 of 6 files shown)


tphakala:
@coderabbitai addressed the nitpick: the two privacy-filter boundary tests are now a single table-driven test with t.Run, and I added a third row for a human voice detected strictly after the bird started.

PR #3462, opened 2026-06-12
fix(datastore): propagate scientific-name query error in GetThresholdEvents

tphakala:
## Summary

`GetThresholdEvents` in the v2only datastore runs two queries and merges them: one by the provided name and one by the resolved scientific name. The second query discarded its error (`if err == nil && len(sciEvents) > 0`), so a real database failure on that query was indistinguishable from "no events": `GET /api/v2/dynamic-thresholds/:species/events` returned HTTP 200 with partial or empty data instead of surfacing a 500.

This propagates the second query's error using the `internal/errors` builder (Component/Category/Context), consistent with the first query and the rest of the file. The redundant `len() > 0` guard is dropped (appending an empty slice is a no-op). The result sort gains an `ID` tie-break so events sharing a timestamp truncate deterministically when the limit is applied.

## Changes

- Propagate the scientific-name query error instead of swallowing it.
- Build the common-name query error with the same Component/Category/Context for symmetry and telemetry.
- Deterministic sort tie-break on event ID.
- Tests: positive coverage of the scientific-name query path (driven via common-name resolution) and an error-injection regression test that fails the second query and asserts the error now surfaces.

## Test Plan

- [x] `go test -race ./internal/datastore/v2only/` (incl. two new tests)
- [x] `golangci-lint run` (full project, 0 issues)
- [x] Regression test verified to fail against the pre-fix code, pass after

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for threshold-event lookups, returning structured context for both primary and secondary queries.
  * Made threshold-event result ordering deterministic: sort by creation time descending, tie-broken by ID descending.

* **Tests**
  * Added regression tests covering scientific-name resolution, propagation of secondary-query errors, and deterministic tie-break behavior for identical timestamps.

### DIFF
--- a/internal/datastore/v2only/datastore.go
+++ b/internal/datastore/v2only/datastore.go
@@ -3094,7 +3094,12 @@ func (ds *Datastore) GetThresholdEvents(speciesName string, limit int) ([]datast
 	// Query 1: Try with the provided name (common name) - finds legacy/incorrectly saved events
 	v2Events, err := ds.threshold.GetThresholdEvents(ctx, speciesName, limit)
 	if err != nil {
-		return nil, err
+		return nil, errors.New(err).
+			Component("datastore").
+			Category(errors.CategoryDatabase).
+			Context("operation", "get_threshold_events").
+			Context("query_type", "common_name").
+			Build()
 	}
 
 	// Query 2: If we can resolve to scientific name, also query with that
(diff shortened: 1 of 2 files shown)


tphakala:
Added in aefe67f09 as `TestV2OnlyDatastore_GetThresholdEvents_SameTimestampTieBreak`.

One nuance I hit while writing it: the per-query `LIMIT` is applied in SQL, so a single-query `limit=1` never reaches the Go-side sort (the DB pre-truncates to one row). The tie-break only governs how the *merged* result of the two queries is truncated. So the test stores one event found by Query 1 (a legacy common-name label) and one found by Query 2 (the resolved scientific name), both sharing a timestamp, and asserts the higher-ID event survives truncation to a single row. Verified it fails without the tie-break and passes with it.

PR #3464, opened 2026-06-12
feat(analytics): localize species names in charts and overview

tphakala:
## Summary

Follow-up to the per-visitor species-name localization overlay. This threads the existing `localizeSpeciesName` display helper through the analytics surfaces that were still showing server-locale names, so each visitor sees species common names in their own UI locale.

Surfaces localized:
- The three D3 species charts: time-of-day, daily trend, and the new-species timeline (tooltips, legends, and axis labels).
- The Analytics overview page: the species-distribution bars and the "Most Common Species" stat card.

## How it works

Display labels localize while D3 join keys, colors, and legend ids stay keyed on the canonical scientific name, so a species keeps a stable row and color across locales. The localized label is computed inside each chart's data-processing `$derived`, so the existing repaint `$effect` re-runs when the dictionary loads or the UI locale switches.

The new-species timeline keys its y band scale on the scientific name and renders the localized label via the axis `tickFormat`, avoiding the band-dedup collapse a name-keyed scale would cause (two species sharing a localized common name would otherwise share one row). An empty-string scientific name falls back to the common name for the key.

This reuses the existing client dictionary overlay, so there are no new translation keys and no locale-file changes. The fallback chain (client dictionary, then server common name, then scientific name) means a user whose UI locale equals the server locale sees identical text to before.

Scope note: the Analytics Species page search/sort, the settings species pickers, and the aria-label i18n pass are intentionally left for separate follow-up PRs, since each carries distinct risk (search/sort consistency, canonical-value handling, locale-file churn).

## Test Plan
- [x] `npm run check:all` (format, lint, typecheck, ast-grep) passes
- [x] `npm test` for analytics charts/pages and localization utils passes
- [x] `npm run build` succeeds
- [x] Two regression tests added for the new-species timeline band-key decoupling
- [ ] Manual: view Analytics and Advanced Analytics in a non-server UI locale; confirm chart tooltips, legends, axes, species bars, and the Most Common stat show localized names, that switching locale live repaints them, and that colors/legend toggles stay stable across the switch

## Summary by CodeRabbit

* **Bug Fixes**
  * Species names are now localized throughout analytics: tooltips, legends, axis labels, summary cards, and distribution bars.
  * Charts correctly differentiate species that share a common name by using stable identifiers so each species renders in its own row.

* **Tests**
  * Added tests covering timeline row-keying and Y-axis label formatting; test setup now resets mocks between runs.

### DIFF
--- a/frontend/src/lib/desktop/features/analytics/components/charts/d3/DailySpeciesTrendChart.svelte
+++ b/frontend/src/lib/desktop/features/analytics/components/charts/d3/DailySpeciesTrendChart.svelte
@@ -9,6 +9,7 @@
   import type { ZoomTransform } from 'd3-zoom';
 
   import { t } from '$lib/i18n';
+  import { localizeSpeciesName } from '$lib/utils/speciesDisplay';
   import BaseChart from './BaseChart.svelte';
   import { getLocalDateString } from '$lib/utils/date';
   import { createTimeScale, createLinearScale } from './utils/scales';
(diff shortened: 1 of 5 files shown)


tphakala:
Added the `beforeEach(() => vi.clearAllMocks())` isolation block in 0e1961b1. Thanks @coderabbitai.

PR #3488, opened 2026-06-13
fix(datastore,telemetry): v2only threshold not-found returns 404 not 500

tphakala:
## Summary

In v2only mode, `(*Datastore).GetDynamicThreshold` returned the bare `ErrDynamicThresholdNotFound` sentinel for a missing threshold. The API handler classifies errors through `handleErrorWithNotFound`, which only maps an `*errors.EnhancedError` with `CategoryNotFound` to 404, so the raw sentinel fell through to 500. The legacy datastore returns a `CategoryNotFound` `EnhancedError` for the same condition and correctly yields 404, so the two backends diverged on the HTTP status for an identical "threshold not found" case.

## Changes

- **`internal/datastore/v2only/datastore.go`**: wrap the not-found sentinel in a `CategoryNotFound` `EnhancedError` (never `CategoryDatabase`, so it is not surfaced to Sentry as a database error). The sentinel stays reachable via `errors.Is` (`EnhancedError.Unwrap`), so existing callers that check the sentinel keep working, and the API layer now maps it to 404, matching the legacy backend.
- **`internal/errors/telemetry_integration.go`**: suppress the benign `"dynamic threshold not found"` message in `shouldReportToSentry`. Building the `CategoryNotFound` error triggers telemetry at construction time; without this it would be reported at `LevelInfo` when telemetry is active. This also covers the legacy backend, which produces the same message.

This is a pre-existing backend-divergence bug (observable only on the v2only backend when querying a species with no stored dynamic threshold): a backend-divergent status code, not data loss.

## Test Plan

- [x] `go test -race ./internal/datastore/v2only/ ./internal/errors/ ./internal/api/v2/`
- [x] v2only not-found path now produces a `CategoryNotFound` `EnhancedError` that still matches the sentinel via `errors.Is`
- [x] `shouldReportToSentry` suppresses the `"dynamic threshold not found"` message
- [x] Handler boundary pinned both ways: `CategoryNotFound` -> 404, bare sentinel -> 500
- [x] `golangci-lint run ./...` clean, `gofmt` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Missing dynamic thresholds now produce a 404 response instead of a 500.

* **Tests**
  * Added tests covering mapping of missing-threshold errors to HTTP status and Sentry-suppression behavior.

* **Chores**
  * Adjusted error classification and monitoring to avoid noisy alerts for expected “dynamic threshold not found” cases.

### DIFF
--- a/internal/api/v2/dynamic_thresholds_test.go
+++ b/internal/api/v2/dynamic_thresholds_test.go
@@ -1,15 +1,19 @@
 package api
 
 import (
+	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 
+	"github.com/labstack/echo/v4"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/tphakala/birdnet-go/internal/analysis/processor"
 	"github.com/tphakala/birdnet-go/internal/conf"
 	"github.com/tphakala/birdnet-go/internal/datastore"
 	"github.com/tphakala/birdnet-go/internal/datastore/mocks"
+	"github.com/tphakala/birdnet-go/internal/errors"
 )
 
 func TestGetMergedThresholdData_NoDuplicates(t *testing.T) {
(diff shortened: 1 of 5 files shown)


tphakala:
Good catch on the stale header comment. Updated it in 06fb791 to say the benign not-found is now wrapped as a CategoryNotFound EnhancedError (still reachable via errors.Is, never reported to Sentry as a DB error) instead of propagating unwrapped.

@coderabbitai review

Issue #3405, opened 2026-06-07
feat: allow management by species instead of by detections


alexbelgium:
### What problem does this solve?

Currenlty birdnet-go allows management at the level of detections, not species. However, when the system runs for some time without supervision there can be a significant bias that appears in terms of accumulation of wrongly identified species due to model bias.

If taken early enough detections can be flagged as false positives and species excluded, but it can require some time to delete by bulk detections by searching for them in the Search tool then bulk delete.

### Proposed solution

In the Analytics > Species > table, show a third mode of "Species edition" once authentificated

This "species edition" mode shows the usual table mode except : 
- Shows all species in the database, not just non-excluded ones
- Shows additional sortable columns :
  - Is the species part of the include or exclude lists
  - If reviews were done, what is the ratio of positive vs false positives
  - Last column that allows to delete the entire species both from detections and file disk

This is what is could look like :
<img width="1194" height="654" alt="Image" src="https://github.com/user-attachments/assets/be0d5176-08d9-4daa-a60e-13c300aa1af3" />

### Alternatives considered

Continue current mode of management : 
- Bulk delete detections by searching for the species name and delete them
- Accept that excluded species are kept in the database, just hidden

### Pre-submission checks

- [x] I have searched existing open and closed issues and this feature has not already been requested.

rodrigodavid:
The suggestion to delete items in bulk might be related to the enhancement proposed in #2838.

alexbelgium:
> The suggestion to delete items in bulk might be related to the enhancement proposed in [#2838](https://github.com/tphakala/birdnet-go/issues/2838).

Thanks ! Indeed this one was addressed through https://github.com/tphakala/birdnet-go/pull/3119 but i was seeing more this pr as proposing a species-centric as opposed to detections centric approach for review and deletion

rodrigodavid:
I really like your suggestion, Alex. I’d definitely use it. I referenced the other suggestion because maybe they could be implemented together. 😉

alexbelgium:
This made me realize that this could also be an interesting discussion on how to manage species list for a location (confirmed ; excluded ; putative) and I've created a thread here : https://github.com/tphakala/birdnet-go/discussions/3413

PR #3490, opened 2026-06-14
feat(analytics): add Species Manage view for curation

alexbelgium:
## Summary

Adds an authenticated, desktop-only **Manage** view mode to the Species analytics page for species curation: excluded / included (whitelisted) / confirmed toggles, review ratio, range probability, and per-species deletion. The Manage view also surfaces **fully-rejected species** — species whose every detection was marked a false positive, so they no longer appear in the period summary — so they remain reviewable and deletable (#693). Grid and list views are unchanged.

Addresses https://github.com/tphakala/birdnet-go/issues/3405

## Backend

- **Config:** `Realtime.Species.Confirmed []string` (analytics-only; deep-copied in `CloneSettings`). It is persisted but never read by detection processing.
- **Datastore:** `GetSpeciesReviewStats` (per-species total / verified / rejected — false positives are counted, now also returning the resolved **common name** so a row can be rendered for species absent from the summary) and `GetSpeciesNoteIDs` (handles the legacy `"ScientificName_CommonName"` label format), implemented for the legacy GORM v1 store, the v2 repository, and the v2only datastore. The legacy store resolves the common name via `MAX(notes.common_name)` within the existing `GROUP BY`; v2only resolves it via `resolveCommonName` (the v2 repository layer stays name-free, as its labels table has no common-name column).
- **API v2** (all authenticated; handlers type-assert optional datastore capability and return **501** when unsupported):

  | Method | Path | Purpose |
  | ------ | ---- | ------- |
  | GET  | `/analytics/species/review-stats` | Per-species total/verified/rejected (+ common name) |
  | POST | `/detections/include` | Toggle always-include list |
  | GET  | `/detections/included` | Fetch always-include list |
  | POST | `/detections/confirm` | Toggle confirmed list |
  | GET  | `/detections/confirmed` | Fetch confirmed list |
  | POST | `/detections/species/delete` | Delete all detections for a species (skips locked) |

  The batch-delete loop was extracted into a shared `deleteNotesByIDs` helper (behavior-preserving) and reused by the species-delete handler. `README.md` updated.

## Frontend

- `Species.svelte`: third `manage` view mode with an auth-gated toggle (`$isAuthenticated`); parallel fetch of review-stats / included / confirmed / excluded plus async, session-cached range scores (`settingsActions.loadRangeFilterSpecies()`); the table renders before range scores arrive (per-cell spinner). Manage hides average confidence and first-seen. Review ratio shows `{verified} / {rejected}` (`—` when no reviews); range probability shows the score / `—` / spinner. Membership uses reactive `SvelteSet`/`SvelteMap` with per-(list, species) in-flight guards. Delete opens a confirmation modal showing the **all-time** detection count; a full delete removes the row locally, a partial delete refetches. Manage-only column sorts live in separate state and never persist to `localStorage`. Search reuses the existing debounced behavior.
- **Fully-rejected species (#693):** the Manage view acts on a superset of the period summary. A `manageSpecies` derived merges summary species with synthesized rows for any review-stats-only species (carrying the all-time count; an em dash for max-confidence and last-seen, which do not survive in the summary), so species with no remaining non-false-positive detections can still be reviewed/deleted. The empty-state condition and the filtered-count badge are Manage-aware, and the grid/list and Manage search filters share a single `speciesMatchesSearch` predicate. `commonName` is optional on the client, so a newer frontend talking to an older backend degrades gracefully to the scientific name.
- i18n: the `analytics.species.manage.*` keys and `analytics.species.switchToManage` are now **translated in all 15 non-English locales** (cs, da, de, es, fi, fr, hu, it, lv, nb, nl, pl, pt, sk, sv) — they had previously been English fallback only. The `{count}`/`{species}` placeholders are preserved; key structure is unchanged, so the generated types and `i18n:sync:check` stay valid.

## Testing

- Go (`-race`): config clone, datastore review-stats / note-IDs (incl. legacy-label regression) across all three store paths, and API v2 endpoints (success / empty / 501; include & confirm add-remove-action-saved; species-delete deletes / no-match / skips-locked / counts / 501). Review-stats tests now also assert the resolved common name.
- Frontend: `npm run check:all` clean (0 errors); `Species.test.ts` extended with Manage-view tests (auth-gating, hidden columns, review-ratio rendering, `—`, sort non-persistence) plus a test asserting a fully-rejected species surfaces with its all-time count, em dashes, and review ratio. `i18n:validate:full` passes (sync-check, type-check, coverage).

## Preflight

The #693 surfacing change was run through the preflight gate (6 review passes: reuse, correctness, quality, i18n, integration wiring, regression/back-compat). No Critical/High/Medium findings; the one reuse finding (duplicated search predicate) was fixed by extracting `speciesMatchesSearch`. The change is additive on the API wire and read-only on the DB; reviewers classified the new Manage rows as a restorative, backward-compatible improvement (no maintainer-confirm items).

## Notes

- `nb.json` and `hu.json` show large churn because those locale files had **pre-existing drift** (missing keys / non-canonical order from earlier PRs). The repo's `i18n:sync:check` gate requires canonical locale files, so running the documented `npm run i18n:sync` workflow normalizes them; reverting them fails the gate. The other 14 locales are a clean additive change. Happy to split those two files into a separate cleanup commit if preferred.
- Excluded/included/confirmed toggles key by common name (matching the existing `/detections/ignore` contract); review-stats, range scores, and delete key by scientific name.

https://claude.ai/code/session_01CnkYpgci2ayc2JZvsdVEBT

---
_Generated by [Claude Code](https://claude.ai/code/session_01CnkYpgci2ayc2JZvsdVEBT)_

### DIFF
--- a/frontend/src/lib/desktop/features/analytics/pages/Species.svelte
+++ b/frontend/src/lib/desktop/features/analytics/pages/Species.svelte
@@ -1,3 +1,10 @@
+<script module lang="ts">
+  // Range-filter scores require a full geo evaluation, so cache them for the
+  // browser session. Module-scoped so the cache survives navigating away from
+  // and back to the Species page within the same SPA session.
+  let cachedRangeScores: Map<string, number> | null = null;
+</script>
+
 <script lang="ts">
   import { t, getLocale, type TranslationKey } from '$lib/i18n';
   import { getLocalDateString, parseLocalDateString } from '$lib/utils/date';
@@ -7,8 +14,15 @@
   import { getStoredValue, setStoredValue } from '$lib/utils/storage';
   import { buildAppUrl } from '$lib/utils/urlHelpers';
   import { localizeSpeciesName } from '$lib/utils/speciesDisplay';
+  import { isAuthenticated } from '$lib/utils/auth';
+  import { settingsActions } from '$lib/stores/settings';
+  import { api } from '$lib/utils/api';
   import { onMount, onDestroy } from 'svelte';
+  import { SvelteSet, SvelteMap } from 'svelte/reactivity';
+  import { Trash2, SlidersHorizontal } from '@lucide/svelte';
   import SortableHeader from '$lib/desktop/components/ui/SortableHeader.svelte';
+  import Checkbox from '$lib/desktop/components/forms/Checkbox.svelte';
+  import Modal from '$lib/desktop/components/ui/Modal.svelte';
   import SpeciesFilterForm from '../components/forms/SpeciesFilterForm.svelte';
   import SpeciesDetailModal from '../components/modals/SpeciesDetailModal.svelte';
   import SpeciesCard from '../components/ui/SpeciesCard.svelte';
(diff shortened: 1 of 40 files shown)


alexbelgium:
This became too messy I'll create a new one based on current main

PR #3497, opened 2026-06-14
feat(health,analysis): surface results-queue detection drops on System Health

tphakala:
## Summary

Detection results dropped because the classifier results queue is full were
only visible in a log line and rate-limited telemetry. A user on constrained
hardware had no in-app way to see they were losing detections, or how many.
Each drop is real data loss: a species that was heard is never recorded.

This adds a `ResultsQueueDropCheck` to the System Health page, modeled on the
existing audio `BufferDropsCheck`:

- **`internal/observability`**: new `MetricPrefixResultsQueueDrops` counter prefix.
- **`internal/analysis`**: the pipeline records each drop into the shared
  `HealthMetricsStore` and health event buffer via an atomic sink, wired from
  the API controller in `APIServerService.Start` and detached in `Stop`. The
  store is the bridge so `internal/api/v2` never imports `internal/analysis`
  (no import cycle): analysis pushes drops, the check reads them.
- **`internal/health/checks`**: `ResultsQueueDropCheck` (analysis category)
  reuses the shared `evalWindowedStats` helper, so it gets windowed counts, a
  sparkline, and recent events for free. Warning on any drop within the
  window, critical on a high or sustained drop rate. Registered next to
  `QueueDepthCheck`.
- **collector**: zero-seeds the new per-source key on first sight so the check
  reads Healthy from startup, consistent with the audio drop checks.

The System Health page renders the new check generically (name, status pill,
message, and the existing sparkline / recent-events detail panel), so there is
no frontend change.

## Test Plan

- [x] `go test -race ./internal/health/... ./internal/observability/ ./internal/analysis/ ./internal/api/v2/` (all green)
- [x] `golangci-lint run` (0 issues)
- [x] Unit tests cover the check thresholds (skipped / healthy-with-history /
  warning-on-any-drop / critical / window boundary / sparkline+events) and the
  drop-to-store recording path (records key + event, clearing the sink stops
  recording, nil event buffer tolerated).
- [x] Collector seeding test asserts the new key is zero-seeded on first tick.

## Summary by CodeRabbit

* **New Features**
  * Added system health monitoring for detection result queue drops, including configurable warning/critical thresholds and windowed statistics.
  * Enhanced diagnostics to surface these queue-drop metrics via health metrics and recent health events.
  * Ensured consistent startup behavior by seeding the queue-drop metric so the check reports healthy (not skipped) initially.
* **Tests**
  * Added/expanded coverage for health metric and event recording, window behavior, and nil/edge-case handling for queue-drop diagnostics.

### DIFF
--- a/internal/analysis/api_service.go
+++ b/internal/analysis/api_service.go
@@ -214,6 +214,15 @@ func (s *APIServerService) Start(ctx context.Context) error {
 		s.server.APIController().SetShutdownRequester(appInstance)
 	}
 
+	// Surface results-queue detection drops on the System Health page by
+	// bridging them from the analysis pipeline into the diagnostics health
+	// store owned by the API controller. The store is the bridge because
+	// internal/api/v2 must not import internal/analysis (import cycle):
+	// analysis pushes drops into the store, ResultsQueueDropCheck reads them.
+	if ctrl := s.server.APIController(); ctrl != nil {
+		SetResultsQueueDropHealthSink(ctrl.HealthMetricsStore(), ctrl.HealthEventBuffer())
+	}
+
 	startSucceeded = true
 	return nil
 }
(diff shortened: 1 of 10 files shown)


tphakala:
Good catch on the magic string. Fixed in 5d5153ef: added `observability.MetricTypeResultsQueueDrops = "queue_drops"` and now build `MetricPrefixResultsQueueDrops` from it (`"results." + MetricTypeResultsQueueDrops + "."`), so the producer (the drop path in `process.go`) and the consumer (`ResultsQueueDropCheck`'s event filter) share one source of truth and can't drift. The tests now reference the constant too.

@coderabbitai review

PR #3499, opened 2026-06-14
refactor(imageprovider): replace jason and html2text deps with stdlib

tphakala:
## Summary

Removes two small third-party Go dependencies from the Wikipedia image provider, replacing them with standard-library equivalents to shrink the supply-chain surface. No new dependencies are added and there is no user-facing behavior change.

- `github.com/k3a/html2text` is replaced with a local `htmlToText` helper built on `golang.org/x/net/html` (already imported by this package). It walks the parse tree, skips `<script>`/`<style>`, decodes entities, and collapses whitespace. Used at the four attribution/error-page stripping call sites.
- `github.com/antonholmquist/jason` (dynamic JSON navigation) is replaced with typed structs decoded by `encoding/json`. The MediaWiki `formatversion=2` response is modeled directly; the raw bytes are retained on the response for the existing debug-logging helpers.

While in the file I also removed pre-existing dead code surfaced by the analyzer (three unused `apiErrorCategory` vars, an unused package logger, and the never-called `BirdImageCache.fetchDirect` / `updateMetrics` methods), and fixed two pre-existing bugs found in review:

- `extractText` rendered only the anchor's first child node, truncating author names split across nodes (for example `John <b>Doe</b>` became `John`). It now renders the whole node.
- `tryInitialize` deleted the per-species init mutex from its `sync.Map` on unlock, which broke mutual exclusion and allowed duplicate concurrent fetches for the same species. The mutex is now retained for the cache lifetime (bounded by the number of distinct species).

One correctness fix is specific to the struct rewrite: MediaWiki `extmetadata` values are mixed-type (most are strings, but `CommonsMetadataExtension` carries a JSON number), so the value field is typed `any` and read through a small `extMetaString` helper. A `string`-typed field would have made `encoding/json` reject every real author-info response.

## Test plan

- [x] `go test -race ./internal/imageprovider/` passes, including new tests: `htmlToText` table (script/style skip, entity decode, whitespace collapse, inline/block separation), typed-decode tests for the thumbnail and author responses, a regression fixture with a numeric `extmetadata` value, and a multi-node anchor author-name case.
- [x] `golangci-lint run` clean across the whole project.
- [x] `go build ./...` clean; `go mod tidy` removes both deps (and jason's `goconvey` transitive) from `go.mod`/`go.sum`.

## Release Notes

* **Bug Fixes**
  * Improved extraction of author/artist attribution from Wikipedia and other HTML sources, including names split across nested elements and anchors.
  * Updated HTML-to-text processing for more reliable plain-text output (inline/nested tags, whitespace/entity normalization) and safer handling of script/style content.
  * Improved request handling for Wikipedia lookups to respect cancellation and avoid blocking during retries.

* **Tests**
  * Added unit tests for HTML-to-text conversion and Wikipedia API JSON decoding, including numeric extmetadata and missing/optional fields.
  * Added a test ensuring anchor text spanning multiple child nodes is parsed correctly.


tphakala:
Good catch on the context propagation. Fixed in 8ba04f4: `createHTTPRequest` and the diagnostic request in `makeRateLimitedRequest` now use `http.NewRequestWithContext`, and the retry backoff selects on `ctx.Done()` with a timer instead of a blind `time.Sleep`, so a cancelled or timed-out fetch returns promptly. These were pre-existing gaps unrelated to the dependency swap, but worth closing while I was in the file.

@coderabbitai review

tphakala:
Right, that was an incomplete fix on my part: I made `makeRateLimitedRequest` take a context but left the caller chain passing `context.Background()`. Threaded it through in 66a7cbc, so `queryWithRetryAndLimiter` now forwards its context down through `handleJSONParsingErrorIfNeeded` and `handleJSONParsingError` to the diagnostic request. I also swept the package: no remaining `http.NewRequest` without a context, and the only `context.Background()` left is the no-context `Fetch` entry point.

PR #3565, opened 2026-06-18
fix(spectrogram): write PNG atomically via temp + rename

tphakala:
## Summary

The spectrogram PNG was written directly to its final path by sox/ffmpeg (`-o finalPath`), so a crash or kill mid-write, or two concurrent generations of the same path, could leave a partial or corrupt `.png` at the final path. The serve fast-path treats any nonzero file as a finished spectrogram, so it would then serve that corrupt image.

`GenerateFromFile` and `GenerateFromPCM` now render into a process-unique temp file in the same directory (`audiotemp.UniquePath`) and atomically rename it onto the final path via SecureFS (`os.Root.Rename`), removing the temp on every failure path. The temp name reuses the existing `audiotemp` naming that the audio-export scan (`isExportTempFor`) and diskmanager already skip, so it is never mistaken for an audio clip or a finished spectrogram. On failure nothing is published, so the serve fast-path correctly misses and regenerates instead of serving a partial.

`audiotemp` gains `FinalizeWith(temp, final, renameFn)` so the spectrogram can inject the sandboxed SecureFS rename while reusing the existing Windows rename-retry; `Finalize` delegates to `FinalizeWith(os.Rename)`, byte-for-byte unchanged for the existing audio-clip writers (FFmpeg/FLAC/WAV).

This mirrors the recently-landed atomic-write treatment for audio clips, extending it to the spectrogram sidecar.

## Test Plan
- [x] `golangci-lint run ./...` (0 issues)
- [x] `go test -race ./internal/spectrogram/ ./internal/audiocore/audiotemp/`
- [x] Stub-sox runtime-failure test: a sox stub writes partial bytes to its `-o` target then exits non-zero; asserts the partial lands only in the temp (cleaned up) and never at the final path. This would fail on the old code (which passed the final path as `-o`), so it discriminates the fix.
- [x] Real-sox success test: asserts the final spectrogram is published and no temp sidecar remains.
- [x] Unit tests for `FinalizeWith` (injected-rename success and error propagation).

## Summary by CodeRabbit

* **New Features**
  * Spectrogram generation now writes to process-unique temporary files and atomically publishes to the final output to prevent partial/corrupt files.
  * Improved Windows reliability by retrying the finalization step when the publish/rename initially fails.

* **Tests**
  * Added coverage ensuring runtime and configuration failures never leave partial final outputs or leftover spectrogram temp artifacts.
  * Added tests validating atomic-write behavior on successful spectrogram generation.

### DIFF
--- a/internal/audiocore/audiotemp/audiotemp.go
+++ b/internal/audiocore/audiotemp/audiotemp.go
@@ -4,6 +4,10 @@
 // unique temp file and atomically renames it into place, so two simultaneous
 // detections that resolve to the same clip path (see GitHub #3323) dedup safely
 // instead of colliding on a shared temp or corrupting the final file mid-write.
+//
+// The spectrogram generator reuses the same naming (UniquePath) and finalize
+// retry via FinalizeWith, injecting a SecureFS rename so its writes stay inside
+// the os.Root sandbox.
 package audiotemp
 
 import (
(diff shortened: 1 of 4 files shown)


tphakala:
Added `t.Parallel()` to the four new tests in 1facc25c0. Confirmed they're safe to parallelize: each uses its own `t.TempDir`/SecureFS/Settings and nothing in the package publishes global `conf` settings, so `currentSettings()` reads the per-test instance.

PR #3568, opened 2026-06-18
fix(classifier): run model-load warm-up outside o.mu

tphakala:
## Summary

The per-model RSS-delta warm-up inference ran while the model loader held `o.mu.Lock()`. For a runtime model install (`LoadModel`, via the model gallery) this briefly blocked all inference: `PredictModel`, `ModelInfos`, and `PrimaryModelID` all wait on `o.mu.RLock()` for the warm-up duration (bounded by `warmupTimeout`, 5s).

This defers the warm-up off the lock:

- Loaders (`loadPerch`, `loadBat`) now register the model entry, then queue the warm-up via `deferWarmup` instead of warming up inline.
- The caller drains the queue via `runPendingWarmups` after releasing `o.mu`, running the warm-up through a serialized path (`inferenceMu` then `entry.mu`) like a normal first inference, without recording into `globalInferenceCounters`.
- `loadAdditionalModels` drains per loop iteration so each model's RSS-after sample is taken before the next model allocates its arena (accurate accounting).
- `LoadModel` drains via `defer`, so a loader that queues a warm-up and then fails cannot orphan a pending entry in the queue.
- Teardown races with `UnloadModel`/`Delete` are handled by re-fetching the entry under `o.mu.RLock()` and skipping a torn-down model (absent, or `instance == nil`), leaving no stale `modelRSS` entry.

The primary warm-up (`NewOrchestrator`) and the secondary hot-reload warm-up (`ReloadSecondaryModels`) already run lock-free and are unchanged.

Lock ordering (`o.mu` -> `inferenceMu` -> `entry.mu`) is unchanged; the new `warmupRegisteredModel` mirrors `PredictModel`'s protocol exactly.

## Test Plan

- [x] `go test -race ./internal/classifier/` passes
- [x] `golangci-lint run ./...` clean
- [x] New unit tests: deferred warm-up records RSS; multi-model drain loop warms every queued model; non-fatal warm-up failure still records; teardown-skip paths (absent entry / nil instance); `LoadModel` drain wiring
- [x] Regression guard proves `o.mu.RLock()` stays free while a warm-up inference is in flight (mutation-verified: fails when the warm-up holds `o.mu`)

## Release Notes

* **Bug Fixes**
  * Model initialization no longer blocks inference requests; warm-up operations now execute after setup completes.

* **Performance**
  * Memory profiling accuracy improved by measuring memory deltas per-model during initialization rather than all at once.

* **Reliability**
  * Warm-up failures are non-fatal and handled gracefully without impacting system stability.

* **Tests**
  * Expanded test coverage for warm-up deferral behavior, including failure scenarios and concurrency validation.

### DIFF
--- a/internal/classifier/orchestrator.go
+++ b/internal/classifier/orchestrator.go
@@ -103,6 +103,22 @@ type Orchestrator struct {
 	modelRSS         map[string]int64
 	runtimeBaseline  int64
 	baselineCaptured bool
+
+	// pendingWarmups queues deferred warm-ups recorded by model loaders while
+	// they hold o.mu (write lock). Drained by runPendingWarmups after o.mu is
+	// released, so the warm-up inference runs via the serialized inference path
+	// instead of stalling PredictModel on o.mu. Appended only
+	// under o.mu.Lock(); snapshotted and cleared under o.mu by the drainer.
+	pendingWarmups []pendingWarmup
+}
+
+// pendingWarmup defers a freshly-registered model's warm-up + RSS measurement
+// until after o.mu is released. Running the warm-up inference under o.mu would
+// block every PredictModel/ModelInfos caller (which wait on o.mu.RLock) for the
+// inference duration.
+type pendingWarmup struct {
+	modelID string // o.models key of the freshly registered instance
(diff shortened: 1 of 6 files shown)


tphakala:
@coderabbitai review

PR #3607, opened 2026-06-21
fix(analytics): localize chart date axis labels to active locale

LeoColman:
## Problem

Analytics chart date/time axis labels were always rendered in **English**, regardless of the app locale. `createDateAxisFormatter` used d3's `timeFormat`, which formats against d3's *global default* locale — and the app never configures it. So a Portuguese user still saw `Mon 15` / `Jan 2024` on their charts.

## Fix

Replace `d3 timeFormat` with `Intl.DateTimeFormat` driven by `getLocale()`. Weekday and month names now follow the active app locale (e.g. `Mon 15` → `seg. 15`, `Jan 2024` → `jan. 2024` in Portuguese).

The original `%a %d` / `%b %d` / `%b %Y` layout — weekday/month first, zero-padded day — is preserved by composing the parts manually (`padDay` helper), so only the translated names change, not the column layout. The `Intl.DateTimeFormat` instances are built once per formatter, not per tick.

## Scope

- `createDateAxisFormatter` only (`day`/`week`/`month`/`year` buckets).
- No new translation keys — `Intl` supplies localized names, so no `*.json` changes needed.

## Testing

- Added a test asserting Portuguese output (`seg. 15`, `jan. 15`, `jan. 2024`) via the mocked `getLocale`.
- `npm run check:all` clean: prettier, eslint, stylelint, `svelte-check` (0 errors/warnings), ast-grep rules.
- Full `axes.test.ts` suite passes (28 tests).

## Bug Fixes
* Chart date axis labels now respect your app’s active locale, displaying localized month and weekday abbreviations for daily, weekly, monthly, and yearly chart views.

### DIFF
--- a/frontend/src/lib/desktop/features/analytics/components/charts/d3/utils/__tests__/axes.test.ts
+++ b/frontend/src/lib/desktop/features/analytics/components/charts/d3/utils/__tests__/axes.test.ts
@@ -1,4 +1,5 @@
-import { describe, it, expect, beforeEach, beforeAll } from 'vitest';
+import { describe, it, expect, beforeEach, beforeAll, vi } from 'vitest';
+import { getLocale } from '$lib/i18n';
 import { select } from 'd3-selection';
 import { scaleLinear } from 'd3-scale';
 import { timeFormatDefaultLocale } from 'd3-time-format';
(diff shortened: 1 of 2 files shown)


tphakala:
Thanks for this, clean fix. I ran it through my full review and the localization approach is solid; the new test is a genuine regression guard since it fails against the old d3 code (`Mon 15` vs `seg. 15`).

I pushed one small polish commit on top (c44958b):

- Switched the day-range clock formatter from `hour12` to an explicit `hourCycle` (`h23` for 24h, `h12` for 12h). This keeps exact parity with the previous D3 `%H:%M` / `%I:%M %p` tokens and guarantees midnight renders as `00:00` instead of `24:00` on any locale whose default 24-hour cycle resolves to h24. None of the locales currently shipped hit that, so it is mostly defensive, but it is also more faithful to the old behavior.
- Removed the now-dead `timeFormatDefaultLocale` setup from the axes test. The formatter no longer touches d3-time-format, so that English-locale block had no effect; the English expectations now come from `Intl.DateTimeFormat` with `getLocale()` mocked to `en` in the shared test setup.

Both files still pass `npm run check:all` and all 28 axes tests. Good to merge once CI is green.

PR #3632, opened 2026-06-22
Updates to the podman install script

barmst02:
I had to make a number of changes the script to get it to work. Perhaps some of these were due to my local setup, but they look to me to apply generally. 

One other thing I had to do was manually create config and data directories in ~/.config/containers/systemd. I don't think this is the expected behavior, but I wasn't sure how to resolve in the code.

## Summary by CodeRabbit

* **Chores**
  * Updated the default web interface port to **8888** (previously **8080**) across port checks and status URLs.
  * Updated container service configuration to pass through the sound device using `AddDevice` and simplified `/config/hls` tmpfs permissions (keeps size/mode).
  * Changed startup behavior so the service starts without an explicit user enablement step.
  * Refreshed end-of-script messaging/spacing to match the updated flow.

### DIFF
--- a/doc/wiki/non-bird-species.md
+++ b/doc/wiki/non-bird-species.md
@@ -1,6 +1,8 @@
 # Non-Bird Species in BirdNET v2.4
 
-BirdNET can detect several non-bird species, including mammals, amphibians, and insects. Below is the full list of non-bird species supported in the BirdNET v2.4 model.
+BirdNET can detect several non-bird species, including mammals, amphibians, and insects, as well as some common environmental sounds. Below is the full list of non-bird labels supported in the BirdNET v2.4 model.
+
+## Biological Species
 
 | Scientific Name | Common Name | Class | Order | Family |
 | --- | --- | --- | --- | --- |
@@ -92,3 +94,19 @@ BirdNET can detect several non-bird species, including mammals, amphibians, and 
 | *Sciurus carolinensis* | Eastern Gray Squirrel | Mammalia | Rodentia | Sciuridae |
 | *Tamias striatus* | Eastern Chipmunk | Mammalia | Rodentia | Sciuridae |
 | *Tamiasciurus hudsonicus* | Red Squirrel | Mammalia | Rodentia | Sciuridae |
+
+## Environmental & Artificial Sounds
+
+The model also includes several soundscape classes to help detect common background noise and human activities:
+
+- Dog
+- Engine
+- Environmental
+- Fireworks
+- Gun
+- Human non-vocal
+- Human vocal
+- Human whistle
+- Noise
+- Power tools
+- Siren


tphakala:
Thanks for digging into this. A few of these are real fixes I want to keep, but a couple of the changes break the install, so I can't merge as-is. Let me walk through it.

**The port change to 8888 breaks the web UI.** The generated `config.yaml` now sets `webserver.port: 8888`, so the app binds to 8888 inside the container. But the publish mapping is still `PublishPort=${WEB_PORT}:8080`, which forwards the host port to container port **8080**. Nothing listens on 8080 inside the container anymore, so the web interface ends up unreachable. The container healthcheck also probes `localhost:8080` internally, so it'll report unhealthy too. The container-internal port needs to stay 8080. If you want a different host port, change only the left side of `PublishPort` (and `WEB_PORT`) and leave `webserver.port` at 8080.

**`AddDevice=/dev/null:/dev/snd` disables sound card capture.** This maps `/dev/null` onto `/dev/snd` in the container, so ALSA sees no real audio device. For anyone capturing from a local sound card that kills audio input. The `Device=` to `AddDevice=` rename is correct (`Device=` isn't a valid Quadlet key), but the target should be the real device. Use `AddDevice=-/dev/snd`: the leading `-` makes it optional so it won't fail the unit on hosts without a sound card, which I'm guessing is what you were working around.

**The fixes only land in the inline fallback.** The script only uses that heredoc when `Podman/quadlet/birdnet-go.container` isn't present (the curl install path). When run from a clone it copies the repo Quadlet files, which still have the old `Device=/dev/snd:/dev/snd` and hardcoded UID. Those files should get the same treatment so both paths behave the same. Note that `create_default_config` always writes the config, so the 8888 port issue above actually hits both paths.

The rest is good and I'd like to keep it:

- `%U` / `%G` instead of `$(id -u)`: correct, systemd doesn't shell-expand unit files so the old form wrote a literal `$(id -u)` string into the env var.
- Dropping the broken `uid=`/`gid=` tmpfs options: same reason, those literal `$(id -u)` values would have failed the mount.
- Removing `systemctl --user enable`: Quadlet-generated units can't be enabled that way; the `[Install] WantedBy=default.target` in the unit handles autostart.

If you can revert the `webserver.port` change, switch to `AddDevice=-/dev/snd`, and apply the working bits to `Podman/quadlet/*.container` as well, I'll get this merged.

barmst02:
Thank you for the response and explanation! I have one question about your comments. I see that I mistakenly changed the port used by the app inside the container to 8888. My intention was to change the port mapped on the host to 8888 because a component of Podman (Pasta) listens on port 8080. That causes the prechecks to fail.

So I've made the change to AddDevice, updated the *.container files, and reverted most of the port changes. However I've left the host port as 8888 for the reason above. 

If you still want me to change the host port back to 8080 I will, but I wanted to make sure I explained the reason clearly first.

barmst02:
I also wanted to point out that I had to manually create config and data directories in ~/.config/containers/systemd. I don't think this is the expected behavior, but I wasn't sure how to resolve in the code.

PR #3662, opened 2026-06-23
feat(ui): help guide user to model selection and downloading pages

alectrocute:
I observed some pretty brutal user friction. It's hard for some users to find where to configure which models are used for which streams. This PR adds a small help message on the `SystemInference` component reminding the user where to find the model configuration for both configuring per-stream and downloading more models.

English-only a11y translation provided, sorry.

Screenshot 2026-06-23 at 1 46 55 PM


tphakala:
Hi @alectrocute, a quick licensing note before this gets merged.

tphakala:

PR #3678, opened 2026-06-24
test(processor): make TestRaceCondition_ProposedSolutionValidation deterministic with synctest

tphakala:
## Problem

`TestRaceCondition_ProposedSolutionValidation` in `internal/analysis/processor` is flaky. It asserts a hardcoded wall-clock ceiling:

```go
assert.LessOrEqual(t, sseDuration, 100*time.Millisecond, "SSE action took too long ...")
```

on an action that performs a real ~50ms `time.NewTimer` sleep, while the test runs under `t.Parallel()`. The 100ms bound leaves only ~50ms of scheduling slack, so under `-race`, single-core scheduling (`-cpu=1`), and CPU contention the timer wakeup plus goroutine scheduling slips past 100ms (observed 121ms and 169ms locally), producing a spurious FAIL. The test passes in isolation and on idle multi-core runs, so it does not fail on `main` under the usual `go test -race`; it only trips on loaded or slow runners.

This is the same wall-clock-bound anti-pattern as the previously fixed jobqueue and notification timing flakes: the bound measures runner/scheduler overhead, not behavior. No data race is reported, it is purely the timing assertion.

## Fix

Wrap the test body in `testing/synctest` (the project-preferred approach over real sleeps) so the action delays run on a virtual clock. When the only goroutine in the bubble blocks on its timer, time advances by exactly the configured delay, so the measured durations are exact and deterministic:

- database action: exactly 200ms
- SSE action: exactly 50ms

with zero scheduling slack. The flake becomes structurally impossible.

The flaky `LessOrEqual` / `GreaterOrEqual` bounds are replaced with exact `assert.Equal` checks against named delay constants, plus an explicit ordering assertion that the SSE action completes after the database action.

The test stays meaningful: it still validates that sequential execution prevents the race (the SSE action only waits its own delay, not the database chain) and that ordering holds. Confirmed that a behavioral regression still fails the test (changing the SSE action's actual delay to 70ms while the assertion expects 50ms produces `Not equal: expected 50ms actual 70ms`).

## Verification

- Target test x30 under `-cpu=1` with full CPU saturation: zero flakes (previously reproduced 121ms and 169ms failures under the same conditions).
- Both reported shuffle seeds under `-cpu=1` + CPU saturation: pass.
- Full package `go test -race ./internal/analysis/processor/`: green.
- Repeated `-race -shuffle=on -count=2` rounds: all green.
- `go vet` and `golangci-lint run` on the package: 0 issues.

## Preflight Status

- [x] Single concern: one test-only flake fix
- [x] Preflight passed: parallel correctness/quality review plus secondary-model (Gemini 3.1 Pro) cross-validation, no findings requiring changes
- [x] Linters clean: `golangci-lint run` on the package reports 0 issues (full-project run shows only the pre-existing, unrelated `frontend/embed.go` typecheck error from an unbuilt frontend `dist/` in this environment)
- [x] Tests pass: `go test -race ./internal/analysis/processor/` passes
- [x] No unrelated changes: diff is a single test function in one file
- [x] No regression / backward-compat break: test-only change, no production code, config, API, DB, or CLI surface touched
- [x] No secrets or PII

Note: sibling tests in the same file still use real-sleep plus wall-clock bounds but with a much larger 500ms tolerance (lower flake risk). Converting those is out of scope for this fix.

### DIFF
--- a/internal/analysis/processor/race_condition_simple_test.go
+++ b/internal/analysis/processor/race_condition_simple_test.go
@@ -19,6 +19,7 @@ package processor
 import (
 	"sync"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


tphakala:
Thanks, but this one is a false positive. `testing/synctest` does export `Test`: `synctest.Test(t, func(t *testing.T))` is the stable API added in Go 1.25 (the older `synctest.Run(f)` was the Go 1.24 experimental form). This repo targets Go 1.26, and the package already uses `synctest.Test` in several places (for example internal/telemetry/worker_test.go and internal/notification/circuit_breaker_test.go). The official Go docs use exactly this form too.

The code compiles and the tests pass: `go build ./internal/analysis/processor/` succeeds and the target test runs 30+ times green under -race -cpu=1. So no change is needed here.

PR #3689, opened 2026-06-26
fix(spectrogram): resample bat/bird clips in the FFmpeg-only fallback

tphakala:
## Summary

Follow-up to #3529. The Sox-failure **FFmpeg-only fallback** (`generateWithFFmpeg`) rendered spectrograms at the clip's native sample rate, ignoring the frequency profile. The Sox paths resample (bird to 24 kHz, bat to 256 kHz) so the spectrogram's frequency axis matches the fixed UI overlay (0-12 kHz for bird, 0-128 kHz for bat); the FFmpeg-only fallback did not.

Consequence: a bat clip captured at 192/384 kHz that fell back to FFmpeg rendered up to its native Nyquist while the player labelled the axis 0-128 kHz, and that axis-mismatched image was then cached permanently under the `-bat` filename and served from the fast path. This broke the "always 0-128 kHz regardless of capture rate" contract introduced in #3529. The bird profile had the same latent gap on this path (a 48 kHz clip rendered 0-24 kHz under a 0-12 kHz axis).

The fallback only triggers when Sox is unavailable or fails, so working-Sox setups (the common case) were never affected.

## Changes

- **`generator.go`**: extract `buildFFmpegSpectrogramFilter` and prepend an `aresample=<rate>` stage to the `showspectrumpic` filtergraph when the profile sets a resample rate, so the FFmpeg backend produces the same frequency axis as the Sox `rate` effect (aresample sets the input rate, so the axis tops out at its Nyquist = rate/2).
- **`generator_test.go`**: table test covering the bird (24 kHz), bat (256 kHz), and native (no resample) cases.
- **`media.go`**: route `ProcessedSpectrogramByID` and `ServeSpectrogramByID` through the existing `resolveDetectionFrequencyProfile` helper, dropping two inline copies of the `GetNoteModelType` + warn-log + `ProfileForModelType` block.
- Refresh stale comments that still claimed bat detections keep their native rate (`BatProfile` resamples to 256 kHz since the bat profile was re-enabled in #3529), and an en-dash to hyphen sweep in `frequency_profile.go`.

## Test Plan

- [x] `go test ./internal/spectrogram/ -race -count=1` passes (new test included)
- [x] `go test ./internal/api/v2/ -race -count=1` (spectrogram/detection subset) passes
- [x] `golangci-lint run` (full project) reports 0 issues
- [ ] Manual: render a bat detection's spectrogram with Sox disabled and confirm the 0-128 kHz axis matches the rendered content

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spectrogram generation so the displayed frequency range now stays consistent across processing paths, reducing mismatches between different spectrogram outputs.
  * Updated spectrogram frequency handling for better alignment with the selected profile, helping results render more reliably for bird and bat audio.

* **Style**
  * Standardized spectrogram axis documentation text for consistency.

### DIFF
--- a/internal/api/v2/media.go
+++ b/internal/api/v2/media.go
@@ -1077,14 +1077,9 @@ func (c *Controller) ProcessedSpectrogramByID(ctx echo.Context) error {
 
 	params := c.parseSpectrogramParameters(ctx)
 
-	// Resolve frequency profile from detection's model type
-	modelType, mtErr := c.DS.GetNoteModelType(noteID)
-	if mtErr != nil {
-		c.logWarnIfEnabled("GetNoteModelType failed, defaulting to bird profile",
-			logger.String("note_id", noteID),
-			logger.Error(mtErr))
-	}
-	profileOpt := spectrogram.WithFrequencyProfile(spectrogram.ProfileForModelType(modelType))
+	// Resolve frequency profile from detection's model type (the helper logs a
+	// warning and falls back to the bird profile when the lookup fails).
+	profileOpt := spectrogram.WithFrequencyProfile(c.resolveDetectionFrequencyProfile(noteID))
 
 	if err := c.spectrogramGenerator.GenerateFromFile(ctx.Request().Context(), tmpPath, tmpSpectrogramPath, params.width, params.raw, profileOpt); err != nil {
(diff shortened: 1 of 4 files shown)


tphakala:
@coderabbitai review

PR #3737, opened 2026-06-28
feat(imports): native BirdNET-Pi import - API endpoints + wizard

tphakala:
Final PR of the native (non-container) BirdNET-Pi import feature, following the scanner (#3732) and staging backend + elevation ladder (#3733).

## What this adds
- **API (internal/api/v2/imports):** three authenticated endpoints
  - `GET /import/sources` - per-environment discovery (candidates + run-as identity + guidance)
  - `POST /import/validate` - validate a manually entered path (found / not-found / unreadable / invalid)
  - `POST /import/elevate` - wires the sudo elevation ladder over HTTP: direct read -> passwordless sudo -> in-app password -> copy-paste fallback, then launches the import from the staged copy
- **Wizard (BirdNetPiImportWizard.svelte):** rebuilt source step, candidate-driven, with three states (readable candidates / found-but-unreadable -> elevation panel with a plain-HTTP password warning / zero-candidates -> guidance + manual entry). Container and native both flow through the same candidate model. i18n across all 16 locales.
- Disk-space preflight before staging, staging dir created under a root-owned sticky base with a crypto-random name, and staging cleanup tied to the import job lifecycle.

## Safety
- The in-app sudo password is memory-only, redacted in all string/JSON/slog forms, cleared on every path, single-use, fed newline-terminated to `sudo -S`, never logged or sent to telemetry; an embedded NUL is rejected. `/import/elevate` is rate-limited per client IP.
- Container path confinement preserved (absolute-within-mount accepted with strict containment); native source paths are symlink-resolved before the root subcommand opens them.
- Elevation is native-only; container unreadable mounts get a host-permission hint instead.

## Review
Pre-push: full multi-agent gate (reuse / correctness / quality / integration / regression / i18n / test-quality / UX) + Gemini cross-validation, plus a dedicated OWASP ASVS pass on the privileged surface (0 critical/high; rate-limit, symlink-resolve, and NUL-reject hardening applied). `go test -race ./...`, `golangci-lint run`, and `npm run check:all` all green locally. The container-import regression the gate caught (candidate absolute path rejected) is fixed and covered by tests.

CI matrix includes linux/macos-arm64/windows-amd64; the Linux-only syscall paths are build-tagged with non-Linux stubs.

## Summary by CodeRabbit

* **New Features**
  * Updated the BirdNET-Pi import wizard with an automatic source discovery step, manual `birds.db` path entry/validation, and clearer guidance when no candidates are found.
  * Added a permissions elevation flow for unreadable sources with admin password prompts, an unencrypted-connection warning, fallback commands, and progression to import status.
* **Accessibility**
  * Added optional `aria-describedby` support for improved screen-reader context on text inputs.
* **Bug Fixes**
  * Strengthened import source/path validation, unreadable and cancelled request handling, and staging cleanup to reduce failed or leftover imports.
* **Documentation**
  * Documented the new API v2 import source discovery, validation, and elevation endpoints.
* **Tests**
  * Expanded unit/integration coverage for the new discovery and elevation flows.


tphakala:
Review feedback addressed in 253e9454 (backend) and 0d825a5a (frontend test):

**Fixed**
- `probe.go` (CodeRabbit Major + Gemini): `Probe` now uses `os.Stat` instead of `os.Lstat`, so a symlinked `birds.db` is supported and a path permission error is preserved as `permission_denied` rather than collapsing to `not_found`; the `!IsRegular()` check still rejects FIFO/socket/device targets (and symlinks to them) before any open. Tests updated; a unix-only symlink-to-FIFO rejection test was added.
- `sources_test.go` (CodeRabbit Major): the `ValidateImportSource` tests are pinned to native mode so per-environment path resolution is deterministic.
- `elevate.go` (CodeRabbit Minor): staging allocation + disk preflight are skipped for a directly-readable source, so a readable import is no longer gated on staging free space.
- `handler.go` (CodeRabbit Minor): corrected the `stagingBase` field comment.
- `BirdNetPiImportWizard.test.ts` (CodeRabbit Minor): `globalThis.location` is now saved and restored around the HTTP-warning test.

**Respectfully pushing back (no change)**
- Gemini, `guidance` null access (lines ~619/~658): already guarded - the steps access is `sourcesResponse?.guidance?.steps && ...` (short-circuits), and `run_as_uid` is inside `{#if sourcesResponse?.containerized}`. svelte-check is 0 errors.
- CodeRabbit, loading vs zero-candidates in `utils.ts`: a separate `{#if isLoading && currentStep === 'source'}` spinner guard renders before `sourceStepState` is ever set, so the two are never the same visible state.

Local gates green: `go test -race ./...`, `golangci-lint run` (0 issues), `npm run check:all`.

Issue #36, opened 2024-02-21
Multiple nodes, single MySQL database


PeteLawrence:
I have setup a single MySQL database with 2 BirdNet-Go nodes sending their detections to that database. I noticed that there's a source_node field in the database so thought the web interface for each node would show only their respective detections, however the web interface on both nodes shows all detections, along with broken spectrograms for detections from the other node as the wav/png file doesn't exist on that node.

Should the web interface show only that nodes detections, or is it meant to show detections from all nodes?

tphakala:
I have planned to implement support for multinode setup, node name and shared database support is part of that, but it also is all I have done about it yet. So any plans and implementation further is still to be done.

PeteLawrence:
Would you be interested in a pull request that adds in station name filtering to the relevant DB queries?

PR #3764, opened 2026-07-01
[codex] fix BirdWeather FLAC upload durations

kivster:
## Summary

BirdWeather soundscape uploads were encoded as FLAC streams directly into an in-memory pipe/buffer. For FLAC, non-seekable output leaves `STREAMINFO.total_samples` unknown. BirdWeather then records the soundscape duration as `0.0`, which can leave the website audio panel stuck loading even though the media file exists.

This change makes BirdWeather upload encoding write to a private seekable temp file first, then reads the finalized FLAC bytes back into the existing upload buffer. That lets both the default FFmpeg path and the opt-in native encoder patch FLAC metadata before POSTing to BirdWeather.

## Changes

- Add a BirdWeather upload helper that encodes through a seekable temporary file and cleans it up after reading.
- Switch default FFmpeg BirdWeather FLAC uploads from `ExportAudioToBuffer` to the existing file-backed `ExportAudio` path.
- Switch native BirdWeather FLAC uploads to file-backed `flac.EncodePCM` for finalized `STREAMINFO` metadata.
- Add regression assertions that parse FLAC `STREAMINFO` and verify non-zero finalized total samples.
- Update FLAC package comments so BirdWeather is no longer described as using the non-seekable buffer encoder.

## Validation

- `PATH=/tmp/go1.26.1/bin:/opt/homebrew/bin:$PATH go test ./internal/birdweather`
- `PATH=/tmp/go1.26.1/bin:/opt/homebrew/bin:$PATH go test ./internal/audiocore/flac ./internal/audiocore/ffmpeg ./internal/birdweather`
- `PATH=/tmp/go1.26.1/bin:/opt/homebrew/bin:$PATH BIRDNET_FLAC_ENCODER=native go test ./internal/birdweather`

## Summary by CodeRabbit

* **New Features**
  * FLAC audio exports now preserve stream metadata (e.g., sample rate and total samples) by generating seekable FLAC before returning/uploading the final bytes.

* **Bug Fixes**
  * Improved FLAC encoding consistency across normal and fallback paths.
  * Fixed validation to ensure the encoded FLAC stream reports expected stream information.

* **Documentation**
  * Updated BirdWeather audio upload docs to reflect seekable FLAC temporary-file behavior.

* **Tests**
  * Expanded FLAC encoding tests to assert STREAMINFO sample rate and total sample counts.

### DIFF
--- a/internal/audiocore/flac/doc.go
+++ b/internal/audiocore/flac/doc.go
@@ -1,7 +1,7 @@
 // Package flac provides native, FFmpeg-free FLAC encoding of audio clips using
 // github.com/tphakala/go-flac. The native path is selected at runtime by the
 // BIRDNET_FLAC_ENCODER environment variable; when it is not selected the caller
-// keeps using the FFmpeg-based exporter. EncodePCM writes a FLAC file (the
-// detection save path); EncodePCMToBuffer returns FLAC bytes in memory (the
-// BirdWeather soundscape upload path).
+// keeps using the FFmpeg-based exporter. EncodePCM writes a seekable FLAC file;
+// EncodePCMToBuffer returns FLAC bytes in memory for callers that do not need
+// finalized seek metadata.
 package flac
--- a/internal/audiocore/flac/gate.go
+++ b/internal/audiocore/flac/gate.go
@@ -10,9 +10,9 @@ import (
 // unset) keeps the FFmpeg exporter.
 //
 // This is a temporary opt-in gate shared by every native FLAC path: the
-// detection save path (EncodePCM) and the BirdWeather soundscape upload path
-// (EncodePCMToBuffer + audionorm normalization). When the native encoder is
-// trusted, the gate is removed and FLAC always routes to the native path. The
+// detection save path (EncodePCM) and the BirdWeather soundscape upload path.
+// When the native encoder is trusted, the gate is removed and FLAC always routes
+// to the native path. The
 // detection save path still falls back to FFmpeg when EBU R128 normalization is
 // enabled; the BirdWeather path normalizes natively via audionorm. Keep the read
 // confined to NativeEncoderEnabled so the gate is a single deletion site.
--- a/internal/birdweather/README.md
+++ b/internal/birdweather/README.md
@@ -63,8 +63,10 @@ func encodePCMtoWAV(pcmData []byte) (*bytes.Buffer, error)
 BirdWeather FLAC uploads use a short-lived, disk-backed temporary file before
 the upload buffer is created. FLAC muxers need seekable output to finalize
 `STREAMINFO` metadata such as the total sample count; without that metadata
-BirdWeather can store the soundscape duration as `0.0`. The temporary file is
-removed after the finalized FLAC bytes are read into the upload buffer.
+BirdWeather can store the soundscape duration as `0.0`. The upload path creates
+the private temp directory with `os.MkdirTemp`, encodes the FLAC there, reads
+the finalized bytes with `os.ReadFile`, and removes the temporary file
+afterward.
 
 ### Loudness Normalization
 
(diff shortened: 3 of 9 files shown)


tphakala:
Thanks for tracking this down. The diagnosis is spot on: FLAC leaves `STREAMINFO.total_samples` at the "unknown" sentinel when the muxer can't seek back over the header, and BirdWeather then stores the soundscape duration as `0.0`. I reviewed the whole change and it is correct and ship-safe: the temp lifecycle is leak-free and race-free (unique dir per upload, no shared temp name), all three upload paths (loudnorm 2-pass, fallback gain, native) are converted, and the STREAMINFO test is a genuine regression guard that would fail on the old code. Build and tests pass.

A couple of things before this lands, plus one follow-up I will handle myself.

**1. The native path shouldn't really need a temp file.**

For the FFmpeg path the temp file is unavoidable: ffmpeg can't finalize STREAMINFO on a pipe, so a seekable output file is the only option. That part is correct where it is.

The native go-flac path is different. It already holds the entire PCM buffer in memory, so it knows the exact sample count up front (`len(pcm) / (channels * bytesPerSample)`). Routing it through disk just so go-flac can seek back and write a number it could have computed at init is the wrong layer, and it reintroduces a per-upload disk write that the README specifically called out as SD-card-unfriendly (this PR removes that rationale). It also stacks an outer temp dir on top of `flac.EncodePCM`, which already does its own unique-temp + atomic-rename + fsync, so the native path ends up writing two temp files and fsyncing one it is about to delete.

The clean fix belongs upstream in go-flac, so the encoder can write a known `total_samples` for a non-seekable sink. I filed that:
- tphakala/go-flac#36 (write a known `total_samples` up front for non-seekable writers)
- tphakala/go-flac#37 (finalize STREAMINFO for the whole-buffer `EncodeInterleaved` path)

Once that lands I will wire it back here so the native path encodes straight to a `bytes.Buffer` again with correct duration and no disk I/O. Leaving the native path on the temp-file route in this PR is fine as a stopgap; I will remove it in the follow-up.

**2. Dead code + stale docs.**

This PR removed the only production caller of `flac.EncodePCMToBuffer` (and of `ffmpeg.ExportAudioToBuffer`). They are now referenced only by their own tests, and `flac/doc.go` / `flac/encode.go` still describe a "BirdWeather" use that no longer exists. I want to keep `EncodePCMToBuffer` for the go-flac follow-up (it becomes the mechanism again once the encoder can set `total_samples` up front), so please do not delete it, but do update the stale comments in `doc.go` and `encode.go` so they stop claiming a caller that is gone. The `gate.go` comment reflow also left a dangling "The" mid-sentence.

**Minor, non-blocking:**
- Gain precision moved from `%.2f` to `%.1f` (about 0.05 dB, imperceptible).
- The native path now emits a SEEKTABLE into short upload clips (harmless but unnecessary for a soundscape).
- New failure surface if `/tmp` is full or read-only, and the FFmpeg failure Sentry `operation` tags shift from `export_buffer_*` to `export_*`. Worth a release-note line about the per-upload temp write.

Happy to take the go-flac side and the in-memory rewire as a follow-up so this can merge. Nice work on the root-cause analysis.

tphakala:
One more bit of direction so the tradeoffs here make sense.

I am planning to promote the native go-flac encoder to the primary FLAC path and drop the current `BIRDNET_FLAC_ENCODER=native` opt-in gate once it has enough real-world mileage (the gate was always meant to be temporary). After that the FFmpeg FLAC code goes away entirely: the BirdWeather FFmpeg FLAC upload path here, and the FFmpeg FLAC fallback on the detection-save path. FLAC would then be pure Go with no FFmpeg dependency at all.

That reframes this PR a bit. The FFmpeg temp-file path it adds is transitional, so I would rather not invest in optimizing it. The piece that actually matters long-term is the native path, and go-flac#36 is what lets it stay fully in-memory (correct `total_samples` written up front, no temp file, no disk I/O on SD-card systems). Once #36 lands and native is the default, the whole FFmpeg-FLAC branch, including the temp-file round-trip, gets deleted rather than kept.

So: happy to merge this as the stopgap that fixes the duration bug today, and I will fold the FFmpeg-FLAC removal into the native-primary switch.

PR #3834, opened 2026-07-09
perf(api): skip name resolution on species scores when names=false

Bruick:
## Summary

`GET /api/v2/range/species/scores` resolves localized common names for **every** geomodel species (~12k) on each call via `resolveLocalizedName`. That per-species resolution dominates the endpoint's latency and, when all species are requested (threshold 0), can make the endpoint effectively hang.

This adds an opt-in `names=false` query parameter that returns `scientificName` → `score` only, skipping common-name resolution. Names are included by default, so existing callers are unaffected.

## Changes

- Add `names=false` query parameter to `GetRangeFilterSpeciesScores`.
- Add `convertSpeciesScoresNoNames`, which builds the score list without resolving localized common names.
- Unit tests for the new helper.

## Testing

- [x] `go test -race ./internal/api/v2/range/...`
- [x] `golangci-lint run --build-tags=noembed,skipfrontend ./internal/api/v2/range/...` (0 issues)
- [x] Manual: `GET /api/v2/range/species/scores?names=false` returns 200 quickly with all species; the default (names included) response is unchanged.

## Notes

Standalone performance improvement. A follow-up PR (rare-species highlighting) will consume the `names=false` path to flag rare detections client-side without an extra per-row call.

## Summary by CodeRabbit

* **New Features**
  * Added an option to return species scores without localized/common names for lighter responses.
  * Kept the default response behavior unchanged when names are requested.

* **Bug Fixes**
  * Improved species score conversion so each result keeps its own score value correctly.
  * Ensured empty inputs return an empty species list.

### DIFF
--- a/internal/api/v2/range/range.go
+++ b/internal/api/v2/range/range.go
@@ -124,16 +124,36 @@ func resolveLocalizedName(resolver *classifier.Orchestrator, locale string, sp d
 // response format with probability score pointers. When resolver and locale
 // are provided, common names are resolved to the user's locale.
 func convertSpeciesScores(scores []classifier.SpeciesScore, resolver *classifier.Orchestrator, locale string) []dto.RangeFilterSpecies {
+	return buildRangeFilterSpecies(scores, func(label string) string {
+		return resolveLocalizedName(resolver, locale, detection.ParseSpeciesString(label))
+	})
+}
+
+// convertSpeciesScoresNoNames converts geomodel scores without resolving localized
+// common names. Name resolution is the dominant cost when converting all geomodel
+// species (threshold 0), so callers that only need scientificName->score skip it.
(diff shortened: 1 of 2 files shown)


Bruick:
Re: the nitpick to use `assert.Equal` for the `Score` float fixtures in `internal/api/v2/range/range_scores_names_test.go` — keeping `assert.InDelta`.

The repo's `testifylint` `float-compare` rule rejects `assert.Equal` on floats, so exact-equality fails the mandatory lint gate:

```
internal/api/v2/range/range_scores_names_test.go:38:2: float-compare: use assert.InEpsilon (or InDelta) (testifylint)
```

`InDelta` with a `1e-9` tolerance keeps the invariant tight while satisfying the linter.

tphakala:
Thank you!

PR #3841, opened 2026-07-10
perf(datastore): label-first lookup in GetSpeciesLastDetectionDateBefore

### LINKED ISSUE #3840: bug: species tracking does a full-table scan per species (GetSpeciesLastDetectionDateBefore)
## Environment
- BirdNET-Go: nightly-20260615 (build_date 2026-06-15)
- Device: Raspberry Pi 5, 8 GB RAM, storage on microSD
- Install: Docker (ghcr.io/tphakala/birdnet-go:nightly)
- DB: SQLite, ~1.7M detections (~350 MB), migrated from an older BirdNET

## Description
With Species / Yearly / Seasonal tracking enabled on a large SQLite database, the log fills with `slow query` warnings (and, on slow storage, occasional detection-save timeouts). With tracking off, everything runs smoothly.

## Root cause
`GetSpeciesLastDetectionDateBefore` (internal/datastore/v2only/datastore.go) joins `labels` and filters a single `scientific_name`, but also filters `detected_at < <now>`. SQLite drives off the `detected_at` range index (~95% of the table) instead of resolving the highly selective species to a `label_id` and using the `(label_id, detected_at)` index. It is called once per species during novelty-episode restore (internal/analysis/species/database.go), so the cost is O(species x full-table scan).

## Measurements (my DB, 1,726,563 detections)
Current plan:
SEARCH d USING INDEX idx_detections_detected_at (detected_at<?) <-- ~1.65M rows
BLOOM FILTER ON l (id=?)
SEARCH l USING INTEGER PRIMARY KEY (rowid=?)
SEARCH dr USING INDEX idx_detection_reviews_detection_id (detection_id=?) LEFT-JOIN

- rows matching `detected_at < now`: 1,646,225 of 1,726,563 (95%)
- rows for the species actually queried (e.g. Myiarchus crinitus): 193
- current query: ~1.4s under load (trips the 1s slow-query threshold)
- `ANALYZE` does not change the plan
Resolving the species to `label_id` first flips it to a covering-index lookup and returns the identical result in ~0.002s (~200x faster):
SEARCH d USING COVERING INDEX idx_detections_label_detected (label_id=? AND detected_at<?)
LIST SUBQUERY 1
SCAN labels USING COVERING INDEX idx_labels_scientific_name
SEARCH dr USING INDEX idx_detection_reviews_detection_id (detection_id=?) LEFT-JOIN

## Steps to reproduce
1. Large SQLite DB (~1M+ detections).
2. Enable Species + Yearly/Seasonal tracking.
3. Watch `datastore` logs: repeated `slow query` on the per-species MAX(date(detected_at ...)) query.
## Proposed fix
Resolve `scientific_name` (incl. the legacy `Name_Common` LIKE form) to `label_id`(s) in a subquery, then filter `detections` by `label_id IN (...)` so the planner uses the `(label_id, detected_at)` index. Identical LIKE/ESCAPE semantics; existing behavioural tests pass. PR to follow.
Note: only this single-species function has the pathology. Sibling queries (GetSpeciesDetectionDatesInPeriod, GetSpeciesDiversityData) GROUP BY all species across a bounded period, so a range scan is correct there.

Kerakis:
## Summary
`GetSpeciesLastDetectionDateBefore` does a full-table scan per species when Species/Yearly/Seasonal tracking is enabled. On large SQLite databases this produces a `slow query` storm.

## Root cause
The query joins `labels` and filters a single `scientific_name`, but also filters `detected_at < <now>`. SQLite drives off the `detected_at` range index (~95% of the table) instead of resolving the highly selective species to a `label_id` and using the `(label_id, detected_at)` index. The novelty-episode restore path calls this once per species, so the cost is O(species x full-table scan).

## Fix
Resolve `scientific_name` (including the legacy `Name_Common` LIKE form) to `label_id`(s) in a subquery, then filter `detections` by `label_id IN (...)`. The planner then uses the `(label_id, detected_at)` covering index. Behaviour and LIKE/ESCAPE semantics are unchanged.

## Measurements (1.7M-row DB)
- Rows matching `detected_at < now`: ~95% of the table
- Rows for a typical single species: a few hundred
- Before: ~1.4s per call (trips the 1s slow-query threshold)
- After: ~0.002s, identical result; `EXPLAIN QUERY PLAN` switches to a covering-index lookup

## Testing
- `go test ./internal/datastore/v2only/ ./internal/analysis/species/` passes (incl. LIKE/ESCAPE edge cases and novelty tests)
- Verified on a real 1.7M-row database: identical result, ~200x faster

## Related Issues
Fixes #3840

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy when determining the last detection date for a species.
  * Preserved filtering for detections recorded before a specified date and excluded false-positive reviews.

### DIFF
--- a/internal/datastore/v2only/datastore.go
+++ b/internal/datastore/v2only/datastore.go
@@ -2956,18 +2956,30 @@ func (ds *Datastore) GetSpeciesLastDetectionDateBefore(ctx context.Context, scie
 	// Escape LIKE metacharacters with '!' (see scientificNameLikeEscaper).
 	escapedScientificName := scientificNameLikeEscaper.Replace(scientificName)
 	prefix := ds.manager.TablePrefix()
+
+	// Resolve the species to its label id(s) in a subquery so the planner drives
+	// off the (label_id, detected_at) index. Joining labels and filtering
+	// scientific_name inline instead makes SQLite drive off the near-unbounded
+	// "detected_at < before" range, scanning almost the entire detections table
+	// for a single, highly selective species — this function is called once per
+	// species during novelty-episode restore, so that plan is O(species * rows).
+	//
+	// Match the bare scientific name exactly, or a legacy concatenated label
+	// stored as "ScientificName_CommonName". The "!_%" suffix is "literal
+	// underscore separator, then anything" ('!_' is an escaped underscore,
+	// '%' is the wildcard), mirroring how such labels are split on the first
+	// underscore (see detection.ExtractScientificName).
+	labelIDs := ds.manager.DB().WithContext(ctx).
+		Table(prefix+"labels").
+		Select("id").
+		Where("scientific_name = ? OR scientific_name LIKE ? ESCAPE '!'", scientificName, escapedScientificName+`!_%`)
+
 	query := ds.manager.DB().WithContext(ctx).
 		Table(prefix+"detections d").
 		Select(fmt.Sprintf("COALESCE(MAX(%s), '') as last_seen_date", dateExpr)).
-		Joins(fmt.Sprintf("JOIN %slabels l ON d.label_id = l.id", prefix)).
 		Joins(fmt.Sprintf("LEFT JOIN %sdetection_reviews dr ON d.id = dr.detection_id", prefix)).
 		Where("d.detected_at < ?", before.Unix()).
-		// Match the bare scientific name exactly, or a legacy concatenated label
-		// stored as "ScientificName_CommonName". The "!_%" suffix is "literal
-		// underscore separator, then anything" ('!_' is an escaped underscore,
-		// '%' is the wildcard), mirroring how such labels are split on the first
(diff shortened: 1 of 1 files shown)


tphakala:
Thanks for the sharp diagnosis here, and for including the query-plan analysis plus the before/after numbers on a real 1.7M-row DB. This is a genuine problem and you nailed the root cause: the inline labels join made SQLite drive off the near-unbounded `detected_at` range instead of resolving the species to a `label_id` and using the `(label_id, detected_at)` index.

The catch is that the same fix already landed on main in #3842 (merged earlier today), which swaps the join for a `label_id` subquery for exactly that reason. Your PR does the same thing, just with a GORM subquery builder instead of raw SQL, so it's now a duplicate of what's on main and conflicts with it.

Closing this as superseded, but full credit for catching it first. Really appreciate the well-researched report.

PR #3847, opened 2026-07-10
fix(spectrogram): swap FFmpeg fallback color modes so default style renders colorful

### LINKED ISSUE #3835: bug: Spectograms of MP3s are shown in black and white

nightly-20260615

### Install method

Native binary (downloaded release, no container)

### Device type

PC / Server (x86_64)

Windows 11

### Audio source

USB or built-in sound card (local capture device)

### Hardware details

Windows 11 Desktop PC, 64GB RAM, HD with enough free space, Boya BY-M3 USB microphone

### Description

When I set clip storage file type to MP3 spectrograms are no longer rendered in the chosen colour scheme but in a simple black and white form. If this is intended/expected behaviour a sentence explaining that in the UI where you chose the file format would be nice.

Version: nightly-20260615
Build: 2026-06-15T15:08:40Z
OS: Microsoft Windows
Architecture: x86_64
Environment: Native
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:152.0) Gecko/20100101 Firefox/152.

### Steps to reproduce

1. Set filetype to MP3 (Settings/Audio/Recording/File Settings/Audio Format: MP3)
Image

2. Save settings
3. Wait for new capture
4. Its spectogram will be black and white
Image

### Expected behaviour

It has the colour scheme chosen in Settings/Interface/Visual Content/Spectrogram Style

### Actual behaviour

Its black and white only.

### When did this start?

First time setting up BirdNET-Go (never worked)

### Relevant log output

```shell

```

Added via web ui feature.

_No response_

### Pre-submission checks

### Additional context

_No response_

pelazas:
Fixes #3835

When clip generation falls back to FFmpeg (the reporter's case: MP3 clips on the native Windows build, where the bundled SoX can't decode MP3), the default spectrogram style renders black and white instead of the colorful style — and, less visibly, the scientific styles render *colorful* through the same fallback.

## Root cause

The FFmpeg color mode constants in `internal/spectrogram/generator.go` had `showspectrumpic` semantics backwards. `ffmpegColorDefault` was `"channel"` with a comment calling it "FFmpeg's default colorful spectrogram", and scientific styles mapped to `"intensity"` commented as grayscale. It's the other way around: `intensity` is the colorful heat map (and the filter's own default), while `channel` assigns one color per *audio channel* — and since the capture pipeline is mono (`conf.NumChannels = 1`), that renders every clip in a single hue, i.e. black and white.

Easy to reproduce with just ffmpeg:

```
ffmpeg -f lavfi -i "sine=frequency=440:duration=3" -lavfi "showspectrumpic=s=400x200:color=channel" ch.png
ffmpeg -f lavfi -i "sine=frequency=440:duration=3" -lavfi "showspectrumpic=s=400x200:color=intensity" int.png
```

Measuring saturation of the outputs (`ffprobe -f lavfi "movie=X.png,signalstats" -show_entries frame_tags=lavfi.signalstats.SATAVG`): `color=channel` → SATAVG 0.04 (black and white), `color=intensity` → SATAVG 4.84 (colorful). `ffmpeg -h filter=showspectrumpic` confirms the same: default is `intensity`, `channel` is "separate color for each channel".

## The fix

- Swap the mapping: default style → `intensity`, scientific styles → `channel`.
- Rename `ffmpegColorIntensity` → `ffmpegColorGrayscale` so the name matches what it actually produces, and fix the backwards comments.
- Pin **literal** color values in `TestGetFFmpegColorMode` — the old test compared against the constants themselves, so it would happily pass with swapped values. The literal expectations fail on the old code.
- While in that test file: `TestFFmpegFallback_AppliesStyleSetting` was rebuilding the filter string with its own `fmt.Sprintf` copy, so it couldn't catch regressions in the real construction. It now calls `buildFFmpegSpectrogramFilter` directly.

Sox path is untouched, so anyone whose spectrograms already render correctly sees no change.

## One caveat worth knowing

Spectrogram PNGs are cached by filename, and the filename doesn't encode the color mode — so clips that were already rendered black-and-white through the fallback will keep their old image until the cached PNG is deleted. Only newly generated spectrograms pick up the fix. Didn't want to grow this PR with cache-busting; happy to look at that separately if you think it's worth it.

## Testing

- `go test -race ./internal/spectrogram/` — pass
- `golangci-lint run -v` — 0 issues
- Empirical FFmpeg verification as above (ffmpeg 8.x)

## Preflight Status

- [x] Single concern: one fix (FFmpeg fallback color mapping)
- [x] Preflight passed: all findings resolved (test tautology fixed, filter-string duplication removed) or documented (stale-cache caveat above)
- [x] Linters clean: `golangci-lint run -v` zero issues (one pre-existing local-only `frontend/embed.go` typecheck from missing `dist/`, not related)
- [x] Tests pass: `go test -race ./internal/spectrogram/`
- [x] No unrelated changes
- [x] Scope complete
- [x] No regression/backward-compat break: Sox path unchanged; FFmpeg fallback output changes only to match what the styles already promise (cache caveat documented above)
- [x] Breaking changes documented: n/a
- [x] i18n complete: no user-facing strings changed
- [x] No secrets or PII

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected spectrogram color modes to better match Sox style presets.
  * Updated Scientific and Scientific Dark styles to use grayscale rendering.
  * Preserved the expected fire-style rendering for High Contrast Dark.
  * Improved validation of color settings applied to generated spectrograms.

### DIFF
--- a/internal/spectrogram/generator.go
+++ b/internal/spectrogram/generator.go
@@ -125,13 +125,17 @@ func fftFriendlyHeight(width int) int {
 // FFmpeg showspectrumpic color mode constants for style mapping.
 // These are the closest FFmpeg equivalents to the Sox style presets.
 const (
-	// ffmpegColorDefault is FFmpeg's default colorful spectrogram (channel mode).
-	// Matches Sox default style (colorful with dark background).
-	ffmpegColorDefault = "channel"
+	// ffmpegColorDefault is FFmpeg's colorful heat map ("intensity", the
+	// showspectrumpic default). Matches Sox default style (colorful with
+	// dark background). Not "channel": that mode assigns one color per
+	// audio channel, which renders mono clips in a single hue, i.e.
+	// black and white (issue #3835).
+	ffmpegColorDefault = "intensity"
 
-	// ffmpegColorIntensity produces a grayscale spectrogram in FFmpeg.
-	// Matches Sox monochrome mode (-m) used by scientific styles.
-	ffmpegColorIntensity = "intensity"
+	// ffmpegColorGrayscale produces a grayscale spectrogram for mono audio
+	// ("channel" assigns one color per audio channel). Closest FFmpeg
+	// equivalent to Sox monochrome mode (-m) used by scientific styles.
+	ffmpegColorGrayscale = "channel"
 
 	// ffmpegColorFire produces a high-saturation warm-toned spectrogram in FFmpeg.
 	// Matches Sox high-contrast mode (-h) used by high_contrast_dark style.
@@ -165,13 +169,13 @@ func getFFmpegColorMode(style string) string {
 	switch style {
 	case conf.SpectrogramStyleScientificDark:
 		// Grayscale - closest FFmpeg equivalent to Sox -m (monochrome)
-		return ffmpegColorIntensity
+		return ffmpegColorGrayscale
 	case conf.SpectrogramStyleHighContrastDark:
 		// High saturation - closest FFmpeg equivalent to Sox -h (high color)
 		return ffmpegColorFire
 	case conf.SpectrogramStyleScientific:
 		// Grayscale - closest FFmpeg equivalent to Sox -m -l (monochrome, light)
-		return ffmpegColorIntensity
+		return ffmpegColorGrayscale
 	default:
 		// Default colorful style
 		return ffmpegColorDefault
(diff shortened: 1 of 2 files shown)


tphakala:
Thank you!

PR #3870, opened 2026-07-11
fix(birdweather): guard soundscape gain against silent-clip +Inf

Kerakis:
## Summary

When a soundscape clip is silent, ffmpeg's `loudnorm` analysis reports
`input_i` as `"-inf"`. `parseDouble` parses that to `-Inf`, so
`gainNeeded = targetIntegratedLoudnessLUFS - inputLUFS` becomes `+Inf`. That
value is then:

1. **logged** as `calculated_gain` — a non-finite float is not JSON-encodable,
   so the structured log records the literal
   `"calculated_gain":"!ERROR:json: unsupported value: +Inf"`; and
2. **clamped to +30 dB**, i.e. the silent clip is amplified, lifting its noise
   floor instead of being left alone.

The newer native-FLAC path already handles this correctly — its comment notes
*"Silence yields GainDB == 0 … so quiet clips stay quiet instead of being
boosted into noise"* (`encode_native.go`). This brings the legacy ffmpeg path in
line.

## Changes

- Extract the loudness→gain mapping into `soundscapeGainDB(inputLUFS float64)`
  (`internal/birdweather/birdweather_client.go`), returning `0` dB for a
  non-finite measurement (leave the clip untouched); the existing clamp/log
  block is unchanged and now always sees a finite value.
- Unit test `TestSoundscapeGainDB` (`internal/birdweather/soundscape_gain_test.go`)
  covering `-Inf`/`+Inf`/`NaN` inputs and the finite mappings.

## Testing

- [x] Go tests pass (`go test ./internal/birdweather/ -run TestSoundscapeGainDB`)
- [x] `gofmt` clean
- [ ] Frontend tests pass (n/a — backend only)
- [x] Full-package pre-existing failures (`TestUploadSoundscape`, `TestPublish`,
      `…MalformedUploadResponse…`) require an `ffmpeg` binary and fail
      identically on clean `main` — unrelated to this change
- [ ] Preflight quality gate passed (run `/preflight` before pushing)

## Prior art

- The native-encoder path already returns 0 dB for silence (#3502 / #3777,
  merged) — its comment: *"Silence yields GainDB == 0 … so quiet clips stay
  quiet."* This PR brings the legacy ffmpeg path in line with that.
- #3671 (merged) hardened this file's upload transport and FLAC encoding but did
  not touch the loudness→gain calculation, which is still unguarded on `main`.

## Related Issues

_No open issue/PR describes the silent-clip +Inf gain; standalone fix._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio loudness normalization by safely handling non-finite loudness readings, preventing invalid gain values.
  * Updated loudness reporting to log non-finite values safely (e.g., `nan`, `+inf`, `-inf`) instead of raw floating-point output.
* **Tests**
  * Added unit tests covering gain calculation for finite and non-finite loudness inputs and verifying safe loudness string formatting.

### DIFF
--- a/internal/birdweather/birdweather_client.go
+++ b/internal/birdweather/birdweather_client.go
@@ -557,7 +557,7 @@ func (b *BwClient) encodeFlacUsingFFmpeg(ctx context.Context, pcmData []byte, ff
 
 	// --- Calculate gain needed to reach target loudness ---
 	inputLUFS := parseDouble(loudnessStats.InputI, -70.0)
-	gainNeeded := targetIntegratedLoudnessLUFS - inputLUFS
+	gainNeeded := soundscapeGainDB(inputLUFS)
 
 	// Apply safety limits to prevent excessive amplification or attenuation
 	maxGain := 30.0 // Maximum gain in dB (absolute value)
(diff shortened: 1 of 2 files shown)


tphakala:
Thank you!

Issue #3527, opened 2026-06-16
bug: Let's Encrypt (autotls) configuration not working


zandr:
### BirdNET-Go version / build date

nightly-20260615

### Install method

Docker via install.sh (recommended installer script)

### Device type

Raspberry Pi 5

### Operating system and version

Raspberry Pi OS Lite (64-bit, Trixie)

### Audio source

USB or built-in sound card (local capture device)

### Hardware details

8GB Ram
64GB MicroSD
USB Microphone

### Description

#3000 remains an issue for me, and I am unable to re-open the bug. I also asked the bot for help there, but apparently closed is closed, so I'm filing a new issue.

I have retested in 20260601, and 20260615, and there is no change in behavior. While #3319 may have fixed a real issue, it isn't the issue that I'm facing. As near as I can tell, my installation doesn't get as far as bringing up servers on 80 or 443, so there's no way I could be rate-limited by Let's Encrypt.

### Steps to reproduce

Install Birdnet-Go
Define external URL and hostname
Choose "Let's Encrypt"
Reboot Pi

### Expected behaviour

Ports 80 and 443 are open, 80 redirects to 443, SSL works.

### Actual behaviour

Only 8080 is open, running https, and failing because it has no cert.

### When did this start?

First time setting up BirdNET-Go (never worked)

### Relevant log output

```shell

```

### Pre-submission checks

- [x] I am running the [latest release](https://github.com/tphakala/birdnet-go/releases) or have tested with the `:dev` container image.
- [x] I have searched existing open and closed issues and this does not appear to be a duplicate.

zandr:
Dump uploaded with this issue number.

brycesub:
PR #3780 fixes this.

The root cause was exactly as diagnosed above: in AutoTLS mode the server called `StartAutoTLS` on the configured web port (`:8080`) instead of `:443`, so nothing listened on 443 for the ACME challenge and you were left with a certless HTTPS listener on 8080.

The PR:
- Binds AutoTLS HTTPS to `:443` and starts the HTTP→HTTPS redirect / ACME-challenge listener on `:80` (so `80` redirects to `443` and Let's Encrypt HTTP-01 validation can complete).
- Grants the container the ability to bind those privileged ports: `--cap-add NET_BIND_SERVICE` on the generated systemd unit plus a `setcap cap_net_bind_service=+ep` file capability on the binary (needed because the entrypoint drops to a non-root user via gosu).
[...]

PR #3879, opened 2026-07-12
fix(autotls): map installer/quadlet ports to 8080/8443 and serve /health over HTTP

tphakala:
## Summary

Closes the two remaining AutoTLS deployment gaps that #3612 left open. #3612 moved AutoTLS to bind the ACME HTTP-01 / redirect listener on container port 8080 and HTTPS on 8443, then remap host `80->8080` / `443->8443` at the host layer (so the container needs no `NET_BIND_SERVICE`). It updated the docker/podman compose files but missed two coupled artifacts, so AutoTLS was still broken for installer- and quadlet-based deployments, and the container healthcheck failed in AutoTLS mode.

Both gaps were surfaced by @brycesub in #3780; thanks for the digging.

## What was broken

1. **Installer and Podman quadlet mapped dead ports.** `install.sh` generated systemd units and `Podman/quadlet/birdnet-go-autotls.container` still published `80:80` / `443:443`. The container binds 8080/8443, so those host ports mapped to container ports where nothing listens: ACME HTTP-01 validation and HTTPS were both unreachable, and the quadlet still requested a now-pointless `NET_BIND_SERVICE`. `podman-install.sh` ships that quadlet verbatim, so every Podman-quadlet AutoTLS install was affected.
2. **AutoTLS + redirect broke the container healthcheck.** With `RedirectToHTTPS` on (the AutoTLS default), the ACME HTTP listener's fallback 308-redirected every path, including `/health`. The Docker `HEALTHCHECK` probes `http://localhost:8080/health | jq '.status == "healthy"'`; a 308 has no JSON body, so the probe failed, and the HTTPS fallback probe then failed autocert's host whitelist (`localhost` is never whitelisted). Containers were marked unhealthy in AutoTLS mode.

## Changes

- **install.sh**: the generated AutoTLS systemd unit now maps host `80->8080` and `443->8443`, matching the compose files. `load_existing_service_config` recognizes both the new `443:8443` and the legacy `443:443` forms, so an existing (broken) unit is detected as AutoTLS-enabled and regenerated with the corrected mapping instead of silently dropping AutoTLS on the next update.
- **Podman/quadlet/birdnet-go-autotls.container**: same `80:8080` / `443:8443` remap, and dropped the unnecessary `NET_BIND_SERVICE` capability.
- **internal/api/server.go**: a small `healthOrRedirect` handler serves the unauthenticated `/health` endpoint as JSON over the AutoTLS ACME listener and the manual-TLS HTTP redirect server, while everything else still redirects. ACME challenge paths are still intercepted first; no Dockerfile change is needed (the existing `:8080/health` probe now returns JSON).

## Testing

- `go test -race ./internal/api/...` (extended `TestAutoTLS_DualListeners` to assert `/health` returns 200 JSON over the plain-HTTP listener, plus a focused `TestServer_healthOrRedirect`).
- `scripts/install_test.sh` (added legacy-unit, localhost-bind-preservation, and a full `generate_systemd_service_content` round-trip that asserts the `80:8080` / `443:8443` mapping, no `NET_BIND_SERVICE`, and that the parser does not mistake the new `80:8080` line for the web port).
- `golangci-lint run`, `go vet`, shellcheck (no new findings), and the local quality gate.

## Related

Refs #3527, #3612.

## Breaking changes

None. Existing AutoTLS installs get a one-time systemd-unit rewrite on their next `install.sh` update that replaces the dead port mapping with the working one; this is the intended repair.

### DIFF
--- a/Podman/quadlet/birdnet-go-autotls.container
+++ b/Podman/quadlet/birdnet-go-autotls.container
@@ -9,8 +9,12 @@ ContainerName=birdnet-go-autotls
 Volume=./config:/config
 Volume=./data:/data
 Volume=./certs:/certs
-PublishPort=80:80
-PublishPort=443:443
+# AutoTLS binds the ACME HTTP-01 + HTTP->HTTPS redirect listener on 8080 and
+# HTTPS on 8443 inside the container, so map the privileged host ports to those
+# non-privileged container ports (no NET_BIND_SERVICE needed). Mirrors
+# Podman/podman-compose.autotls.yml.
+PublishPort=80:8080
+PublishPort=443:8443
 Environment=TZ=UTC
 Environment=BIRDNET_UID=1000
 Environment=BIRDNET_GID=1000
@@ -19,7 +23,6 @@ Environment=BIRDNET_SECURITY_HOST=%i
 Device=/dev/snd:/dev/snd
 Network=bridge
 Tmpfs=/config/hls:exec,size=50M,uid=1000,gid=1000,mode=0755
-AddCapability=NET_BIND_SERVICE
 
 [Service]
 Restart=always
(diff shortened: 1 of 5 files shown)


tphakala:
@coderabbitai review

Issue #3855, opened 2026-07-11
Modernize container distribution pipeline: buildx bake, xx cross-compilation, docker-init, and build attestations


tphakala:
## Summary

Modernize how the BirdNET-Go container images are built and shipped by adopting the
build tooling and packaging patterns that the upstream Docker engine project
([moby/moby](https://github.com/moby/moby)) uses for its own images. This is a
build/packaging/distribution change only. It does not touch application code.

The work is split into independent items (P1 highest value first). They can land as
separate PRs; none depend on the others except where noted.

## Background

I compared our container distribution against how moby builds and ships its own
images (its `Dockerfile`, `docker-bake.hcl`, and use of `tonistiigi/xx` and
`docker-init`/tini). moby is the Docker engine source, so there is nothing to
`import` as a Go library here. The value is in the build tooling it produces and the
Dockerfile patterns it demonstrates. Several of those directly address weak spots in
our current pipeline.

## Current state (verified against the tree)

- **Dev/nightly build** (`.github/workflows/docker-build.yml`): already does modern
  single-shot multi-arch via one `docker/build-push-action` with
  `platforms: linux/arm64, linux/amd64`. buildx emits the manifest list in one pass.
  Sets `provenance: false`.
- **Release build** (`.github/workflows/docker-release.yml`): does it the OLD manual
  way. A per-platform matrix (`linux/amd64`, `linux/arm64` at line ~15) builds two
  separate images tagged `-amd64` / `-arm64`, then a second job stitches them with
  `docker manifest create --amend` + `docker manifest push --purge`, wrapped in a
  `retry` helper (lines ~132-147). This manual manifest assembly is fragile and is
  exactly what buildx replaced.
- **Cross-compilation** (`Dockerfile`): the build stage is
  `FROM --platform=$BUILDPLATFORM golang:1.26-trixie` (line ~11), so we cross-compile
  CGo (TFLite / ONNX / ALSA) from an amd64 runner. It is hand-rolled: install
  `gcc-aarch64-linux-gnu` (line ~23) and repeat `case "${TARGETPLATFORM}"` arch
  switches for the ONNX runtime download (line ~87), OpenVINO (lines ~114-146), and
  the TFLite lib (line ~221).
- **PID 1 / init**: `ENTRYPOINT` is `entrypoint.sh` -> `startup-wrapper.sh`
  (`Dockerfile` lines ~319-320). `startup-wrapper.sh` is a bash script that runs the
[...]

PR #3880, opened 2026-07-12
ci(docker): publish container images via native multi-arch buildx (#3855 P2)

tphakala:
Part of #3855 (P2). Modernizes the container publish path so it stops hand-assembling the multi-arch manifest.

## What changed

`docker-build-push.yml` (the reusable `workflow_call` used by `nightly-build.yml` and `release-build.yml`) now does a **single multi-platform `build-push-action`** that emits the manifest list natively, the same pattern `docker-build.yml` already uses. Gone: the per-arch build matrix, the `-amd64`/`-arm64` intermediate tags, and the entire `create-manifests` job with its `docker manifest create --amend`/`push --purge` + `retry` dance. Net `-274/+52` lines.

## What is preserved (deliberately)

- **Every user-facing tag**, on GHCR **and** Docker Hub: `nightly`, `nightly-<DATE>`, `podman-nightly`, `podman-nightly-<DATE>`, `latest`, `podman-latest` for the nightly strategy; `<version>`, `podman-<version>`, `latest`, `podman-latest` for the release strategy. `install.sh` pins `:nightly`, so that tag keeps working unchanged.
- The `workflow_call` **input/output contract** (callers untouched).
- **Docker Hub login resilience**: Docker Hub tags are only added when its `continue-on-error` login actually succeeded, so a Docker Hub outage still lets GHCR publish.
- The **`validate-podman`** job (pulls and runs the published `:nightly` with Podman, rootless included) and the `cleanup` job, both unchanged.
- `provenance: false` unchanged; attestations are a separate evaluation (issue #3855 P4).

## What is dropped

The per-arch intermediate tags (`nightly-amd64-<DATE>`, `nightly-amd64`, the bare `:amd64`/`:arm64` release tags, and their `podman-` variants). These were only scaffolding for the manual manifest step; users pull the multi-arch tags. This is the intended modernization from the issue ("delete the per-platform matrix, the -amd64 / -arm64 suffix tags").

## Trade-off

The build now runs once (both platforms on one runner via buildx cross-compile) instead of two parallel per-arch runners, so the build job is longer wall-clock but simpler and drops the fragile manifest assembly. This matches how `docker-build.yml` already builds.

## Validation status (please read before merging)

Static: `actionlint` clean apart from pre-existing `info`/`style` shellcheck notes in the unchanged copied sections; structure and `needs` wiring verified; tag set diffed against the original. I could **not** run it end-to-end, because it only truly exercises when it pushes to GHCR/Docker Hub. It should get one `workflow_dispatch` validation run (confirm a multi-arch `:nightly` manifest lands on both registries and `docker buildx imagetools inspect` shows both platforms, and the `validate-podman` job stays green) before a release relies on it.

## Release notes
- audience: internal
- summary: Container images are now published with native multi-arch buildx instead of a hand-assembled manifest; same tags, more robust release/nightly publishing.
- feature: none
- breaking: none
- config: none
- schema: none

## Summary by CodeRabbit

* **New Features**
  * Docker images are now published as unified multi-architecture images under each tag.
  * Added consistent nightly, release, and Podman-compatible image tags.
  * Added optional `latest` tags when applicable.

* **Bug Fixes**
  * Podman validation now checks the exact nightly image produced by the current build.
  * Docker Hub publishing failures no longer prevent publishing to other configured registries.
  * Improved build status reporting and cleanup handling.

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -63,10 +63,6 @@ jobs:
       version: ${{ inputs.version }}
       ghcr-success: ${{ steps.push-status.outputs.ghcr-success }}
       dockerhub-success: ${{ steps.push-status.outputs.dockerhub-success }}
-    strategy:
-      fail-fast: false
-      matrix:
-        platform: ${{ fromJson(inputs.platforms) }}
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


tphakala:
Ran an independent (Gemini) review of this change. It confirmed the important parts: the user-facing tag set is byte-for-byte preserved (nightly, nightly-DATE, podman-nightly, latest and podman variants on GHCR + Docker Hub), the single multi-arch `build-push` is correct, no `setup-qemu-action` is needed (the ubuntu runners ship binfmt), and the multiline tags heredoc is consumed fine by build-push-action. It also noted this removes an implicit buildcache race the old amd64/arm64 matrix had (both jobs raced to overwrite the single `:buildcache` tag).

Two things from that review:

Fixed in bc1e417a3: `validate-podman` was pulling the floating `:nightly`, so a failed build could false-pass by validating the previous day's image. It now pulls the exact `nightly-<DATE>` tag this run built. Matters more since this becomes the release workflow.

Known trade-off (not changed): folding the per-registry `docker manifest push` retry loop into one `build-push` means a Docker Hub push flake after a successful Docker Hub login can now fail the whole job, GHCR included. In practice the old per-arch build step already pushed to both registries atomically, GHCR is the primary (install.sh pins ghcr.io), and a failed nightly just re-runs. If we want hard GHCR-independence back later, the clean way is to push GHCR from build-push and mirror to Docker Hub with `docker buildx imagetools create` + retry, rather than the old `docker manifest create`. Left out to keep this focused.

Pre-existing (left alone): the `Check push status` step has no `if: always()`, so it never runs on a build failure. That predates this PR; out of scope here.

Still needs one `workflow_dispatch` validation run before a release depends on it, since it only truly exercises against the registries.

PR #3900, opened 2026-07-13
feat(audio): make native go-flac the only FLAC encoder, drop FFmpeg FLAC path

tphakala:
## Summary

The native go-flac encoder, previously opt-in behind the experimental `BIRDNET_FLAC_ENCODER=native` environment variable, has proven reliable. This promotes it to the sole FLAC encoder for both the detection-save and BirdWeather soundscape-upload pipelines and removes the FFmpeg FLAC encoding code those paths used.

- Remove the `BIRDNET_FLAC_ENCODER` env gate (deleted `internal/audiocore/flac/gate.go`).
- `encodeClip`: FLAC now always encodes natively (with audionorm EBU R128 normalization + true-peak limiting when enabled); the FFmpeg branch only handles MP3/AAC/Opus.
- BirdWeather upload: always native go-flac + audionorm, in memory with finalized `STREAMINFO.total_samples`; deleted the FFmpeg encode helpers and the now-dead temp-file, gain, and parse helpers.
- `conf`: a `flac` export config is no longer downgraded to WAV when FFmpeg is missing (WAV and FLAC are both encoded natively).
- Updated the package README, `doc/wiki` install/guide/FAQ pages, and the shipped config template to match.

FFmpeg is still required for MP3/AAC/Opus export and for on-demand clip download transcoding (`ExtractClip`), which decodes stored files that go-flac (an encoder) cannot decode.

## Behavioral changes for existing users

- FLAC clips and BirdWeather uploads now normalize via native EBU R128 with true-peak limiting instead of FFmpeg's two-pass loudnorm. Integrated loudness targets the same value, but the algorithm is a single linear gain, so loudness may differ slightly, and the Normalization `LoudnessRange` (LRA) setting no longer affects FLAC (it still applies to the FFmpeg formats: MP3/AAC/Opus).
- On a host without FFmpeg, a `flac` export config now produces real `.flac` clips instead of being silently downgraded to `.wav` at startup. Heads-up for any script globbing `*.wav` in the clips directory on such a host.
- The undocumented `BIRDNET_FLAC_ENCODER` environment variable is removed and ignored; native is now unconditional.

## Test plan

- [x] `go build ./...` clean
- [x] `golangci-lint run` on the touched packages: 0 issues
- [x] `go test -race` green for `internal/audiocore/flac`, `internal/birdweather`, `internal/conf`, `internal/analysis/processor`
- [x] Added a test covering the default FLAC path (normalization disabled, static gain applied natively)

## Release notes

- audience: user
- summary: FLAC clips (saved detections and BirdWeather uploads) are now always encoded by the built-in Go encoder; FFmpeg is no longer used or required for FLAC.
- feature: native-flac-encoder
- breaking: none (the experimental `BIRDNET_FLAC_ENCODER` env var is removed/ignored; it was never a documented setting)
- config: none (FLAC export no longer needs FFmpeg; the Normalization LoudnessRange setting no longer affects FLAC output)
- schema: none

## Summary by CodeRabbit

* **New Features**
  * BirdWeather FLAC uploads now use native Go FLAC encoding (no FFmpeg dependency).
* **Bug Fixes**
  * FLAC normalization now uses native audionorm when in-range; if out of range, normalization is skipped (with a WARN) and encoding stays FFmpeg-free.
  * FFmpeg-missing export fallback now only rewrites formats that require FFmpeg (WAV remains WAV; FLAC stays FLAC).
* **Documentation**
  * Updated FFmpeg requirements for RTSP/HLS and audio export formats.
  * FAQ now includes guidance to resend Home Assistant MQTT discovery after updates.


tphakala:
Both CodeRabbit nitpicks are addressed in 808d356:

- Added `TestEncodeClipFLACNormalizationSkipped`, which drives the defensive branch in `encodeClip` (normalization enabled with an out-of-range true peak) and asserts it still returns `encoderNativeFLAC` and applies the static `Export.Gain`, so there is no FFmpeg fallback.
- Consolidated the four `applyFfmpegFormatFallback` tests into one table-driven test with parallel subtests.

@coderabbitai review

PR #3933, opened 2026-07-15
fix(classifier): auto-select ONNX range filter for default "latest" model

### LINKED ISSUE #3932: bug: After update the range filter is not working

2026-07-13T17:06:31Z

### Install method

### Device type

Raspberry Pi 4

Debian Linux

### Audio source

Network audio stream - RTSP / RTSPS (e.g. IP camera)

### Hardware details

Version: 20260713
Build date: 2026-07-13T17:06:31Z
Operating system: Debian Linux
Architecture: aarch64
Hardware: Raspberry Pi 4 Model B Rev 1.4
Environment: Docker

### Description

I updated my working birdnet go instance ( version nightly-20260429) to 20260713, and after the update I started getting detections for weird species. I haven't changed any settings so I expected the update to preserve the detection characteristics.

If I go into species settings there's this warning/error:
Image

And in the analysis settings there is this:
Image
It's showing over 6000 species (normally for my location there are around 100 species)

My location in general settings looks fine:
Image

### Steps to reproduce

Be on the old version mentioned above
Update the the new version mentioned above

### Expected behaviour

Detection behaviour (including range filters) should be preserved after the update.

### Actual behaviour

No range filters seem to be in place

### When did this start?

Worked before, started after I updated

### Relevant log output

```shell

```

_No response_

_No response_

### Pre-submission checks

### Additional context

_No response_

tphakala:
Fixes #3932.

## Problem

After updating, arm64 users (e.g. Raspberry Pi via Docker) started getting detections for out-of-range species: the active species list jumped from ~100 for their location to 6000+, i.e. the range filter stopped working entirely.

## Root cause

The config default is `birdnet.rangefilter.model: "latest"`, persisted in essentially every existing `config.yaml`. The two range-filter auto-selection gates in `initializeMetaModel` only fired when the model string was empty (`rf.Model == ""`), so the `"latest"` default triggered neither and fell through to the TFLite range-filter backend.

On ONNX-only arm64 container images this is fatal: those images ship `BirdNET_MData_V2.onnx` (the ONNX range-filter model) but no `.tflite` range-filter file, so the TFLite path cannot find its model, fails to load, and the instance runs with no range filter. amd64 images ship the `.tflite` files, so they were unaffected. This matches the reports (Raspberry Pi 4 / aarch64).

## Fix

Treat the `"latest"` sentinel the same as `""` for auto-selection, so it means "pick the best available range filter": the v3 geomodel when its files are present, then the shipped ONNX MData model on arm64, then the embedded/shipped TFLite MData. For affected arm64 installs this restores the exact v2.4 range filter with no download and no change to detection characteristics.

Both auto-select gates now honor an explicit `rangefilter.modelpath`: the v3 geomodel gate gains the same ModelPath guard the arm64 gate already had, so a user-provided range-filter path is never overridden by the stock geomodel.

## Behavior change (release note)

arm64 users who have been running unfiltered since `"latest"` became the default will get range filtering back, so out-of-range species that were showing up get suppressed again. This is the intended restoration of pre-regression behavior, but it visibly changes detection output for anyone who got used to the unfiltered results, so it is worth a line in the release notes.

## Tests

- `TestIsAutoSelectRangeFilterModel`, `TestShouldSelectDefaultONNXRangeFilter` cover the arm64 ONNX MData selection matrix.
- `TestShouldAutoSelectV3GeomodelForConfig` covers the v3 geomodel gate, including the explicit-modelpath guard.
- Full `internal/classifier` suite passes with `-race`; `golangci-lint` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic range-filter and geomodel selection consistency.
  * Respects explicitly provided range-filter model paths without overriding them.
  * Enhanced default ONNX range-filter selection for compatible ARM64 setups, with better safeguards when files are missing or configurations are incompatible.
  * Standardized accepted range-filter model names during configuration/environment validation.
* **Tests**
  * Added unit tests covering v3 geomodel auto-selection gating and default ONNX range-filter selection behavior (including explicit vs auto modes).

### DIFF
--- a/internal/classifier/birdnet.go
+++ b/internal/classifier/birdnet.go
@@ -531,8 +531,10 @@ func (bn *BirdNET) initializeMetaModel(settings *conf.Settings) error {
 
 	// Auto-select v3 geomodel for compatible classifiers when files exist on disk.
 	// Only applies locally for routing; does NOT publish settings to avoid
-	// inconsistency if the backend fails to initialize.
-	if rf.Model == "" && bn.modelsDir != "" && shouldAutoSelectV3Geomodel(bn.ModelInfo.ID, bn.modelsDir) {
+	// inconsistency if the backend fails to initialize. Skipped when an explicit
+	// rangefilter.modelpath is set, so a user-provided range-filter path is never
+	// overridden by the stock geomodel (mirrors the arm64 default gate below).
+	if shouldAutoSelectV3GeomodelForConfig(rf.Model, rf.ModelPath, bn.ModelInfo.ID, bn.modelsDir) {
 		localSettings := conf.CloneSettings(settings)
 		applyAutoSelectedGeomodelPaths(localSettings, bn.modelsDir)
 		settings = localSettings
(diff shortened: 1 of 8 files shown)


tphakala:
Addressed the nitpick: centralized the `latest`/`legacy`/`v3` range-filter sentinels into exported `RangeFilterModel{Latest,Legacy,V3}` constants in the `conf` package (the shared home, since `classifier` imports `conf`) and threaded them through the default, validation, env parsing, the auto-select helper, and the tests. Value-identical, no behavior change.

@coderabbitai review

PR #3970, opened 2026-07-19
fix(frontend): stop the hour axis stretching past the last hour

Kerakis:
## Summary

Every hour-based chart on the Analytics **Patterns** tab stopped drawing before the right edge, leaving a dead ~1 hour gap — and the axis label at that edge was in the wrong place. Reported during QA: the tooltip showed real 23:00 values while the series appeared to end around 22:30.

## Root cause

`createLinearScale` applied d3's `nice()` **unconditionally**. `nice()` rounds a domain outward, which is right for a value axis but wrong for an hour-of-day axis whose domain is already exact:

```js
scaleLinear().domain([0, 23]).nice()   // -> domain [0, 24]
scale(23)                              // -> 958 of 1000  (~96%)
scale.ticks(12)                        // -> 0,2,…,22,24
```

Two consequences on each affected chart:

1. **Dead space.** Hour 23 renders at ~96% of the width, so every series stops short of the plot edge.
2. **The axis mislabels itself.** `nice()` adds a tick at hour **24**, and `createHourAxisFormatter` clamps values to 23 (`Math.min(23, …)`), so that tick renders as **"23:00"** — at a position where the 23:00 data is *not*. The real 23:00 point sits to its left, which is why hovering 23:00 showed values while the line looked finished.

Affected (all three build the same hour scale): `TimeOfDaySpeciesChart` (`[0, 23]`), `AcousticSuccessionChart` (`[0, SUCCESSION_HOURS - 1]`), `SpeciesRidgeline` (`[0, xLen - 1]`).

## Changes

- **`utils/scales.ts`** — `LinearScaleConfig` gains an optional `nice` flag, **defaulting to `true`** so every existing value axis is byte-for-byte unchanged. Only the hour axes opt out.
- **`utils/axes.ts`** — new `hourAxisTickValues(lastHour, step)`. d3's own ticks stop at the last nice multiple (22:00), leaving the final hour unlabeled; this always includes `lastHour` so the axis ends with a correctly positioned label at the edge. A regular tick no more than half a step from the final hour is dropped so the two can't collide (with a 2-hour step that removes 22:00).
- **The three charts** — pass `nice: false` and take their ticks from the shared helper, replacing two hand-rolled `Array.from(...)` tick expressions.

## Tests

- `scales.test.ts` (new): documents that the default still rounds `[0,23]` → `[0,24]`, and that `nice: false` keeps the domain exact with `scale(23)` landing on the plot edge. Also covers that domain validation still runs when `nice` is disabled.
- `axes.test.ts`: `hourAxisTickValues` always ends on the final hour, drops a colliding neighbour, and never emits a tick beyond it (checked across several steps).
- 202 chart tests pass; `npm run check:all` clean (svelte-check: 0 errors, 0 warnings).

## Notes

Pre-existing — it predates the current time-of-day work and is unrelated to it; it only became obvious once there was enough data in the final hour to notice. Kept separate from #3969 for that reason.

## Summary by CodeRabbit

* **Improvements**
  * Updated analytics charts to preserve exact hour ranges and display the final hour tick, including 23:00.
  * Improved tick spacing to reduce label collisions and ensure consistent hour-axis formatting across charts.

* **Tests**
  * Added coverage for hour-axis tick generation and exact linear-scale domain behavior.
  * Verified validation remains intact for invalid scale domains.

### DIFF
--- a/frontend/src/lib/desktop/features/analytics/components/charts/d3/AcousticSuccessionChart.svelte
+++ b/frontend/src/lib/desktop/features/analytics/components/charts/d3/AcousticSuccessionChart.svelte
@@ -30,7 +30,13 @@
   import { localizeSpeciesName } from '$lib/utils/speciesDisplay';
   import BaseChart from './BaseChart.svelte';
   import { createLinearScale } from './utils/scales';
-  import { createAxis, styleAxis, addAxisLabel, createHourAxisFormatter } from './utils/axes';
+  import {
+    createAxis,
+    styleAxis,
+    addAxisLabel,
+    createHourAxisFormatter,
+    hourAxisTickValues,
+  } from './utils/axes';
   import { ChartTooltip } from './utils/interactions';
   import type { ChartTheme } from './utils/theme';
   import { getSpeciesColor, registerChart } from './utils/speciesColor';
@@ -192,7 +198,13 @@
     }
     if (!Number.isFinite(yMin) || !Number.isFinite(yMax) || yMax - yMin < MIN_Y_SPAN) return;
 
-    const xScale = createLinearScale({ domain: [0, SUCCESSION_HOURS - 1], range: [0, innerWidth] });
+    // nice: false - the hour domain is exact; rounding it outward would leave the final hour short
+    // of the right edge and open a gap the bands never reach.
+    const xScale = createLinearScale({
+      domain: [0, SUCCESSION_HOURS - 1],
+      range: [0, innerWidth],
+      nice: false,
+    });
     const yScale = scaleLinear().domain([yMin, yMax]).range([innerHeight, 0]);
 
     // X axis: sampled hour ticks along the bottom. No y-axis (the wiggle baseline is meaningless).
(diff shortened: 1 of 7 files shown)


tphakala:
Reviewed and verified locally.

The root cause holds up. `nice()` on an already-exact hour domain is the bug, and the second-order effect you documented is the nastier half: `nice()` adds a tick at hour 24, `createHourAxisFormatter` clamps with `Math.max(0, Math.min(23, ...))`, so hour 24 renders as a "23:00" label sitting to the right of where the actual 23:00 data is. That fully explains the QA report of a tooltip showing 23:00 values on a series that looked finished.

Defaulting `nice` to `true` keeps all 12 other `createLinearScale` call sites byte-identical, which is the right call.

I walked `hourAxisTickValues` through the callers rather than trusting the tests:

- `TimeOfDaySpeciesChart` uses the defaults (23, 3), so 13-ish auto ticks become 9. Strictly less crowded than the old `tickCount: 12`, not more.
- `AcousticSuccessionChart`: `SUCCESSION_HOURS` is 24 and `X_TICK_STEP` is 3, so `(23, 3)`. Adds the 23 label the old `Array.from` expression dropped.
- `SpeciesRidgeline`: `xLen = Math.max(1, ...)` and `step = Math.max(1, xTickStep)`, so `lastHour >= 0` and `step >= 1`. The `xLen === 1` case degenerates to `[0]` rather than misbehaving, and the loop cannot fail to advance. Since the tick label there comes from a caller-supplied `xTickFormat` over bin indices, forcing the final index stays correct for non-hourly binnings too.

The collision rule checks out at every step I tried: step 2 drops 22 (1 hour away, exactly half a step, so `>` excludes it), step 3 keeps 21 (2 > 1.5), step 4 keeps 20 (3 > 2).

202 chart tests pass and `npm run check:all` is clean (svelte-check 0/0). Merging.

Issue #3974, opened 2026-07-19
Custom species config badge is missing when config uses an alias


ejhnsn:
## Problem

The Active Species table determines the Configured badge by literally comparing each displayed common/scientific name with the keys in realtime.species.config.

When a config key uses a supported common-name or taxonomic alias, the range-filter pipeline can resolve and force-include the species under its canonical display name, but the frontend comparison no longer matches. The species is configured functionally, yet the Active table does not show the Configured badge.

Example:

1. Add a custom species config keyed by Red-crowned Parrot.
2. The backend resolves the species as Amazona viridigenalis and the Active table displays Red-crowned Amazon.
3. The row is active, but the Configured badge is absent because neither displayed name equals the original config key.

## Relevant code

- frontend/src/lib/desktop/features/settings/pages/SpeciesSettingsPage.svelte: frontend mapping of hasCustomConfig
- internal/classifier/range_filter.go: alias resolution for config-key overrides

## Suggested direction

Carry custom-config provenance through backend alias resolution as a field distinct from Always Include provenance, then consume that field in the Active table. Retain literal frontend matching as a compatibility fallback for older backends.

Add coverage for a config-only alias whose canonical common name differs from the configured key.

Issue #3973, opened 2026-07-19
Species rarity lookup does not use the active geomodel taxonomy


ejhnsn:
## Problem

Species rarity coverage and score matching use the primary classifier labels and raw scientific-name equality rather than the active geomodel taxonomy.

speciesHasGeomodelCoverage checks bn.Labels(), which is the primary classifier vocabulary. With the universal 12K geomodel, a species can have geomodel coverage while being absent from that primary vocabulary. If it is below today's threshold, the API reports Unknown instead of Very Rare.

The score lookup also compares extracted scientific names with EqualFold only. Supported legacy/current taxonomic aliases can therefore miss an available native geomodel score and fall into the wrong rarity result.

## Relevant code

- internal/api/v2/species/species.go: speciesHasGeomodelCoverage
- internal/api/v2/species/species.go: getSpeciesRarityInfo score lookup
- internal/classifier/range_filter.go: universal geomodel label and taxonomy mapping

## Suggested direction

Expose the active geomodel label set or a geomodel-coverage query through the classifier/orchestrator boundary, and use the same canonical taxonomic identity used by range-filter mapping for both coverage and native-score matching.

Add tests for a universal-geomodel-only species and for a legacy/current scientific-name alias pair.

PR #3976, opened 2026-07-19
fix(range): preserve native scores for included species

### LINKED ISSUE #3975: Occurrence lookup treats force-included score as geomodel probability
## Problem

Species occurrence lookup can treat a force-included override's synthetic score of 1.0 as a native geomodel probability.

The probable-species pipeline appends Always Include and custom-config overrides at score 1.0. internal/classifier/birdnet.go currently stores every returned score in getCachedSpeciesScores, and the fallback scan in GetSpeciesOccurrenceAtTime also returns the first matching row. For an overridden species, detections and integrations such as MQTT can therefore report occurrence 1.0 even when the native geomodel score is lower or unavailable.

## Relevant code

- internal/classifier/birdnet.go: getCachedSpeciesScores
- internal/classifier/birdnet.go: GetSpeciesOccurrenceAtTime fallback scan
- internal/classifier/range_filter.go: user override score insertion

## Suggested direction

Distinguish synthetic override rows from native geomodel rows and exclude the synthetic rows from both the occurrence cache and fallback lookup. If native and override labels differ by a supported alias, match on the canonical taxonomic identity rather than returning the synthetic row.

Add regression tests for both the cache and fallback paths with a force-included species whose native probability is below 1.0 or absent.

ejhnsn:
## Summary

- build on #3980's pointer-based Included/Configured provenance instead of duplicating its resolver or wire fields
- distinguish appended synthetic `score: 1.0` override rows from native geomodel probabilities
- preserve the compatibility `score` value while exposing the colliding native probability through the optional `rangeScore` field
- keep synthetic override sentinels out of raw geomodel scores, species rarity, and detection occurrence calculations
- display and sort Active Species by the native range score when available while preserving existing CSV values and ordering

Closes #3975.

## Problem

An Always Include or custom-config entry can resolve to the same species under a different model/common-name label. The classifier then legitimately contains both a synthetic always-active row at `1.0` and a native geomodel row with the real probability.

For example, BirdNET may label `Amazona viridigenalis` as `Red-crowned Parrot` while the geomodel/OpenFauna display name is `Red-crowned Amazon`. Display deduplication previously kept the `1.0` sentinel and discarded the native probability. The same sentinel could also leak into raw-score, rarity, and detection occurrence consumers.

This change tags only appended override rows as synthetic. At the display boundary, the synthetic row remains the compatibility survivor while `rangeScore` carries the highest colliding native probability. Consumers that promise native occurrence values ignore synthetic rows.

## Compatibility

- No existing JSON field is removed, renamed, or retyped.
- `rangeScore` is additive and omitted when no native/synthetic collision exists.
- #3980's `hasCustomConfig` and `isManuallyIncluded` pointer/`omitempty` contract is retained.
- Existing API and CSV `score: 1.0` override behavior, CSV columns, and CSV ordering are preserved.
- The internal synthetic marker is never serialized.

## Testing

- [x] `golangci-lint run -v ./...`: clean
- [x] `go test -race ./...`: passed
- [x] `npm run check:all`: formatting, ESLint, Stylelint, Svelte typecheck, and AST rules passed with 0 diagnostics
- [x] `npm test`: 181 files and 2,681 tests passed
- [x] Focused alias, deduplication, DTO, CSV, raw-score, rarity, and occurrence regressions passed under the race detector

## Preflight Status

- [x] Single concern: one native-vs-synthetic range-score bug fix
- [x] Six review passes completed: reuse, correctness/safety, quality, i18n, integration wiring, and regression/backward compatibility
- [x] Preflight finding fixed: detection occurrence indexing now excludes synthetic override sentinels, with alias regression coverage
- [x] Post-fix re-review found no remaining changed-line or maintainer-decision items
- [x] Diff contains only files relevant to the fix; no core TODO/FIXME, secrets, credentials, or PII
- [x] No breaking config, database, CLI, route, request, or response changes

Secondary-model cross-validation was unavailable; the six primary review passes and all required local gates completed successfully.

An unrelated pre-existing i18n issue found during preflight is tracked separately in #4041.

## Summary by CodeRabbit

- **New Features**
  - Active Species now separates displayed range scores from CSV export scores.
  - CSV exports use deterministic ordering while preserving API-compatible score values.
  - Improved handling of manually included species, aliases, and localized labels.

- **Bug Fixes**
  - Synthetic override entries no longer affect native rarity or occurrence scoring.
  - Active Species deduplication now preserves native scores and inclusion state.

- **Tests**
  - Added coverage for score handling, CSV formatting, JSON compatibility, and override filtering.

## Release note

Occurrence and rarity for force-included species (Always Include or custom-config entries) now use the native geomodel probability instead of the synthetic 1.0 inclusion score. Detection occurrence values, including those published over MQTT, and the rarity badge for such species can change to their true geomodel-based values (or unknown / very rare when the geomodel has no data for them). The API and CSV `score` fields keep the existing 1.0 override value for compatibility, and the native probability is additionally exposed through the new optional `rangeScore` field, which the Active Species table uses for display.


tphakala:
Heads up, and apologies for the extra work: I merged #3980 (the fix for #3974, the custom-config badge case you filed alongside this one) and it landed in the same files as this PR. I should have checked the open PR queue before merging and did not, so this needs a rebase that it would not otherwise have needed.

Concretely, this branch had one conflict against main before that merge (`internal/api/v2/species/species.go`, from unrelated drift). It now has five:

```
frontend/src/lib/desktop/features/settings/pages/SpeciesSettingsPage.svelte
internal/api/v2/dto/range.go
internal/api/v2/range/range.go
internal/api/v2/species/species.go
internal/classifier/range_filter.go
```

The one that needs a decision rather than a mechanical resolve is `isManuallyIncluded`. Both PRs add that exact JSON field to `dto.RangeFilterSpecies`, with different Go types:

- this PR: `IsManuallyIncluded bool` with `omitempty`
- what merged: `IsManuallyIncluded *bool` with `omitempty`, alongside `HasCustomConfig *bool`

The pointer version is deliberate. With a plain `bool` plus `omitempty`, a `false` is omitted from the payload too, so the client cannot tell "this species has no manual inclusion" from "this backend does not send the field at all". The merged frontend depends on that distinction: it does `s.isManuallyIncluded ?? isInNameSet(includeSet, ...)`, falling back to the old literal name matching only when the field is genuinely absent, which is what keeps an older backend working. A non-pointer bool would collapse that fallback into always-on.

So my suggestion is to drop `IsManuallyIncluded` from this PR entirely and consume the merged one. The provenance plumbing through `addUserOverrideSpeciesScores` is already on main, including the alias resolution and the flag union in `dedupeSpeciesForDisplay`, with tests. That leaves this PR's genuinely distinct contribution intact and unduplicated: `rangeScore`, exposing the native geomodel probability separately from the synthetic `1.0` override sentinel, plus keeping synthetic rows out of the raw-score and rarity consumers. None of that is covered by #3980 and it is the more interesting half of this change.

Worth a look at `internal/classifier/range_filter.go` on main before you rebase. `resolveOverrideLabelsWithSource` now returns per-label provenance from a single resolution pass, and `resolveOverrideLabels` is a thin wrapper over it, so if this PR was threading include-provenance through separately there is probably an existing seam to hook into rather than a parallel path to maintain.

Happy to do the rebase myself if you would rather not untangle it, just say so. And thanks for the three well-scoped issues you split out of this work, #3973/#3974/#3975 were unusually easy to act on.

ejhnsn:
Rebased onto current `main` and force-pushed. I resolved the #3980 conflicts as requested:

- retained #3980's `*bool` + `omitempty` provenance fields and `resolveOverrideLabelsWithSource`
- removed the parallel resolver/plain-boolean implementation from this branch
- kept this PR focused on the distinct native-vs-synthetic score behavior: internal synthetic tagging, additive `rangeScore`, and excluding sentinel rows from native score consumers
- extended the native-score filtering to detection occurrence indexing as well as the raw-score and rarity paths (closes #3975)

The PR is mergeable again. Full local gates pass: `golangci-lint`, `go test -race ./...`, `npm run check:all`, and 2,681 frontend tests. The description and preflight certification have been refreshed for the rebased scope.

tphakala:
Hi @ejhnsn, a quick licensing note before this gets merged.

tphakala:

ejhnsn:
@tphakala consented [here](https://github.com/tphakala/birdnet-go/pull/4091#issuecomment-5621656890)

ejhnsn:
Addressed CodeRabbit's nitpick in 82b2a3697 by naming the synthetic override compatibility sentinel as `syntheticOverrideScore`. Full preflight, lint, Go race tests, and frontend checks/tests pass.

PR #3979, opened 2026-07-19
fix(audio): normalize quiet clips to target on the native export path

tphakala:
Native loudness normalization was leaving two classes of quiet clip short of the configured target. Both are user-visible on FLAC export, which is unconditionally native, and both are fixed here so the FFmpeg `loudnorm` path can be removed in a follow-up without changing what users hear.

## Sub-gate clips were exported inaudible

A clip whose gated integrated loudness falls below EBU R128's -70 LUFS absolute gate measures as `-Inf`, so no gain was planned at all and the clip was written out untouched. The FFmpeg path has never had this hole: `loudnormGateFallbackOffset` derives an offset from the true peak instead of giving up. `gateFallbackGainDB` ports that idea.

The true-peak anchor alone is not enough. It leaves the output loudness at `ceiling - crestFactor`, which drives a flat signal far past the target: measured, a steady noise floor (a dead or muted microphone, crest factor of a few dB) landed at **-13.8 LUFS against a -23 target**. That would be loud hiss where users previously got a silent clip. Gating failing is itself information, so a second bound caps the lift at `target - gate`, which cannot overshoot however flat the clip turns out to be.

That bound has to assume the worst case, so alone it left clips well short of where FFmpeg put them. The lift usually raises a clip above the gate though, at which point it *can* be measured, so `refineLiftedGainDB` re-measures the lifted signal and lets a normal plan finish the job. A clip still under the gate afterwards keeps the conservative bound rather than guessing.

## The 30 dB gain ceiling was the dominant constraint

It bound 7 of 20 measured cases, leaving exports up to 8.6 LU below target where FFmpeg reached it. The new `nativeExportMaxGainDB` is derived rather than picked: the widest lift any in-range config can ask for is `conf.MaxTargetLUFS` (-10) minus the absolute gate (-70), so 60 dB. It also happens to match `ffmpeg.MaxGainDB`.

BirdWeather soundscape uploads deliberately keep the 30 dB ceiling. Those go to a public platform and are not this path's to change.

## Verification

A new comparison harness (`internal/audiocore/normbench`, build tag `normcompare`, never runs in CI) normalizes real recordings both ways and measures both outputs with the same analyser:

```
go test -tags normcompare -v ./internal/audiocore/normbench/
```

After this change the native path tracks FFmpeg **within 0.05 LU on 16 of 20 cases**, and the 6 previously untouched sub-gate clips are normalized. The residual gap is true-peak-limited clips, where FFmpeg silently abandons `linear=true` and applies dynamic range compression while the native path stays linear. That is the intended behaviour for nature recordings, which is also why `Normalization.LoudnessRange` stays unconsumed: on 15 s clips `loudnorm`'s dynamic mode barely moves LRA (7.4 to 7.1 on a purpose-built high-LRA clip) and misses the loudness target by 1.1 LU where the linear gain lands within 0.02.

Also verified: full-project `golangci-lint` under the CI build tags, `go test -race` across the audio packages, and the existing `TestEncodeClip_Ultrasonic*` regressions at 96/192/256/384 kHz.

## Release note

Quiet clips now reach the configured loudness target on FLAC export. Clips that were previously exported silent because they fell below the R128 measurement gate are now audible. MP3, AAC and Opus are unchanged.

### DIFF
--- a/internal/analysis/processor/actions_database.go
+++ b/internal/analysis/processor/actions_database.go
@@ -6,6 +6,7 @@ package processor
 import (
 	"context"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -20,6 +21,7 @@ import (
 	"github.com/tphakala/birdnet-go/internal/audiocore/ffmpeg"
 	"github.com/tphakala/birdnet-go/internal/audiocore/flac"
 	"github.com/tphakala/birdnet-go/internal/audiocore/opus"
+	"github.com/tphakala/birdnet-go/internal/audiocore/pcmgain"
 	"github.com/tphakala/birdnet-go/internal/audiocore/resample"
 	"github.com/tphakala/birdnet-go/internal/conf"
 	"github.com/tphakala/birdnet-go/internal/datastore"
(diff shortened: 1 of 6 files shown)


tphakala:
@coderabbitai review

PR #3915, opened 2026-07-14
feat(analytics): honor the species selection on all patterns charts

Kerakis:
## Summary

On the Analytics **Patterns** tab, the charts didn't reflect the control bar's species selection: the ridgeline (Who Sings When) and acoustic-succession streamgraph always rendered a top-N-by-volume set, and the heatmap/dawn-onset aggregated all species. This PR makes **every patterns chart honor the species selection at all times**:

- **≥1 species selected** → the chart shows exactly those species (still volume-ordered).
- **0 species selected** → a *"Select species and a date range"* empty state, matching the existing Time-of-Day chart. No more top-N / all-species default.

It also fixes a bug where a **multi-label species silently dropped selected species** from the ridgeline/succession ("10 selected, 7 drawn").

## Behavior

| Selection | Ridgeline / Succession / Heatmap / Dawn-onset |
|---|---|
| ≥1 species | exactly the selected species |
| 0 species | empty state prompting a selection |

Selecting *N* species now draws all *N* — including low-volume ones — regardless of how many model labels each species owns.

## Key changes

**Datastore / ranking** (`repository/detection_impl.go`, `v2only/datastore.go`)
- `GetTopSpecies` takes an optional `species []string` filter — a **parameterized** `WHERE labels.scientific_name IN (?)` applied *before* `ORDER BY count DESC / LIMIT` — and excludes false positives (`LEFT JOIN reviews … verified != false_positive`) so its ranking matches the summary/hourly/confidence queries.
- **Multi-label truncation fix:** a species can own several model labels, and `GetTopSpecies` groups/limits by label **row**, so a high-volume species' extra rows exhausted `LIMIT = selection.length` before the lowest-volume selected species was reached — dropping it *before* the rows were merged back into one series per species. `GetTopSpecies` now treats `limit <= 0` as no-limit, and `selectTopSpeciesHourly` passes `0` when a selection is present (the `IN` filter already bounds the set).
- `GetHourlyDistributionBySpecies` / `GetAcousticSuccession` thread the optional filter.

**API (v2)** (`analytics.go`)
- Both time endpoints accept an optional, repeatable `?species` scientific-name filter (trimmed, empty-filtered, de-duplicated), documented in the endpoint README. No new routes.

**Frontend** (`registry/charts.ts`, `en.json`)
[...]

PR #3981, opened 2026-07-19
fix(analytics): exclude false positives from new species detections

Kerakis:
## Problem

The **New Species Detected** chart on the Analytics Summary page lists species whose detections were all reviewed and flagged as false positives. Reported twice during Raspberry Pi QA — on that station, the only two "new" species shown were both ones the owner had marked incorrect.

## Root cause

`detectionRepository.GetNewSpecies` computes each species' lifetime first detection with a derived table over `detections JOIN labels`, with no join to the reviews table and no false-positive filter — at either level of the query. Every neighbouring analytics query (`GetSpeciesFirstSeenInPeriod`, `GetSpeciesPhenologyInPeriod`, `GetHourlyDistribution`, …) applies the filter via `buildAnalyticsBaseQuery`; this one was missed.

Two visible consequences:

1. A species whose detections were **all** reviewed away is still reported as newly detected.
2. A false-positive detection can **pin** a species' first-seen date earlier than its first real detection.

The legacy `internal/datastore/analytics.go` implementation already excludes false positives in both of its CTEs, so only the v2 datastore path is affected.

Same class of bug as #3915 fixed for `GetTopSpecies`.

## Fix

Apply the established exclusion pattern at both levels of the query:

- the **derived lifetime-first table**, so a reviewed-away detection cannot establish or move a species' first-seen;
- the **outer detail join**, so the reported `detection_id`/`confidence` never come from a false-positive row sharing that timestamp.

`DetectionReview` has a unique index on `detection_id`, so neither `LEFT JOIN` multiplies rows — the aggregates are unaffected.

## Tests

New `internal/datastore/v2/repository/new_species_test.go`, following the existing `TestGetSpeciesFirstSeenInPeriod_ExcludesFalsePositives` precedent:

- `TestGetNewSpecies_ExcludesFalsePositives` — an all-false-positive species is absent; a species whose earliest detection is a false positive reports its first *real* detection as first-seen. Fails on `main` with exactly the reported symptom.
- `TestGetNewSpecies_FalsePositiveDoesNotHideEarlierLifetimeFirst` — a species first detected before the window is still not reported as new (guards the window bound while changing the query).

## Verification

- `go test ./internal/... -race -count=1` — datastore packages green. One unrelated pre-existing failure (`TestAudioExportPartialUpdate` in `internal/api/v2`), confirmed failing identically without this change.
- `golangci-lint run --timeout 10m` — 0 issues.
- No frontend files touched; no interface change, so no mock regeneration.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved “new species” results by excluding detections marked as false positives.
  - Updated first-seen and lifetime representative calculations to ensure false-positive detections don’t impact whether a species is considered newly detected.
  - Prevented species with only false-positive detections from appearing in new species results.
- **Tests**
  - Added/extended coverage to verify false-positive exclusion behavior both inside and outside the query window.

### DIFF
--- a/internal/datastore/v2/repository/detection_impl.go
+++ b/internal/datastore/v2/repository/detection_impl.go
@@ -1751,7 +1751,7 @@ func (r *detectionRepository) GetNewSpecies(ctx context.Context, start, end int6
 		JOIN %s d ON d.label_id = l.id AND d.detected_at = species_first.lifetime_first
 		LEFT JOIN %s dr ON dr.detection_id = d.id
 		WHERE (dr.verified IS NULL OR dr.verified != ?)
-		GROUP BY species_first.scientific_name, species_first.lifetime_first
+		GROUP BY species_first.scientific_name, species_first.lifetime_first, species_first.lifetime_last
 		ORDER BY first_detected DESC
 		LIMIT ? OFFSET ?
 	`, r.tableName(), r.labelsTable(), r.reviewsTable(), r.labelsTable(), r.tableName(), r.reviewsTable())
--- a/internal/datastore/v2/repository/detection_impl.go
+++ b/internal/datastore/v2/repository/detection_impl.go
@@ -1708,7 +1708,8 @@ func (r *detectionRepository) GetDetectionTrends(ctx context.Context, period str
 	return r.GetDailyAnalytics(ctx, startTime, now, tzOffsetSeconds, nil, modelID)
 }
 
-// GetNewSpecies returns species detected for the first time ever within the range.
+// GetNewSpecies returns species detected for the first time ever within the range, false positives
+// excluded, so a species whose qualifying detections were all reviewed away is not reported as new.
 // Groups by scientific_name to aggregate across all models for the same species.
 // Uses MIN(id) as tie-breaker to avoid duplicates when multiple detections share the same timestamp.
 func (r *detectionRepository) GetNewSpecies(ctx context.Context, start, end int64, limit, offset int) ([]NewSpeciesData, error) {
@@ -1720,6 +1721,12 @@ func (r *detectionRepository) GetNewSpecies(ctx context.Context, start, end int6
 	// Approach: Use a derived table to compute lifetime first detection per species,
 	// then filter and join back to get detection details. This is O(n) instead of
 	// the O(n²) correlated subquery approach.
+	// Both levels exclude false positives (same filter as buildAnalyticsBaseQuery): the derived table
+	// so a reviewed-away detection cannot pin a species' lifetime first-seen (or make an
+	// all-false-positive species look new at all), and the outer join so the reported
+	// detection_id/confidence never come from a false-positive row sharing that timestamp.
+	// DetectionReview has a unique index on detection_id, so neither LEFT JOIN multiplies rows.
+	fpFilter := string(entities.VerificationFalsePositive)
 	rawSQL := fmt.Sprintf(`
 		SELECT
 			MIN(d.label_id) as label_id,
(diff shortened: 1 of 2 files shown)


Kerakis:
Thanks — addressed in 4cc3aff78.

**1. Seed a false positive in `TestGetNewSpecies_FalsePositiveDoesNotHideEarlierLifetimeFirst` — valid, fixed differently.**

You were right that the name promised false-positive involvement the body didn't have. I went further than the suggested diff, because marking the in-window detection at `1000` as a false positive would still not have exercised the fix: with a real detection at `100`, `lifetime_first` is `100` either way, so the species is excluded before and after the change. That test would pass against the buggy query.

Replaced it with the case that does discriminate — a false positive *predating* the window:

```go
fpID := seedDetection(t, db, label, 100, 0.6) // false positive before the window
seedFalsePositiveReview(t, db, fpID)
seedDetection(t, db, label, 1000, 0.8)        // first real detection, inside the window
```

The old query treated the false positive as prior history and dropped the species; the new one reports it as new with `FirstDetected == 1000`. This is the inverse failure of the reported bug — it silently *hid* genuinely new species — and it was untested. Verified failing against the pre-fix query and passing after.

The window-bound assertion is kept as its own test (`TestGetNewSpecies_RealDetectionBeforeWindowIsNotNew`), so both properties are covered.

**2. `GROUP BY species_first.lifetime_last` — applied.**

`lifetime_last` is functionally dependent on `scientific_name` within the derived table, so the grouping is unchanged and this is a no-op on results. Naming it explicitly is free and avoids depending on MySQL's functional-dependency detection under `ONLY_FULL_GROUP_BY`.

**3. `t.Parallel()` — declining.**

`setupInsightsTestDB` opens `file::memory:?cache=shared`, a process-wide shared SQLite database with `SetMaxOpenConns(1)`. Tests using it are not independent — they share one database and seed overlapping species, so running them in parallel would cross-contaminate. The existing sibling suites (`species_first_seen_test.go`, `species_phenology_test.go`) don't parallelize for the same reason, and `internal/CLAUDE.md` scopes `t.Parallel()` to tests that don't share mutable state.

Gates: `go test ./internal/datastore/... -race -count=1` green, `golangci-lint run --timeout 10m` 0 issues. Also verified on real hardware — the previously-misreported species no longer appear on the chart.

tphakala:
Reviewed and verified locally. Real bug, correct fix.

I confirmed the bug both ways by running your new tests against a clean `main` worktree:

```
--- FAIL: TestGetNewSpecies_ExcludesFalsePositives
    Error: "[{2 Turdus merula ...} {1 Parus major 1000 ...}]" should have 1 item(s), but has 2
--- FAIL: TestGetNewSpecies_FalsePositiveBeforeWindowDoesNotHideNewSpecies
    Error: "[]" should have 1 item(s), but has 0
```

The second one is the more interesting half: on `main` a false positive predating the window also *hid* genuinely new species. Good catch documenting that direction too.

The filter matches `buildAnalyticsBaseQuery` exactly, and I verified the no-row-multiplication claim against the entity: `DetectionReview.DetectionID` is `gorm:"not null;uniqueIndex"` and `Verified` is `not null`, so the `LEFT JOIN` yields NULL only for unreviewed detections. Adding `lifetime_last` to the outer `GROUP BY` is a free correctness win under `ONLY_FULL_GROUP_BY` as well. Full `internal/datastore/v2/...` suite is green.

One thing I measured that is worth recording, but is **not** a reason to hold this up. I built a synthetic 1.7M-detection SQLite DB (300 species, ~3% reviewed) and timed the realistic "new species in the last 30 days" query:

| query | time |
|---|---|
| `main` (buggy) | 163 ms |
| this PR | 603 ms |

The cost is that the derived table computes a *lifetime* min/max, so the exclusion has to be evaluated for all 1.7M rows. Without it the plan is a clean covering-index scan of `idx_detection_label_date`; with it every row needs a lookup into `detection_reviews`. I tried `NOT EXISTS` and `NOT IN` as alternatives and both land in the same place (2.7s / 2.2s over the full range vs 2.8s for the `LEFT JOIN`), so there is no cheap swap here.

There is a real fix, but it is a restructure rather than a tweak: bound the aggregate to `detected_at >= start` and replace the lifetime scan with a `NOT EXISTS` short-circuit for prior history. That measured **26 ms**, which is faster than `main` was before the bug fix. I would rather land your correctness fix now than bolt a query rewrite onto it, so I am tracking the optimization separately.

Merging. Thanks for the thorough test coverage on this one.

PR #3982, opened 2026-07-20
fix(frontend): stop species selector overlapping charts when max species selected

Kerakis:
## Summary

Selecting the maximum 10 species on the Analytics → Activity Patterns page makes the species panel overlap the Seasonal Activity Heatmap card header.

`SpeciesSelector`'s root element applied a **fixed** height from its `sizeConfig` (`h-10` at `md`), but the `chip` variant wraps to as many rows as it needs. At 10 selections the extra "Max 10 species selected" warning chip pushes the content to a fourth row, past the fixed 40px, so it overflowed onto the content below instead of pushing it down. The `list` variant has the same latent problem.

## Changes

- `sizeConfig[*].container`: `h-*` → `min-h-*`, so the size preset is a floor rather than a cap. (`min-h-12` was already used on the inner chip wrapper in the same component, so this is consistent with the existing intent.)
- Removed the `min-h-[100px]` band-aid on the analytics species panel wrapper, which existed to paper over the same overflow.

CSS-class-only change; no logic, props, or markup structure touched.

## Testing

- [x] Frontend tests pass (`npm test`) — 2644 passed, 1 failed: `src/lib/stores/settings.test.ts > should convert empty modelPath to null when saving` times out at 5000ms. **Pre-existing on `main`** — reproduced identically in a clean `main` worktree.
- [x] Targeted tests — `SpeciesSelector.test.ts` + `AnalyticsControlBar.test.ts`, 13/13 passed.
- [x] Linting passes — `npm run check:all` clean: prettier, eslint, stylelint, `svelte-check` 0 errors / 0 warnings, all four ast-grep rule sets.
- [x] Go tests (`task test`) — all packages pass except `internal/support`: `TestCollectDockerMounts_NotInContainer` and `TestCollectDeploymentInfo_Orchestration` fail because the suite was run **inside** the dev container. Environmental, and no Go files are touched by this PR.
- [x] Manual testing completed — verified on a Raspberry Pi running this branch: with 10 species selected the panel now pushes the heatmap card down instead of overlapping its header.

All checks were run in the project dev container.

## Preflight Status

All six review passes were run inline by the coordinating agent (sequentially, not as parallel subagents) over the full diff.

- Reviewer 1 (reuse/efficiency): no findings — no new code paths.
- Reviewer 2 (correctness/safety): no findings — no logic changed.
- Reviewer 3 (quality/patterns): one comment reworded to explain *why* the height is a floor. `min-h-<number>` verified as valid Tailwind v4 and already used elsewhere in the file.
- Reviewer 4 (i18n): no new user-facing strings. **[PRE-EXISTING]** `SpeciesSelector` hardcodes several English strings (`Max {n} species selected`, `Clear`, `Loading species...`) instead of using the i18n library — left alone to keep this PR to a single concern.
- Reviewer 5 (integration/wiring): no new fields, functions, events, or options.
- Reviewer 6 (regression/back-compat): `min-h-*` only relaxes a constraint — nothing can render shorter than before. `SpeciesSelector` has one production consumer (`AnalyticsControlBar`). No config keys, API fields, DB columns, CLI flags, or defaults touched. **Secondary-model cross-validation was unavailable** (no independent model CLI in this environment).

### Preflight Certification

- [x] Single concern: one fix
- [x] Preflight passed: all findings resolved; the one pre-existing i18n gap is reported above, not fixed
- [x] Linters clean: `npm run check:all` passes with zero warnings (`golangci-lint` not applicable — no Go changes)
- [ ] Tests pass: frontend and Go suites pass **except** the two pre-existing/environmental failures detailed above, neither reachable from this diff
- [x] No unrelated changes
- [x] Scope complete: no TODO/FIXME left behind
- [x] No regression/backward-compat break
- [x] Breaking changes documented: none to document
- [x] Maintainer-confirm items resolved: none raised
- [x] i18n complete: no new strings
- [x] No secrets or PII

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved species selector sizing so wrapped content displays without overflow across all size variants.
  * Updated selected-option indicators for clearer visual feedback.
  * Improved disabled-option styling and appearance.
  * Adjusted the analytics species selector panel to expand naturally based on its content.

### DIFF
--- a/frontend/src/lib/components/ui/SpeciesSelector.svelte
+++ b/frontend/src/lib/components/ui/SpeciesSelector.svelte
@@ -58,10 +58,9 @@
   // Generate unique ID for dropdown (client-only to avoid SSR hydration mismatch)
   let dropdownId = $state('');
 
-  // Size configurations.
-  // ponytail: container height is a floor, not a fixed height — the chip variant wraps to
-  // multiple rows (chips + "max selected" warning) and a fixed h-* made it overflow onto the
-  // content below instead of pushing it down.
+  // Size configurations. Container height is a floor (min-h-*), not a fixed height: the chip
+  // and list variants wrap to multiple rows, and a fixed h-* made the extra rows overflow onto
+  // the content below instead of pushing it down.
   const sizeConfig = {
     xs: {
       container: 'min-h-6 text-xs',
--- a/frontend/src/lib/components/ui/SpeciesSelector.svelte
+++ b/frontend/src/lib/components/ui/SpeciesSelector.svelte
@@ -58,38 +58,41 @@
   // Generate unique ID for dropdown (client-only to avoid SSR hydration mismatch)
   let dropdownId = $state('');
 
-  // Size configurations
+  // Size configurations.
+  // ponytail: container height is a floor, not a fixed height — the chip variant wraps to
+  // multiple rows (chips + "max selected" warning) and a fixed h-* made it overflow onto the
+  // content below instead of pushing it down.
   const sizeConfig = {
     xs: {
-      container: 'h-6 text-xs',
+      container: 'min-h-6 text-xs',
       chip: 'h-5 px-2 text-xs',
       button: 'h-6 w-6 text-xs',
       input: 'input-xs',
       list: 'text-xs py-1',
     },
     sm: {
-      container: 'h-8 text-sm',
+      container: 'min-h-8 text-sm',
       chip: 'h-6 px-3 text-sm',
       button: 'h-8 w-8 text-sm',
       input: 'input-sm',
       list: 'text-sm py-2',
     },
     md: {
-      container: 'h-10 text-base',
+      container: 'min-h-10 text-base',
       chip: 'h-7 px-3 text-sm',
       button: 'h-10 w-10',
       input: 'input-md',
       list: 'text-sm py-2',
     },
     lg: {
-      container: 'h-12 text-base',
+      container: 'min-h-12 text-base',
       chip: 'h-8 px-4',
       button: 'h-12 w-12',
       input: 'input-lg',
       list: 'py-3',
     },
     xl: {
-      container: 'h-14 text-lg',
+      container: 'min-h-14 text-lg',
       chip: 'h-9 px-4 text-base',
       button: 'h-14 w-14 text-lg',
       input: 'input-lg',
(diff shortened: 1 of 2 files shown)


Kerakis:
Re: the nitpick on `AnalyticsControlBar.svelte` lines 225-245 (named constants for `analyticsSpeciesToggle` / `analyticsSpeciesPanel`) — skipping it deliberately, for three reasons:

1. **Out of diff.** This PR's only change to that file is line 264, dropping a `min-h-[100px]` wrapper class. Lines 225-245 are untouched pre-existing code.

2. **Fixing only the two flagged ids would create the inconsistency the rule exists to prevent.** Inline id literals are the file-wide convention here — six id families across lines 164-261: `analyticsStartDate`, `analyticsEndDate`, `analyticsSourceFilter`, `analyticsSpeciesToggle`, `analyticsSpeciesPanel`, `analyticsSpeciesHint`. Converting two of six is worse than converting none. Converting all six is a refactor, and CONTRIBUTING.md requires exactly one concern per PR — this one is a CSS overflow fix.

3. **Low drift risk in practice.** These are static ids in markup, each used twice within ~30 lines, in a file that constructs no ids dynamically. Flagged 🔵 Trivial.

Glad to open a standalone cleanup PR that converts all six consistently, if the maintainer wants the convention enforced in this file.

Separately, for anyone reading the history: the earlier review at 02:52 UTC is stale. This branch was accidentally cut from a fork `main` that sat 35 commits ahead of upstream, so that review covered unrelated work (`false_positive_filter.go`, `AnalysisSettingsPage.svelte`, and others). The branch has since been rebased onto `upstream/main` and is now 2 commits touching 2 files. The 03:00 UTC review is the one against the real diff.

tphakala:
Reviewed and verified locally.

Confirmed the mechanism: the `.species-selector` root carried a fixed `h-*` from `sizeConfig`, while the chip variant's inner wrapper is `flex-wrap` with its own `min-h-12`. Once selections wrapped past the fixed height the overflow rendered over the next card instead of pushing it down. Relaxing to `min-h-*` is the right fix, and I checked no child depends on the parent being exactly that tall (no `h-full` descendants).

Verified the Tailwind side rather than assuming: built the frontend and grepped the emitted CSS. All five utilities generate, so none of them silently no-op:

```
.min-h-6{min-height:calc(var(--spacing) * 6)}
.min-h-8{min-height:calc(var(--spacing) * 8)}
.min-h-10{min-height:calc(var(--spacing) * 10)}
.min-h-12{min-height:calc(var(--spacing) * 12)}
.min-h-14{min-height:calc(var(--spacing) * 14)}
```

Dropping the `min-h-[100px]` wrapper is safe too: the dropdown is `absolute z-50` inside the `relative` root, so it never needed reserved space, and `git log -S` confirms that value was added as a band-aid for this same overflow.

Targeted tests pass (13/13). Merging.

PR #3985, opened 2026-07-20
fix(audio): attribute exported clips to their species in the logs

tphakala:
`GET /api/v2/system/events/detections` builds `SpeciesEntry.ClipPaths` by reading `species` and `clip_path` back off the `audio_export_success` log line, but the producer never emitted `species`. The endpoint has therefore returned an empty `ClipPaths` for every bucket it has ever served. `SaveAudioAction` had no species field at all, so this adds one, populates it at all three `buildSaveAudioAction` construction sites, and logs it.

The value is lowercased deliberately. The aggregator keys species with an exact map lookup, and the sibling operations it consumes (`approve_detection`, `discard_detection`, `flush_detection`) all log `strings.ToLower(CommonName)`. Emitting the original case would have filed the clip paths under a second, phantom species row with zero counts while the real row kept its empty `ClipPaths`, which is the same defect one layer further down.

`clip_path` also changes from a bare basename to the clip's name relative to the export directory. The `year/month` segments are what let a consumer resolve the value back to a file, and the export directory itself stays stripped because support dumps get read by someone other than the operator. A path that would escape the export directory falls back to the basename rather than emitting `..`.

### Other fixes on the two export log sites

The failure guard classified shutdown by inspecting the context rather than the error. ENOSPC, corrupt PCM, or a missing FFmpeg binary hit while a shutdown happened to be in progress were all logged at Debug under a message asserting they had been cancelled; on a container that OOM-restarts, that misattributed every in-flight export failure in the restart window. It now tests only the error. `resolveExportGainDB` joins the cancellation onto a measurement that cancellation turned terminal, so a genuine shutdown still lands at Debug. I checked every path that cancellation can actually interrupt: FFmpeg wraps `ctx.Err()` at each exit, FLAC checks per chunk, AAC and Opus check before the temp file is created, and the stretches that take no context cannot be interrupted by cancellation in the first place.

`encode_ms` covered the loudness measurement as well as the encoder, so on a Pi with a long extended capture it blamed the codec for a full-clip EBU R128 pass. It is now split into `measure_ms` and `encode_ms`. On a failure that never reached the encoder, `gain_db` and `encode_ms` are omitted rather than reported as a zero indistinguishable from a real value, and the failure line now carries the timings it had already measured.

Several doc comments misstated the job queue's reachable log strings, the encoders' context handling, and which attempts `LogJobRetryScheduled` records. Corrected.

### Format-downgrade logs

The two downgrade sites disagreed with each other. The bat downgrade was Info once per process; the stranded-without-encoder fallback was Warn on every single clip, forever. Both mean "you are not getting the export format you configured", so both are Warn now and both are guarded.

The guards are keyed on the format and sample rate rather than a bare `sync.Once`. `Export.Type` is hot-reloadable and the capture rate is per source, so a single `sync.Once` let whichever condition fired first silence every later, different one, leaving the operator an explanation naming a format they had since changed away from.

### Notes for reviewers

`encode_ms` and the failure line both shipped yesterday in #3984 and are in no tagged release (`git tag --contains` is empty, newest tag is `20260716`), so narrowing `encode_ms`'s meaning and making `gain_db` conditional break no existing parser.

Once clip paths actually attach, an export whose success line falls in a later hour bucket than its approval can add a zero-count species row to that later bucket. That is pre-existing aggregator behaviour that was simply unreachable while `species` was never emitted; I left it alone rather than change the endpoint's semantics in a logging PR.

## Release notes
- audience: user
- summary: The detection events API now reports the audio clip files recorded for each species instead of always returning an empty list, and clip export log lines separate loudness-measurement time from encoding time.
- feature: native encoder rollout
- breaking: none
- config: none
- schema: none

## Summary by CodeRabbit

- **Improvements**
  - Audio export logs now include clearer species information and safer relative clip paths.
  - Export timing is reported separately for loudness measurement and encoding.
  - Failure logs more accurately distinguish cancellation from other errors and omit unavailable details.
  - Logged bitrate values now reflect effective encoder settings.
  - Format fallback warnings are tracked independently for each format and sample-rate combination.

- **Reliability**
  - Improved handling of export cancellation and measurement failures.
  - Expanded validation ensures consistent species attribution and safer logging behavior.

### DIFF
--- a/internal/analysis/processor/actions_database.go
+++ b/internal/analysis/processor/actions_database.go
@@ -496,7 +496,7 @@ func (a *SaveAudioAction) Execute(ctx context.Context, _ any) error {
 			logger.String("component", "analysis.processor.actions"),
 			logger.String("detection_id", a.CorrelationID),
 			logger.Error(err),
-			logger.String("path", outputPath),
+			logger.String("clip_path", a.relativeClipPath(outputPath)),
 			logger.String("operation", "audio_export_stat"))
 	}
 
@@ -561,7 +561,7 @@ func (a *SaveAudioAction) Execute(ctx context.Context, _ any) error {
 				logger.String("component", "analysis.processor.actions"),
 				logger.String("detection_id", a.CorrelationID),
 				logger.Any("note_id", a.NoteID),
-				logger.String("clip_path", outputPath),
+				logger.String("clip_path", a.relativeClipPath(outputPath)),
 				logger.Error(err),
 				logger.String("operation", "prerender_submit"))
 		}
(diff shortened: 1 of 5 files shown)


tphakala:
@coderabbitai review

PR #3988, opened 2026-07-20
fix(notifications): make template preview and Reset reflect real defaults

dk1:
## Summary

The New Species notification template UI and the **Send Test** preview both
showed default text that no real notification ever produces. Two related bugs.

## Backend

**1. `renderTemplateWithDefault()` ignored its own default for empty templates.**
It only used the `defaultVal` argument when template rendering *failed* (bad
syntax). An empty/uncustomized template returned `""` unconditionally — so a
test-fire with no template configured produced a notification with a genuinely
**blank** title and message, not a preview of the real default. Empty template
now falls through to `defaultVal`, matching what the function name already
promises. (Its only callers are the test-fire handler.)

**2. Those default values were invented preview text.** `"New Species: Test Bird
Species"` and `"First detection of ... at Fake Test Location"` matched neither
the seeded rule nor what a real detection notification produces. Real
notifications fall back to the seeded rule's own **name** for the title and
`"<species> detected with <confidence>% confidence"` for the message (see
`dispatcher.go`'s `detectionMessage()`).

Exported `DefaultDetectionMessage()` and the `RuleNameNewSpecies` constant from
`internal/alerting` so the dispatcher and the test-fire handler share **one**
definition instead of keeping separate copies that drift apart. The message a
real detection produces is unchanged (identical `%.0f` formatting as before —
verified by inspection and covered by a new dispatcher test).

## Frontend (`NotificationsSettingsPage.svelte`)

The template editor pre-filled a hardcoded `defaultTemplate`
(`New Species: {{.CommonName}}` / `First detection of ...`) on load and on
**Reset**, presenting it as "the default" even though it matched neither the
backend default nor the shipped config. Load and Reset now clear the fields to
empty, so an uncustomized template means "use the built-in default" (rendered
correctly by the backend fix). The inputs already carry example placeholders
(`e.g., New Species: {{.CommonName}}`), so the empty state stays
self-explanatory.

## Testing

- Added `TestDefaultDetectionMessage` (alerting) and
  `TestCreateTestNewSpeciesNotification_EmptyTemplate_UsesRealDefault`
  (notifications API), which asserts the empty-template fallbacks equal the
  seeded rule name and `alerting.DefaultDetectionMessage` exactly.
- Frontend: `prettier`, `eslint`, and `svelte-check` (0 errors / 0 warnings)
  pass. No frontend test asserted the old hardcoded behaviour.
- **Note:** I could not run the Go build/tests in my local environment (no Go
  toolchain on this machine); please rely on CI for the Go tests. The existing
  `_Success` test is unaffected — it configures a non-empty template that still
  renders.

## Scope

Backend: `internal/alerting/{constants,defaults,dispatcher}.go` +
`internal/api/v2/notifications/notifications.go` and their tests. Frontend: one
Svelte file. No i18n keys changed.

### DIFF
--- a/frontend/src/lib/desktop/features/settings/pages/NotificationsSettingsPage.svelte
+++ b/frontend/src/lib/desktop/features/settings/pages/NotificationsSettingsPage.svelte
@@ -301,12 +301,6 @@
     }))
   );
 
-  const defaultTemplate = {
-    title: 'New Species: {{.CommonName}}',
-    message:
-      '{{.ImageURL}}\n\nFirst detection of {{.CommonName}} ({{.ScientificName}}) with {{.ConfidencePercent}}% confidence at {{.DetectionTime}}.\n\n{{.DetectionURL}}',
-  };
-
   let isServiceFormValid = $derived.by(() => {
     switch (selectedService) {
       case 'discord':
@@ -546,13 +540,13 @@
 
       if (data.templates?.newSpecies) {
         templateConfig = {
-          title: data.templates.newSpecies.title ?? defaultTemplate.title,
-          message: data.templates.newSpecies.message ?? defaultTemplate.message,
+          title: data.templates.newSpecies.title ?? '',
+          message: data.templates.newSpecies.message ?? '',
         };
         editedTitle = templateConfig.title;
         editedMessage = templateConfig.message;
       } else {
-        templateConfig = { ...defaultTemplate };
+        templateConfig = { title: '', message: '' };
         editedTitle = templateConfig.title;
         editedMessage = templateConfig.message;
       }
@@ -567,7 +561,7 @@
         originalPushSettings = JSON.parse(JSON.stringify(pushSettings));
       }
     } catch {
-      templateConfig = { ...defaultTemplate };
+      templateConfig = { title: '', message: '' };
       editedTitle = templateConfig.title;
       editedMessage = templateConfig.message;
     } finally {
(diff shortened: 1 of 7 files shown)


dk1:
Closing this. On a closer look the change is built on a mistaken premise.

These `notification.templates.newspecies` fields ship **non-empty** by default
(`internal/conf/defaults.go` seeds `"New Species: {{.CommonName}}"` and the
matching message), and the real consumer — `detection_consumer.go` —
deliberately treats an empty field as "user wants this empty" (falling back to
`"New Species Detected: …"` only on a render *failure*). This PR instead made
the test-fire preview fall back to the alert-rule name and
`DefaultDetectionMessage`, which belong to a **different** subsystem (alert-rule
→ bell), so the preview would show text a real notification never produces — the
opposite of the intended "make the preview match reality."

The frontend Reset change is also a regression: it clears the fields to empty,
and saving an empty template persists it (the viper default only applies when
the key is absent), so Reset would leave real new-species notifications blank
instead of restoring the non-empty default.

The only sound part left is a small alerting-side refactor (single source of
truth for the rule name / default message), which has no consumer once the
test-fire change is dropped — not worth landing on its own.

If the underlying worry is that an emptied template silently produces empty
notifications, the only way that happens *unexpectedly* is the settings-PUT
erasure tracked in #3993; the blank-means-empty behavior itself is intentional.
A UX clarification (help text: "leave blank for no title") would be a better
direction than changing the fallback semantics.

PR #4020, opened 2026-07-23
perf(hls): bound the live audio queue by bytes, not chunk count

tphakala:
## Summary

The per-stream PCM queue between the audio router and the HLS feed loop was a 1024-slot channel with no byte bound (`make(chan audioChunk, defaultReadBufferSize)`). With RTSP ingest delivering 32 KiB frames, a stalled consumer could hold 1024 x 32 KiB = 33.5 MB of PCM per stream, about 349 seconds of 48 kHz mono audio. None of that depth is servable: a live listener sits roughly 4 seconds behind the edge (hls.js keeps `liveSyncDurationCount` segments buffered), so anything past the first couple of segments is encoded into segments no player will ever request. Because Go sizes the heap from its high-water mark and returns pages lazily, a single hiccup that fills the queue raises RSS for the life of the process. On a platform where 512 MB is still an unmet target this is worth bounding whether or not it turns out to dominate a given heap profile.

The fix bounds the queue by queued PCM bytes instead of by chunk count. Chunk size is a property of the producer, not of HLS: RTSP ingest delivers 32 KiB frames, a directly captured sound card delivers miniaudio's default 10 ms period (about 960 bytes at 48 kHz mono), and a route whose source rate differs from the consumer rate has a resampler in between that changes the size again. A chunk-count bound therefore means a different memory ceiling and a different amount of buffered audio for every source; one small enough to bound RTSP frames would leave a sound card with a fraction of a second of slack. A byte budget gives every producer the same ceiling and the same wall-clock depth.

The native encoder path caps queued PCM at 512 KiB (about 5.4 seconds), dropping the oldest chunks past that. That is generous against the roughly 4 second live edge and the roughly 2.5 ms of CPU per second of audio the encoder actually costs, so the queue only ever fills during a multi-second stall, and audio that far behind the edge is past the point any player would still ask for it. Worst case per stream falls from 33.5 MB to 512 KiB plus the 48 KB slot ring.

The legacy encoder path is deliberately left byte-unbounded and keeps the prior 1024-slot behaviour exactly. It feeds PCM to a separate process that cannot see a gap in what it is handed, and its playlist timeline epoch is computed once, so a dropped chunk there would shift the timeline permanently rather than briefly. The native path has no such coupling: it hands the muxer each chunk's own capture timestamp, so a drop shows up as a timestamp jump the muxer accounts for and the timeline stays true.

## Implementation notes

- New `audioFeed` type wraps the channel with an `atomic.Int64` byte counter and a byte budget. `Write` evicts the oldest chunks down to the budget before a non-blocking send; the send never blocks the router's dispatch goroutine. The three feed loops each report a dequeue via `release()` immediately after receiving, before any early return, so the counter cannot leak on a mid-loop exit.
- The counter is advisory under the single-producer / single-consumer model (bounded one-chunk skew), which is sufficient for a queue bound and never used for correctness.
- Minor operator-facing change: the drop warning log message and fields changed (`channel_cap` int replaced by `queued_bytes` / `max_queued_bytes` int64, message wording updated). This is a throttled degraded-condition WARN, not routine parsed output.

## Test plan

- [x] New `audio_hls_feed_test.go` covers the byte bound for large (32 KiB) and small (960 B) frames, small-frame time depth, drop-oldest ordering, `release` accounting round-trips, oversized-chunk admission, the slot-exhaustion drop path, and the unbounded path.
- [x] `go test -race ./internal/api/v2/audio/...` green.
- [x] `go vet` clean, `golangci-lint run` reports 0 issues.
- [x] Regression review confirms the legacy path is byte-for-byte behaviourally identical and no config, API, DB, or CLI surface is touched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HLS audio streaming stability by managing buffered audio more effectively.
  * Prevented excessive audio backlog and memory growth during slow or overloaded playback conditions.
  * Preserved smooth audio delivery across native, FFmpeg, and Windows playback paths.
  * Added safer handling when audio consumers close or cannot accept additional data.

* **Documentation**
  * Clarified how audio buffering capacity affects playback behavior and resource usage.

### DIFF
--- a/internal/api/v2/audio/audio_hls.go
+++ b/internal/api/v2/audio/audio_hls.go
@@ -112,6 +112,25 @@ const (
 	// gives the native path's muxer configuration something to be checked
 	// against rather than a second bare literal.
 	hlsConsumerChannels = 1
+
+	// hlsFeedQueueUnbounded disables an audio feed queue's byte budget, leaving
+	// the channel's slot count (defaultReadBufferSize) as the only bound.
+	//
+	// This is what the FFmpeg path uses, deliberately. Its consumer writes PCM
+	// into a FIFO that the FFmpeg process reads, and FFmpeg cannot see a gap in
+	// what it is handed: it encodes whatever arrives as one continuous stream.
+	// Since pdtOffset is computed once from the first playlist and then applied
+	// to every later one, every dropped chunk shifts EXT-X-PROGRAM-DATE-TIME
+	// permanently behind wall-clock time, by the duration dropped. A budget
+	// tight enough to bite before fifoWriteTimeout (30s) gives up on the stall
+	// while FFmpeg is still willing to wait it out, and pays for it with a
+	// timeline that never recovers.
+	//
+	// The native path has no such coupling: it hands the muxer each chunk's own
+	// capture timestamp, so a drop shows up as a timestamp jump the muxer
+	// accounts for, and the timeline stays true. That is why it can afford
+	// nativeHLSFeedQueueBytes and this path cannot.
+	hlsFeedQueueUnbounded = 0
 )
 
 // HLSStreamInfo contains information about an active HLS streaming session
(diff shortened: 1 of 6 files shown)


tphakala:
Applied the `evictOldest` extraction in 5033c0f. It also welds the receive-and-release pairing into one spot, which reinforces the every-dequeue-releases-its-bytes invariant this change relies on. Confirmed it inlines at all three call sites (`go build -gcflags=-m`), so the per-frame hot path pays nothing.

PR #4093, opened 2026-08-05
fix(media): preserve request timeout errors

### LINKED ISSUE #4085: media: ctx.Err() collapses Canceled and DeadlineExceeded, masking server timeouts
### Summary

Several media serve paths treat any non-nil request-context error as "client disconnected" and return `nil` (success), which collapses `context.Canceled` and `context.DeadlineExceeded` into the same branch. A genuine server-side timeout (`DeadlineExceeded`) is then silently swallowed as if the client had gone away, so it is neither logged as a server fault nor surfaced with an appropriate status.

### Locations (`internal/api/v2/media/media.go`)

```
1131:  if ctx.Request().Context().Err() != nil { return nil // Client disconnected }
1273:  if ctx.Request().Context().Err() != nil { return nil // Client disconnected }
1391:  if ctx.Request().Context().Err() != nil { return nil // Client disconnected }
1414:  if ctx.Request().Context().Err() != nil { return nil // Client disconnected }
2920:  if ctx.Err() != nil { ... }
```

### Suggested direction

Differentiate the two causes, e.g. `if errors.Is(ctx.Err(), context.Canceled)` for a true client abort (return `nil`/no-op), and treat `context.DeadlineExceeded` distinctly (log it, and where a response is still possible return 504/408) so server-side timeouts are not misreported as client disconnects. Low severity: media serving is mostly best-effort, but the pattern masks a real failure mode and is worth a consistent cleanup across all five sites.

Pre-existing; surfaced by static analysis and the Gemini cross-check during review of PR #4083 (out of that PR's scope).

w3lld1:
## Summary

I separated canceled media requests from request deadlines so server-side timeouts are no longer silently treated as client disconnects.

## Changes

- preserve the no-response behavior for canceled clients
- translate expired request deadlines into the existing HTTP 408 timeout response
- apply the same handling across clip extraction, audio processing, and processed spectrogram generation
- add regression coverage for active, canceled, and expired request contexts

## Testing

- `go test -tags noembed,skipfrontend,notflite,goaac_simd -race ./internal/api/v2/media -count=1`
- `golangci-lint run --build-tags=noembed,skipfrontend,notflite,goaac_simd ./internal/api/v2/media/...`
- `go vet -tags noembed,skipfrontend,notflite,goaac_simd ./internal/api/v2/media`
- `git diff --check`

The full native TensorFlow Lite build and repository-wide suite are deferred to CI because the local environment does not provide the required TensorFlow Lite C headers and library.

## Preflight Status

I reviewed the complete diff for reuse, correctness and safety, code quality, integration consistency, and backward compatibility. The change is limited to one bug fix, introduces no user-facing strings, and leaves no unresolved preflight findings.

## Related Issues

Fixes #4085

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of canceled media-processing requests so interrupted operations end cleanly without unnecessary responses.
  * Requests that exceed their deadline now return a clear request-timeout response.
  * Audio extraction, processed-audio generation, and spectrogram processing now handle cancellation and timeout conditions consistently.
  * Improved error handling provides more reliable behavior across media-processing workflows.

### DIFF
--- a/internal/api/v2/media/media.go
+++ b/internal/api/v2/media/media.go
@@ -470,7 +470,7 @@ func (c *Handler) translateSecureFSError(ctx echo.Context, err error, userMsg st
 			logger.Bool("tunneled", isTunneled),
 			logger.String("tunnel_provider", tunnelProvider),
 		)
-		return c.HandleError(ctx, err, "Request timed out", http.StatusRequestTimeout)
+		return c.HandleError(ctx, err, requestTimeoutMsg, http.StatusRequestTimeout)
 	case errors.Is(err, context.Canceled):
 		c.LogInfoIfEnabled("Request canceled by client",
 			logger.Error(err),
--- a/internal/api/v2/media/media.go
+++ b/internal/api/v2/media/media.go
@@ -65,6 +65,7 @@ const (
 const (
 	headerAcceptRanges = "Accept-Ranges"
 	acceptRangesBytes  = "bytes"
+	requestTimeoutMsg  = "Request timed out"
 )
 
 // Cache duration in seconds for HTTP Cache-Control headers on media responses.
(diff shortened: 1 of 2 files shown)


w3lld1:
Thanks — extracted the timeout text into the named `requestTimeoutMsg` constant in `32a7bd9`, keeping the deadline handling behavior unchanged. The focused source assertion and `git diff HEAD^ --check` pass.

tphakala:
Thank you for contributing!

PR #4126, opened 2026-08-15
feat(models): variant RAM floors, benchmarks, and BirdNET v3.0 gallery availability

tphakala:
## Summary

This populates per-variant hardware metadata in the model catalog and makes BirdNET v3.0 installable from the gallery.

It adds `MinRAMMB` floors to every global variant of BirdNET v3.0, Perch v2, and BirdNET v2.4, an `openvino-gpu-intel-gen12` exclude token to the v3.0 fp16 variant, and measured per-device benchmarks to the v3.0 fp32 and fp16 variants. All values are mirrored verbatim from the acoustic-models model manifests. The hardware recommender already reads these fields, but no shipped variant populated them, so the RAM hard filter never fired and rankings compressed toward the backend term.

With the RAM floors in place the recommender can keep the heavy global builds (fp32 is 557 MB, fp16 is 279 MB) off hosts that cannot run them, so the BirdNET v3.0 catalog entry is un-hidden and installable from the gallery. It stays labelled a developer preview. Its `CommercialUse` flag is corrected to true, since the model is CC-BY-SA-4.0 which permits commercial use.

Benchmarks currently exist upstream only for BirdNET v3.0; Perch v2 and v2.4 have no measured latencies in the manifests yet, so their latency ranking is unchanged.

## Notes

The recommender maps an amd64 host to a generic `x86` device class, while the manifest benchmark rows name a specific CPU (`x86-i7-1260P`), so the x86 latency rows do not yet influence ranking on amd64 hosts. The x86 rows are included for fidelity to the manifest and are consumed on ARM device classes today. Wiring up x86 device-class matching would flip the v3.0 recommendation on an Intel iGPU host from fp16 to fp32 (fp32 benchmarks faster on the CPU than fp16 on the iGPU), which is a recommender-policy decision, so it is deliberately deferred to a separate change.

## Test Plan

- [x] `go test -race ./internal/classifier/... ./internal/api/v2/models/...` passes
- [x] `golangci-lint run` reports 0 issues
- [x] New `TestEmbeddedCatalog_VariantRequirementsPopulated` pins every RAM floor, the fp16 exclude token, and all eight v3.0 benchmark rows (device, backend, latency, RSS)
- [x] Updated visible/hidden catalog counts, wildlife category count, and the v3.0 visibility/commercial-use tripwires

## Summary by CodeRabbit

* **New Features**
  * BirdNET v3.0 is now visible in the model gallery and available for commercial use.
  * Added minimum memory requirements and benchmark information for supported model variants.
  * Added memory requirements for Perch v2 and hidden BirdNET v2.4 variants.
  * Updated BirdNET v3.0’s developer-preview description.

### DIFF
--- a/internal/classifier/birdnet_v3_onnx_test.go
+++ b/internal/classifier/birdnet_v3_onnx_test.go
@@ -117,8 +117,8 @@ func TestBirdNETV3_LoaderRegistered(t *testing.T) {
 	assert.True(t, ok, "BirdNET v3.0 must have an OpenVINO-capable secondary builder")
 }
 
-// TestBirdNETV3_CatalogEntry guards the catalog entry shape: correct repo, hidden
-// during preview, and the expected model/labels/geomodel/taxonomy companion files.
+// TestBirdNETV3_CatalogEntry guards the catalog entry shape: correct repo, visible
+// in the gallery, and the expected model/labels/geomodel/taxonomy companion files.
 func TestBirdNETV3_CatalogEntry(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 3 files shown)


tphakala:
@coderabbitai review

tphakala:
Addressed the test-parallelization nitpick in efc931804: the per-row benchmark subtests now call t.Parallel() (they only read immutable catalog data, so it is safe). Thanks!

PR #4127, opened 2026-08-15
feat(models): region axis for the gallery recommender and recommended-first ordering

tphakala:
## What

Completes the region axis of the hardware- and region-aware model gallery recommender. The hardware axis and the region resolver already shipped; this wires the resolved region into the pure ranker, sorts the gallery recommended-first, and converts the entry-level incompatibility message to a translatable code.

## Changes

**Recommender (`internal/classifier/recommend`)**

- New `Input.ResolvedRegion` and two additive score terms: `region.matched` (+100) when a variant's regional slice matches the host's resolved region, and `region.global_fallback` (+50) for the global variant when no slice matches.
- A per-entry `hasMatchingRegion` precompute keeps the global fallback firing whenever the entry ships no matching slice, so a wrong-region slice can never outrank the global model on hardware terms alone.
- `Rank` now takes `*Input` (the struct grew past gocritic's hugeParam threshold); the documented "Input is never mutated" contract is unchanged and still covered by a test.

**API (`internal/api/v2/models`)**

- `resolveRecommendRegion` resolves the region slug from settings, honoring a pinned region even without coordinates and otherwise gating auto-resolution on a configured location, delegating the mode classification to the shared `region.Select`.
- `GetModelCatalog` sorts entries recommended-first (stable, recommendation-eligible requests only) and reports the ONNX-unavailable case as the structured code `backend.onnx_unavailable` instead of a raw, untranslatable ONNX Runtime error string. The specific error is still returned in full on an install attempt.

**Frontend**

- The entry-level incompatibility now renders through the i18n reason path with a localized fallback, so a structured code is never shown raw.
- Three new `analysis.gallery.reasons.*` keys, translated across all 15 locales.

## Testing

Region-matched scoring is engine-only until regional variants land in the catalog, so it is validated with synthetic-catalog table tests, including a regression test proving a wrong-region slice never outranks the global model, plus tests for the recommended-first ordering and the region resolver (including the pinned-without-coordinates path). `go test -race`, `golangci-lint`, frontend typecheck and unit tests all pass.

## Summary by CodeRabbit

* **New Features**
  * Model recommendations now prioritize entries compatible with the device and available backends.
  * Recommendations account for regional availability, with clear explanations for regional matches and global fallbacks.
  * Model incompatibility reasons are shown as localized, user-friendly messages.

* **Localization**
  * Added translated recommendation and compatibility messages across supported languages.

* **Bug Fixes**
  * Replaced raw technical runtime errors with understandable localized explanations.
  * Improved ordering of recommended models in the catalog.

### DIFF
--- a/frontend/src/lib/desktop/features/settings/pages/AnalysisSettingsPage.svelte
+++ b/frontend/src/lib/desktop/features/settings/pages/AnalysisSettingsPage.svelte
@@ -68,7 +68,7 @@
   import { buildAppUrl } from '$lib/utils/urlHelpers';
   import { toastActions } from '$lib/stores/toast';
   import { formatBytes, formatNumber } from '$lib/utils/formatters';
-  import { pickPreselectedVariant } from '$lib/utils/variantSelection';
+  import { pickPreselectedVariant, reasonKey } from '$lib/utils/variantSelection';
   import { safeArrayAccess } from '$lib/utils/security';
   import { loggers } from '$lib/utils/logger';
   import { t } from '$lib/i18n';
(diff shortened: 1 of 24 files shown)


tphakala:
@coderabbitai review

PR #4156, opened 2026-08-21
fix(models): batched low-severity gate follow-ups (HF failover, notflite tests, region-mode constants)

tphakala:
## Summary

Folds several deferred low-severity cleanups from earlier model-gallery reviews into one change. It is a behavior-preserving refactor plus test hygiene; existing users see no functional difference.

**HuggingFace endpoint resolution.** Extracts a shared `huggingFaceOverrideSource` helper so `ResolveHuggingFaceEndpoint` and `ResolveHuggingFaceEndpointChain` share the settings-then-`HF_ENDPOINT` precedence and cannot silently diverge if that order ever changes. The resolved endpoint, the chain contents, the warn-on-invalid behavior (only the single resolver warns; the chain falls back silently), and the redacted log value are all unchanged.

**Download resilience.** The model-download HTTP client now uses a `Transport` cloned from `http.DefaultTransport` with a `ResponseHeaderTimeout` of 30s, so a host that completes TCP+TLS then stalls before sending response headers fails over to the mirror promptly instead of holding the full 30-minute budget. The response body remains bounded only by the total timeout, so legitimate slow large-file downloads are unaffected. The endpoint-state `save()` path now logs its previously-silent temp-file write/close failures at debug level, matching its other branches.

**Startup.** The config directory is resolved once in `Start` and threaded into `loadModelCatalog` and `initModelManager` rather than being resolved twice; each consumer's on-error behavior is preserved exactly.

**Tests.** `TestNewBirdNET_LocaleNormalization` and the BirdNET v2.4 case of `TestHasNativeRangeFilter` now respect the `notflite` build tag (they assume a linked TFLite backend), so ONNX-only builds stay green. Gateway-status failover is now driven end to end for 502 and 504, not only 503, via a table-driven test.

**Frontend.** The region-mode string literals are centralized as `DEFAULT_REGION_MODE` and `GLOBAL_REGION_MODE` and used at the comparison sites in the region selector and analysis settings page (the DOM `value=` attributes stay literal). A param-aware assertion confirms the remove-success toast interpolates the model name.

## Test Plan

- [x] `go test` (default build) and `go test -tags notflite` green for internal/classifier and internal/conf
- [x] `golangci-lint run` on changed packages: 0 issues
- [x] `npm run check:all` (typecheck, eslint, stylelint, ast-grep): clean
- [x] Frontend unit tests for the region selector, gallery page, and variant helpers pass
- [x] Pre-push quality gate: no Critical/High/Medium findings

## Summary by CodeRabbit

- **Improvements**
  - Improved model download responsiveness by detecting stalled server responses sooner.
  - Enhanced download reliability by automatically switching to an available mirror when a server is temporarily unavailable.
  - Improved handling of custom model-hosting endpoints and fallback mirrors.
  - Clarified region selection behavior, including automatic and global modes.
  - Improved diagnostics when model configuration cannot be located.

### DIFF
--- a/frontend/src/lib/desktop/features/settings/components/ModelRegionSelector.svelte
+++ b/frontend/src/lib/desktop/features/settings/components/ModelRegionSelector.svelte
@@ -17,7 +17,11 @@
   import { t, getLocale } from '$lib/i18n';
   import { fetchModelRegions, fetchRegionCoverageMap } from '$lib/utils/modelsApi';
   import { localizedCountryNames } from '$lib/utils/countryNames';
-  import { normalizeRegionMode } from '$lib/utils/variantSelection';
+  import {
+    normalizeRegionMode,
+    DEFAULT_REGION_MODE,
+    GLOBAL_REGION_MODE,
+  } from '$lib/utils/variantSelection';
   import { loggers } from '$lib/utils/logger';
   import { birdnetSettings, settingsActions } from '$lib/stores/settings';
   import type { ModelRegionsResponse, RegionOption } from '$lib/types/models';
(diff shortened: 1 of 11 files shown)


tphakala:
Good call, done in ac4d83888. Extracted `downloadTotalTimeout` for the 30-minute budget so it matches the `downloadResponseHeaderTimeout` const right next to it.

PR #4161, opened 2026-08-22
fix(models): batched gallery follow-ups (reload rollback tests, region-mode default, hardware-token vocab)

tphakala:
## Summary

Three low-risk model-gallery follow-ups from recent reviews, folded into one PR.

**Reload rollback coverage.** Adds `internal/classifier/birdnet_reload_rollback_test.go` covering `reloadModelInternal`'s transactional rollback, which was previously 0% unit-covered (`NewOrchestrator` skips without a real model, and the primary-swap tests only reach the `orchestrator == nil` path). Two early-failure subtests inject a taxonomy-read failure (no native model needed) and assert the up-front-mutated state (ModelInfo, taxonomy, settings) is restored across both entry points, variant-swap and settings-reload. A third subtest uses a new nil-by-default `reloadInitFn` seam to install a fake backend and fail after it, exercising the rollback branch that tears down the failed backend and reverts the runtime triplet and model version, none of which the early-failure path reaches. The seam is unexported, nil in every production construction path, and semantically identical to the direct `initializeModel` call.

**Region-mode default.** `settings.ts` now sources the `modelRegion` store default from the shared `DEFAULT_REGION_MODE` constant instead of a bare `'auto'` literal (value unchanged), so the default matches the constant every comparison site already uses.

**Hardware-token vocabulary.** Aligns the client fallback `variantHardwareClass` to the server's authoritative tokens (`gpuNvidia`/`gpuIntel` instead of `gpu`/`intelGpu`), so the same hardware renders the same variant chip label whether the token comes from the server or from the older-server client fallback. Drops the now-unreachable `gpu`/`intelGpu` keys from all locale files and regenerates the i18n types.

## Test Plan

- [x] `go test -race ./internal/classifier/` (full package, green)
- [x] `golangci-lint run` clean, `go vet` clean, `gofmt` clean
- [x] `npm run check:all` (typecheck, lint, css, ast) clean
- [x] `npm run i18n:validate:full` (sync, types, usage) clean
- [x] `npm test` for the touched frontend suites (variantSelection, settings, gallery, region selector)

## Summary by CodeRabbit

- **Improvements**
  - Model gallery hardware labels now clearly distinguish NVIDIA and Intel GPUs, with redundant generic labels removed across supported languages.
  - Hardware-based model selection now uses consistent NVIDIA and Intel GPU classifications.
  - Default model region settings now follow the application’s configured regional mode.
  - Improved model reload reliability: failed reloads now restore the previously working model and settings instead of leaving an incomplete replacement active.


tphakala:
Good call, done in 5dca86dd1. Pulled the hardware-class tokens into a single HARDWARE_CLASS constant object, derived VariantHardwareClass from it, and switched the mapper, the label resolver, and the tests over to the constants. Values and i18n keys are unchanged.

PR #4167, opened 2026-08-23
fix(audiocore): reset RTSP restart backoff after stable recovery

### LINKED ISSUE #4080: bug: RTSP recovery backoff escalates across widely-spaced transient stalls, causing ~5x more downtime than the fault, and trips the liveness watchdog

 2026-07-16T19:54:20Z

### Install method

Docker run / docker-compose (manual, ghcr.io/tphakala/birdnet-go)

### Device type

Raspberry Pi 5

Debian GNU/Linux 13 (trixie)

### Audio source

Network audio stream - RTSP / RTSPS (e.g. IP camera)

### Hardware details

BirdNET-Go host:

- Raspberry Pi 5, 8GB RAM
- Storage: NVMe SSD mounted at /mnt/nvme
- 2GB swap configured
- Docker CE 29.6.2 from Docker's official apt repo (not distro-packaged), with the Compose plugin
- OS: Debian 13
- wired ethernet

RTSP bridge host — pnet-sbc-2

- Orange Pi Zero 2W, 4GB RAM 
- Allwinner H618, quad-core Cortex-A53
- OS: Orange Pi 1.0.2 Bookworm, kernel 6.1.31-sun50iw9 (Debian 12)
- 64GB microSD
- wired ethernet via IO expansion board
- Runs mediamtx v1.15.2 (linux_arm64) + ffmpeg 5.1.8-0+deb12u1

Audio chain

- Clippy EM272Z1 mono mic (plug-in-power, TRS) → UGREEN US205 USB audio interface → 10m active USB extension → Orange Pi
- ffmpeg command: -f alsa -i plughw:Device,0 -ar 48000 -ac 1 -c:a libopus -b:a 128k -rtsp_transport tcp -f rtsp rtsp://localhost:8554/birdmic

### Description

My RTSP audio source experiences a brief network stall (3–12s) on a consistent ~1200s cycle. FFmpeg's default 10s socket timeout means the longer stalls kill the process, which is expected and fine. What isn't expected is the recovery backoff escalates 5s → 10s → 20s → 40s → 80s across these events, even though they are 20+ minutes apart with entirely healthy streaming in between. The resulting self-imposed silence then trips the liveness watchdog, which raises an audio source silence detected error and force-restarts the source.

Net effect over a recent 24h period is roughly 300 seconds of deliberate backoff silence in response to underlying faults totalling maybe 60 seconds. The recovery mechanism is costing several times more audio than the problem it's recovering from.

What I expected: restart_count to reset after a sustained healthy period, so that transient faults 20 minutes apart are each treated as a first failure with a short backoff, rather than as a consecutive failure streak.

What happens instead: the counter escalates across events separated by 1200s of healthy streaming, and only resets when the liveness watchdog intervenes.

**Three related observations**

- restart_count doesn't reset after healthy operation. Consecutive teardowns 20–200 minutes apart, each with a clean ~4s reconnect and continuous healthy streaming in between, still ladder 1→2→3→4→5. Twenty minutes of flawless operation arguably shouldn't count as a consecutive failure.

- The liveness watchdog measures silence the application itself creates. During backoff, ffmpeg-stream deliberately captures nothing. liveness observes "no audio for N seconds" and raises an error — the application alarming on a condition it caused on purpose. The two subsystems appear to have no shared state; a "backoff in progress, expected silence until T" flag would eliminate this class of false alarm entirely.

- The two mechanisms override each other. Because liveness force-restarts on breach, it truncates the backoff it doesn't know about. Example: a 40s backoff scheduled at 11:39:26 was cut short by liveness at 11:40:00 (~34s actual). This means the configured backoff schedule isn't what actually runs, and tuning silenceThreshold silently changes real backoff behaviour.

**Config testing**

realtime.audio.watchdog.silenceThreshold works as documented. At default (0), alerts fire at ~33–37s. Set to 60, alerts moved to ~61–63s. Confirmed across multiple cycles.
realtime.audio.watchdog.retryBackoff: 10 appears to be inert. The ladder ran 5 → 10 → 20 → 40 both before and after, identical to default — it acts as neither a base nor a cap.

Note a side effect: raising silenceThreshold to 60 made things worse. It removed the ceiling liveness was providing, so the ladder ran to restart_count: 5 (80s) every cycle instead of stopping at 4, increasing actual downtime while quieting the alert.

**Question**

Is the escalating backoff intended for transient RTSP stalls, given the server itself is healthy throughout and accepts a new connection within ~4s every time? Backoff makes sense for an unavailable endpoint, but here the target was never in trouble — the fault is momentary packet loss on the path.

If the escalation is intended, would it be reasonable to (a) reset restart_count after a configurable healthy period, and (b) make the liveness watchdog aware of in-progress backoff so it doesn't alarm on expected silence?

### Steps to reproduce

Steps to reproduce

1. Configure BirdNET-Go with an RTSP audio source over TCP (mediamtx on a separate host in my case).
2. Ensure the network path between BirdNET-Go and the RTSP server experiences intermittent stalls longer than ffmpeg's 10s socket timeout. In my case these occur naturally on a ~1200s cycle due to an unresolved issue with a wireless hop in the path; to reproduce deliberately, briefly interrupt the connection (e.g. iptables drop rule on the BirdNET-Go host for ~15 seconds, or unplug/replug the link) so ffmpeg exits.
3. Let the stream recover, then repeat the interruption after 20+ minutes of healthy, uninterrupted streaming.
Repeat several times, leaving a long healthy interval between each interruption.
4. Watch /data/logs/audio.log:
```
grep "waiting before restart attempt" /data/logs/audio.log
grep "silence detected" /data/logs/audio.log
```

### Expected behaviour

What I expected: restart_count to reset after a sustained healthy period, so that transient faults 20 minutes apart are each treated as a first failure with a short backoff, rather than as a consecutive failure streak.

### Actual behaviour

What happens instead: the counter escalates across events separated by 1200s of healthy streaming, and only resets when the liveness watchdog intervenes.

### When did this start?

Always been present

### Relevant log output

```shell
Log evidence

Backoff escalation across widely-spaced events:

2026-08-01T19:39:25  backoff_seconds:5   restart_count:1
2026-08-01T20:39:25  backoff_seconds:10  restart_count:2
2026-08-01T20:59:26  backoff_seconds:20  restart_count:3
2026-08-01T21:59:26  backoff_seconds:40  restart_count:4
2026-08-01T22:19:25  backoff_seconds:80  restart_count:5

Liveness firing on self-imposed silence (with silenceThreshold at default):

2026-08-02T15:19:25  waiting before restart attempt  backoff_seconds:40  restart_count:4
2026-08-02T15:19:52  ERROR audio source silence detected  silence:37.178s

With silenceThreshold: 60, alerts tracked the 80s rung instead:

2026-08-01T19:00:19  silence:1m3.222s
2026-08-01T22:20:19  silence:1m3.32s
2026-08-02T05:00:19  silence:1m1.151s
```

Support dump uploaded via the in-app tool, linked to this issue. System ID: C3E0-2F3F-B2D5

RTSP source: mediamtx v1.15.2 (linux_arm64) running on a separate host — Orange Pi Zero 2W, Debian 12 Bookworm, kernel 6.1.31-sun50iw9.

The stream is produced by ffmpeg 5.1.8 capturing from a USB audio interface via ALSA and publishing to mediamtx on the same host:

`ffmpeg -f alsa -i plughw:Device,0 -ar 48000 -ac 1 -c:a libopus -b:a 128k -rtsp_transport tcp -f rtsp rtsp://localhost:8554/birdmic`

Stream consumed by BirdNET-Go: rtsp://192.168.0.25:8554/birdmic — Opus, mono, 48kHz, TCP transport.

Path between hosts: wired ethernet at both ends, but crossing a wireless mesh hop in between.

Relevant config:
```
realtime:
  rtsp:
    streams:
      - url: rtsp://192.168.0.25:8554/birdmic
        transport: tcp
        channelMode: left
    health:
      healthydatathreshold: 120
      monitoringinterval: 30
    ffmpegParameters: []
```

mediamtx logs show the teardown originating from the client side (torn down by 192.168.0.27), and mediamtx itself never restarts - its systemd unit shows continuous uptime throughout.

### Pre-submission checks

### Additional context

_No response_

tphakala:
The FFmpeg stream keeps two independent failure counters: `restartCount`, which drives the escalating restart backoff (5s/10s/20s/40s...), and `consecutiveFailures`, which drives the circuit breaker. `conditionalFailureReset` runs on every audio frame and, once the process proves stable (at least 30 seconds of uptime and at least 100KB received), reset `consecutiveFailures` but left `restartCount` untouched.

As a result a series of widely-spaced transient stalls kept escalating the backoff even though each fault recovered cleanly in between. The long backoff eventually kept the source down long enough to trip the liveness/silence watchdog, turning a minor fault into roughly 5x more downtime than the fault itself.

This resets `restartCount` under the same stability gate as the circuit-breaker reset, so genuine sustained recovery clears the accumulated backoff. Storm protection is preserved: a genuine reconnect storm never sustains the 30s + 100KB window, so its restart count keeps climbing and the escalating backoff still engages. Both stability inputs are per-process, so a flapping source must independently earn the stable window before its backoff decays.

`restartCountMu` is acquired sequentially with `circuitMu`, never nested, matching the existing locking discipline in the surrounding reset paths.

## Testing

Extended the existing `conditionalFailureReset` tests and added a deterministic table-driven regression (`TestStream_ConditionalFailureReset_RestartCount`) covering the reset, both gate-fail no-ops, the failures-already-zero independence case, and the zero-start-time no-op. `go test -race ./internal/audiocore/ffmpeg/...` passes.

Fixes #4080.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stable stream operation now resets accumulated restart backoff along with circuit-breaker failures.
  * Restart backoff is preserved when stability requirements are not met, preventing premature recovery.

* **Tests**
  * Added coverage for stable recovery, insufficient runtime or data, existing zero failures, and missing process-start times.

### DIFF
--- a/internal/audiocore/ffmpeg/stream.go
+++ b/internal/audiocore/ffmpeg/stream.go
@@ -2103,8 +2103,9 @@ func (s *Stream) maybeEngageAudioOnlyFallback() bool {
 	return true
 }
 
-// conditionalFailureReset resets failures only after the process has proven
-// stable operation with substantial data reception.
+// conditionalFailureReset resets the circuit breaker failure counter and the
+// restart-backoff counter only after the process has proven stable operation
+// with substantial data reception.
 func (s *Stream) conditionalFailureReset(totalBytesReceived int64) {
 	s.cmdMu.Lock()
 	processStartTime := s.processStartTime
@@ -2131,6 +2132,42 @@ func (s *Stream) conditionalFailureReset(totalBytesReceived int64) {
 				logger.String("operation", "conditional_failure_reset"))
 		} else {
 			s.circuitMu.Unlock()
+		}
+
+		// Reset the restart count that drives the escalating restart backoff, under
+		// the same stability gate as the circuit-breaker reset above. Without this,
+		// a series of widely-spaced transient stalls keeps escalating the backoff
+		// (5s, 10s, 20s, 40s, ...) even though each fault recovered cleanly in
+		// between, which then keeps the source down long enough to trip the
+		// liveness/silence watchdog (issue #4080). Storm protection is preserved:
+		// a genuine reconnect storm never sustains the required stable window, so
+		// its restart count keeps climbing.
+		//
+		// restartCountMu is acquired separately from circuitMu, never nested, to
+		// match the sequential (non-nested) locking used elsewhere (Restart, the
+		// silence-timeout reset, the audio-only fallback) and avoid any
+		// lock-ordering hazard. It is reset independently of consecutiveFailures
+		// because other paths can drive consecutiveFailures to zero while leaving
+		// the restart count high: the circuit-breaker cooldown (isCircuitOpen /
+		// circuitCooldownRemaining) zeroes consecutiveFailures without touching
+		// restartCount, and a normal process exit calls resetFailures (zeroing
+		// consecutiveFailures) and then still runs handleRestartBackoff (which
+		// increments restartCount).
+		s.restartCountMu.Lock()
+		if s.restartCount > 0 {
+			previousRestarts := s.restartCount
+			s.restartCount = 0
+			s.restartCountMu.Unlock()
+
+			getStreamLogger().Info("resetting restart count after successful stable operation",
+				logger.String("url", s.config.safeURL()),
+				logger.Float64("runtime_seconds", timeSinceStart.Seconds()),
+				logger.Int64("total_bytes", totalBytesReceived),
+				logger.Int("previous_restarts", previousRestarts),
+				logger.String("component", "ffmpeg-stream"),
+				logger.String("operation", "conditional_failure_reset"))
+		} else {
+			s.restartCountMu.Unlock()
 		}
 	}
 }
--- a/internal/audiocore/ffmpeg/stream_test.go
+++ b/internal/audiocore/ffmpeg/stream_test.go
@@ -154,6 +154,18 @@ func (s *Stream) getTotalBytesReceivedForTest() int64 {
 	return s.totalBytesReceived
 }
 
+func (s *Stream) getRestartCountForTest() int {
+	s.restartCountMu.Lock()
+	defer s.restartCountMu.Unlock()
+	return s.restartCount
+}
+
+func (s *Stream) setRestartCountForTest(count int) {
+	s.restartCountMu.Lock()
+	defer s.restartCountMu.Unlock()
+	s.restartCount = count
+}
+
 func (s *Stream) getStreamCreatedAtForTest() time.Time {
 	s.streamCreatedAtMu.RLock()
 	defer s.streamCreatedAtMu.RUnlock()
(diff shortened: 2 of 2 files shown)


tphakala:
@coderabbitai review

Issue #4059, opened 2026-07-28
bug: storage space becomes full, auto delete doesn’t run


dand11:
### BirdNET-Go version / build date

Version: 20260716 Build: 2026-07-16T19:54:20Z OS: Debian Linux Architecture: aarch64 Hardware: Raspberry Pi 5 Model B Rev 1.0 Environment: Docker Browser: Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.6 Mobile/15E148 Safari/604.1

### Install method

Docker via install.sh (recommended installer script)

### Device type

Raspberry Pi 5

### Operating system and version

Raspberry Pi OS (latest)

### Audio source

USB or built-in sound card (local capture device)

### Hardware details

8GB RAM
128GB SD card
384kHz USB microphone for bats

### Description

My setup is set to detect and record bats, with a 384kHz microphone. This may or may not be relevant to the issue.

Storage is set to the default. To keep disk space <80%, but storage space fills up without anything being deleted. Currently sitting at 98%.

I detect a *lot* of bats.

### Steps to reproduce

Use a 384kHz microphone and detect a lot of bats.
Wait.
Storage space will reach 100% in a few days.

### Expected behaviour

Storage is constantly trimmed to 80%

### Actual behaviour

It gets up to 100%

### When did this start?

First time setting up BirdNET-Go (never worked)

### Relevant log output

```shell

```

### Pre-submission checks

- [x] I am running the [latest release](https://github.com/tphakala/birdnet-go/releases) or have tested with the `:dev` container image.
- [x] I have searched existing open and closed issues and this does not appear to be a duplicate.

dand11:
On a pi5, is there any reasonable disadvantage to changing the max files to delete to 10000, and the time between deletions to 10ms?

tphakala:
File delete limit and delay between deletions is to avoid slow SD cards from stalling whole system, so it is not matter of RPi model and performance but more a matter of SD card write performance. You could try and see how it works on your system.

dand11:
Cool, thanks.

I've just realised that the 1000 file limit is a compile-time thing rather than run-time, so I can't change it without recompiling I don't think.

I think, in combination with the fact that bat recordings are stored as wav instead of m4a like birds, I might have to pause bat recording for now!

tphakala:
[...]

Issue #3388, opened 2026-06-05
bug: Detection Patterns by Time of Day shows flat line at zero despite having detection data


prestoy:
### BirdNET-Go version / build date

nightly-20260601

### Install method

Docker via install.sh (recommended installer script)

### Device type

Raspberry Pi 4

### Operating system and version

Raspberry Pi OS Bookworm (Debian 12), 64-bit

### Audio source

Network audio stream - RTSP / RTSPS (e.g. IP camera)

### Hardware details

- RAM: 4GB
- Storage: SD-mmc, 64G :  12G used   44G free  22% use
- Audio device: none, sound comes from RTSP-stream which works fine.

### Description

The "Detection Patterns by Time of Day" chart in the Analytics section shows a flat horizontal line at zero for all hours, despite having several days of detection data in the database.

I have four full days of detection data. Other analytics are working (Species Detection Trends, Species Diversity Over Time)

<img width="1009" height="1070" alt="Image" src="https://github.com/user-attachments/assets/409ce16c-9789-4f52-98d9-c34ce06c0f20" />

### Steps to reproduce

1. Install BirdNET-Go via install.sh (Docker)
2. Allow detections to accumulate over several days
3. Navigate to Analytics → Detection Patterns by Time of Day
4. Observe flat line at zero

### Expected behaviour

The chart should display detection counts aggregated by hour of day across the selected time period.

### Actual behaviour

The chart renders but shows a flat line at zero for all 24 hours, regardless of the selected time period.

### When did this start?

First time setting up BirdNET-Go (never worked)

### Relevant log output

```shell
journalctl -u birdnet-go --since "1 hour ago" | grep -i "analytic\|chart\|time\|error\|warn"
Jun 05 12:43:34 passer-e33 docker[36734]: INFO  [analysis.processor] discarding detection species=storspove source="RTSP Stream" best_model_id=BirdNET_V2.4 total_count=1 model_count=1 reason="false positive, matched 1/2 times" operation=discard_detection
Jun 05 12:43:47 passer-e33 docker[36734]: INFO  [weather] Successfully fetched weather data provider=yrno time="2026-06-05 12:00:00+02:00" temp_c=19.9 wind_mps=2.8 humidity_pct=63 pressure_hpa=996 description=cloudy city=
[...]

PR #4172, opened 2026-08-23
refactor: batch low-severity tech-debt cleanups (byte-format + log-capture dedup, hourly self-check parity)

tphakala:
## Summary

Four low-severity tech-debt items folded into one change. Everything is behavior-preserving except a single added WARN diagnostic. No public API changes (one exported helper is retained as a delegator), no config or schema changes.

**1. Hourly by-hour self-check parity.** `GetHourlyDistribution` already emits an invariant WARN when detections have an unparseable time column that collapses the by-hour chart (GitHub #3388). Its sibling `GetHourlyAnalyticsData` grouped by the same hour expression with no such guard. `warnIfHoursUnparseable` is generalized to take a filter closure and an operation name so both aggregations share one implementation, and the check is extended to `GetHourlyAnalyticsData`. Both call sites now gate the extra COUNT on an hour-0 bucket being present: an unparseable time makes the hour expression evaluate to SQL NULL, which scans into `Hour == 0`, so the absence of an hour-0 bucket proves there is nothing to count and the diagnostic COUNT is skipped.

**2. Byte-formatter dedup.** Three identical KMGTPE byte-formatter copies existed (`apicore.FormatBytesUint64`, `diskmanager.humanizeBytes`, and an `imports` package `formatBytes`). They are replaced by a new dependency-free leaf package `internal/formatutil` (`Bytes`, `BytesUint64`). `apicore.FormatBytesUint64` is kept as a one-line delegator so its external callers are untouched; the two unexported copies are removed and their call sites rewired. Output is byte-identical for every input range.

**3. Log-capture helper dedup.** Seven packages each carried a near-identical helper that swaps the process-global logger for a buffer and restores it on cleanup. They are consolidated into a new importable test-support package `internal/logger/logtest` (`Capture`, `CaptureBuffer`, `CaptureBufferAt`), following the existing `conftest`/`apitest` pattern. `CaptureBufferAt` preserves the one Info-level capture (the birdweather attribution test) that must not be widened to Debug, so that test's support-dump regression guard stays meaningful.

## Related Issues

Extends the unparseable-time diagnostic from GitHub #3388 to the sibling analytics query. No functional change to the retention-cleanup diagnostics from #4059.

## Test Plan

- [x] `golangci-lint run` clean on the full project (0 issues)
- [x] `go test -race` green for every touched package (formatutil, logtest, datastore, diskmanager, apicore, imports, api/middleware, birdweather, analysis/processor, weather)
- [x] New tests: `formatutil` boundary/negative/max cases; `logtest` capture, level-filtering, and global-restore; `GetHourlyAnalyticsData` unparseable-time self-check including the midnight-only (hour-0 bucket, no NULL rows) and per-filter scoping cases
- [x] Byte-formatter output verified identical to the three removed copies across int64/uint64 ranges

## Summary by CodeRabbit

* **New Features**
  * Added consistent formatting for file sizes and storage amounts across the application, including support for very large values and negative amounts.

* **Bug Fixes**
  * Improved hourly analytics diagnostics by detecting and reporting unparseable timestamps more consistently.
  * Ensured diagnostic checks respect the selected analytics filters.

* **Refactor**
  * Standardized logging behavior and test coverage across analytics, imports, storage, and weather integrations.

### DIFF
--- a/internal/analysis/processor/export_log_guards_test.go
+++ b/internal/analysis/processor/export_log_guards_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/tphakala/birdnet-go/internal/conf"
+	"github.com/tphakala/birdnet-go/internal/logger/logtest"
 )
 
 // fallbackTestTime is a fixed timestamp so the generated clip names are
@@ -103,7 +104,7 @@ func TestResampleKeyDistinguishesRatePairs(t *testing.T) {
 // enabled with a target audionorm cannot accept, which is the branch that used
 // to emit on every single detection of an affected install.
 func TestNormalizeSkipLogsOncePerFormat(t *testing.T) {
-	logs := captureExportLogs(t)
+	logs := logtest.CaptureBuffer(t)
 	resetExportLogGuards(t)
 
 	a := newExportLogAction(t, t.TempDir(), "clip.mp3", "mp3")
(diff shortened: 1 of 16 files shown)


tphakala:
@coderabbitai full review

PR #4212, opened 2026-08-28
fix(classifier): preserve privacy labels beyond top-k

### LINKED ISSUE #4177: bug: privacy filter misses human speech dropped by top-10 prediction truncation
## Summary

The privacy filter only ever sees the top 10 predictions per audio chunk, so low-confidence human vocalizations are discarded before the filter can act on them. This is why human speech leaks into saved recordings even when `realtime.privacyFilter.confidence` is lowered to 5-10%, as reported in discussion #4174.

## Root cause

Every model truncates its prediction list to `defaultTopKResults = 10` (`internal/classifier/tracing.go:61`) before results reach the processor:

- `internal/classifier/analyze.go:99` (primary BirdNET v2.4 / TFLite path)
- `internal/classifier/perch_onnx.go:265` (Perch v2)
- `internal/classifier/birdnet_v3_onnx.go:264` (BirdNET v3.0)
- `internal/classifier/bat_onnx.go:322` (Bat)
- the ONNX classifier also defaults to `topK: 10` (`internal/inference/onnx/classifier.go:36`)

The privacy filter records a human hit in `handleHumanDetection` (`internal/analysis/processor/processor.go:1319`), which runs only over the results the classifier returned, i.e. those 10 survivors. BirdNET scores human speech at roughly 1-10% confidence, so in any chunk where 10 or more bird/other classes outrank the human class, the "Human vocal" / "Speech" prediction is cut at the top-K step and never reaches `handleHumanDetection`. `LastHumanDetection[source]` is never set, and `shouldDiscardDetection` (`processor.go:1498`) therefore keeps the concurrent bird detection.

## Why lowering the confidence setting does not help

`handleHumanDetection` gates on `result.Confidence > settings.Realtime.PrivacyFilter.Confidence` (`processor.go:1321`), but that check runs *after* the top-10 truncation. Dropping `privacyFilter.confidence` to 5-10% only matters if the human class is still present in the result set, and by that point it usually is not. The signal the filter depends on was discarded one step upstream, in the classifier.

## Reproduction

1. Enable the privacy filter, set `realtime.privacyFilter.confidence` to 0.05.
2. Play a recording with active birdsong (several species above ~10% confidence) and a person talking over it.
3. Observe that bird detections from the speech interval are still saved. The human vocalization does not appear in logs because it never survives top-K.

## Expected

When a human vocalization is present in a chunk above the privacy confidence threshold, the privacy filter should suppress detections from that chunk, regardless of how many bird classes outrank the human class in that chunk.

## Suggested fixes

- **Preserve non-bird / human classes past top-K.** When trimming to `defaultTopKResults`, always retain any prediction the privacy and dog-bark filters care about (`isHumanVocalization` / `isDogDetection`), so the filter sees them even when they rank below the top 10. This is the smallest targeted fix.
- **Or evaluate the human/dog gate inside the classifier before truncation**, passing a "human present" signal alongside the trimmed results.
- **Longer term**, a dedicated VAD removes the dependency on the bird model ranking speech at all (see #4176).

Note there is also a threshold inconsistency worth cleaning up while here: the save-suppression path uses `result.Confidence > baseThreshold` (`processor.go:1014`, the global species threshold, ~15-30%) while the cross-detection record uses `PrivacyFilter.Confidence` (`processor.go:1321`, ~5-10%). Same feature, two different knobs.

## Related

- Discussion #4174 (original report)
- #4176 (dedicated Silero VAD classifier, the robust long-term fix)
- #2661 (privacy filter observability, would make this measurable)

Oxygen56:
Fixes #4177

## Summary

- retain the strongest human and dog signals even when they rank below the normal top-K result limit
- carry those filter-only signals separately so they cannot become stored additional predictions
- share the existing locale-safe BirdNET and Perch label recognition between ranking and processing
- add regression coverage for low-ranked localized BirdNET and Perch labels and the persistence boundary

## Root Cause

The classifier copied only the ten highest-confidence results before handing a prediction batch to the processor. Human speech can score above the configured privacy threshold while still ranking below those ten results, so the label was removed before the privacy filter could inspect it. Simply appending every matching label would also expand the stored additional-prediction set, so the filter signals need a separate path after ranking.

## Verification

- Focused baseline-versus-patched probe: passed; the pre-fix path loses lower-ranked filter signals, while the patched path retains only the strongest missing human and dog signals and keeps both out of stored additional predictions.
- Locale and collision checks: passed for BirdNET, Perch, mixed casing, and bird names containing similar substrings.
- The focused offline regression also passed with race detection; formatting and test-discovery checks passed.
- The repository's Go package tests, lint, cross-platform builds, and container checks will run in GitHub Actions because the local module cache does not contain the newly locked archives or native test dependencies.

## Preflight Status

- Reviewed reuse and performance, correctness and safety, project conventions, localization impact, queue ownership, storage behavior, integration wiring, and backward compatibility.
- No unresolved findings remain in the changed code.

## Compatibility

The existing top-K order, result-count reporting, and ownership guarantees are unchanged. At most one strongest missing signal per filter family is carried internally, while the queued normal result and stored additional-prediction sets remain capped at ten. Configuration, public APIs, and storage schemas are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Privacy and dog-bark filters now detect relevant lower-ranked audio signals, even outside the primary prediction results.
  * Human vocalization and dog-sound labels are recognized more consistently across supported label formats.

* **Bug Fixes**
  * Filter-only signals no longer appear as additional saved predictions or inflate ranked prediction metrics.
  * Improved separation between filter detections and primary prediction results, providing cleaner detection records and reporting.

### DIFF
--- a/internal/analysis/process.go
+++ b/internal/analysis/process.go
@@ -309,10 +309,15 @@ func ProcessData(ctx context.Context, bn *classifier.Orchestrator, bufMgr *buffe
 			Build()
 	}
 
+	// Keep lower-ranked human/dog labels available to the filters without mixing
+	// them into the normal top-ranked results used by metrics, diagnostics, and
+	// additional-prediction persistence.
+	topResults, filterSignals := classifier.SplitFilterSignals(results)
+
 	log.Debug("ProcessData inference complete",
 		logger.String("source", source),
 		logger.String("model_id", modelID),
-		logger.Int("result_count", len(results)),
+		logger.Int("result_count", len(topResults)),
 		logger.Duration("inference_duration", inferenceDuration),
 		logger.Int("sample_bytes", len(data)))
 
@@ -321,14 +326,14 @@ func ProcessData(ctx context.Context, bn *classifier.Orchestrator, bufMgr *buffe
 
 	// Record result count metric
 	if pm != nil {
-		pm.RecordBirdNETResults(source, len(results))
+		pm.RecordBirdNETResults(source, len(topResults))
 	}
 
 	// DEBUG print all BirdNET results
 	if conf.Setting().BirdNET.Debug {
 		debugThreshold := float32(0) // set to 0 for now, maybe add a config option later
 		hasHighConfidenceResults := false
-		for _, result := range results {
+		for _, result := range topResults {
 			if result.Confidence > debugThreshold {
 				hasHighConfidenceResults = true
 				break
@@ -338,7 +343,7 @@ func ProcessData(ctx context.Context, bn *classifier.Orchestrator, bufMgr *buffe
 		if hasHighConfidenceResults {
 			log.Debug("birdnet results",
 				logger.String("source", source))
-			for _, result := range results {
+			for _, result := range topResults {
 				if result.Confidence > debugThreshold {
 					log.Debug("birdnet result",
 						logger.Float64("confidence", float64(result.Confidence)),
@@ -394,7 +399,8 @@ func ProcessData(ctx context.Context, bn *classifier.Orchestrator, bufMgr *buffe
 		AudioCapturedAt: audioCapturedAt,
 		ElapsedTime:     elapsedTime,
 		PCMdata:         pcmCopy,
-		Results:         results,
+		Results:         topResults,
+		FilterSignals:   filterSignals,
 		Source:          audioSource,
 		ModelID:         modelID,
 	}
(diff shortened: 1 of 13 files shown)


tphakala:
Hi @Oxygen56, a quick licensing note before this gets merged.

tphakala:

tphakala:
Closing this due no response to relicensing request.

Issue #2702, opened 2026-04-09
Live spectrogram waterfall is blank on Safari/WebKit; propose server-side FFT streaming for all browsers plus x- and y-axes


davehaas:
## Issue Type

  - [x] Bug report
  - [x] Feature request
  - [ ] Question / Support

  ## Description

  On Safari/WebKit, the Live Audio spectrogram waterfall and Live Spectrogram widget waterfall both stay blank, even though live HLS audio playback works. Chrome renders the live waterfall correctly. Bird IDs are printed, even though there's no spectral waterfall being shown.

  The current live spectrogram path depends on browser-side Web Audio analysis of an HLS-backed HTMLAudioElement:

  HLS audio -> HTMLAudioElement -> createMediaElementSource() -> AnalyserNode -> canvas waterfall

  In local testing, Safari/WebKit returned all-zero FFT data from analyser.getByteFrequencyData(...) while audio playback itself was working.

  I prototyped a local fix that replaces the browser-analyser dependency with a server-side live FFT stream for the live spectrogram only. This works on  macOS Safari, iOS Safari, and Chrome, while keeping HLS for audible playback.

  This issue is about the live spectrogram path only. It does not cover static/recent-detection spectrogram generation or clip playback.

  ## If RTSP issue, please provide exact make and model of your RTSP source device (camera etc)

  Not applicable.

  ## Steps to Reproduce (for bugs)

  1. Open BirdNET-Go in Safari/WebKit on macOS, iOS, iPadOS.
  2. Open the live spectrogram widget or full live spectrogram page.
  3. Start/observe live audio playback.
  4. Compare Safari/WebKit behavior to Chrome on the same instance.

  ## Expected behavior

  The live spectrogram waterfall should render in Safari/WebKit the same way it does in Chrome.

  ## Actual behavior

  On Safari/WebKit:

  - live HLS audio playback works
  - static spectrograms work
  - live spectrogram overlays/labels still appear
  - but the live waterfall itself remains blank because browser-side FFT analysis yields zero data

  ## Screenshots

  I can provide before/after screenshots or a short screen recording if helpful.

  ## Environment

  ### Version Information (build date)

  Bug first found on a Kubuntu-based '[nightly-20260322]' Dockerized instance. Replicated after I installed a test version built from latest repo clone and tested against the local source-build prototype branch based on the current upstream code, rather than an official release tarball.

  ### System Information

  - Device:
[...]

PR #4217, opened 2026-08-29
fix(spectrogram): render the live waterfall on Safari/WebKit

alexbelgium:
## Summary

The live spectrogram waterfall is blank on Safari/WebKit (macOS and iOS) while audio playback works — reported in #2702.

Root cause is [WebKit bug 180696](https://bugs.webkit.org/show_bug.cgi?id=180696), open since 2017: `createMediaElementSource()` on an HLS-backed `<audio>` element yields silence, so `AnalyserNode.getByteFrequencyData()` returns all zeros. There is no client-side workaround — the samples never reach the Web Audio graph.

So the bins have to come from the server. Rather than add a dedicated spectrogram endpoint, this rides three things that already exist: the PCM consumer that already computes audio levels, the SSE stream the spectrogram pages already open, and `SpectrogramCanvas`'s existing `frequencyData` buffer. No new endpoint, no new lifecycle manager, no config surface.

**Browsers with a working `AnalyserNode` are untouched** — they never open the extra connection and never receive spectrum data.

## Changes

**Server — magnitude columns beside the existing audio level**

- `internal/analysis/audio_level_spectrum.go` (new): a rolling 1024-sample Hann window and a package-local radix-2 FFT producing 512 bins, scaled over the `AnalyserNode` defaults (−100/−30 dB), rate-limited to one column per 50 ms per source — the same rate the SSE stream already enforces, so no column is computed that would be dropped.
- `AudioLevelConsumer.Write` attaches the column to the value it already publishes. `AudioLevelData` gains `Spectrum`, `SpectrumSampleRate` and `SpectrumTime`.
- Columns are computed unconditionally rather than on subscriber demand. `BenchmarkSpectrumAnalyzer` (included) measures **~42 µs/column on an Intel N150** — under 0.1% of a core per source at 20 columns/s, and a fraction of a percent on a Pi. Demand plumbing across the analysis/API boundary would cost more than it saves.

**API — opt-in, off by default**

- `internal/api/v2/audio/audio_level_spectrum.go` (new): `?spectrum=1` for every source, or `?spectrum=<sourceID>` for one. Absent by default, so level-meter clients (the bulk of this endpoint's traffic) get a byte-identical payload. A column is ~700 bytes per source per update, which is exactly why it is not unconditional.
- The spectrum is also stripped from the long-lived `latestLevels` health-check copy, so it cannot reach a client that never asked.

**Frontend — adopt the server bins only when the local analyser is proven blind**

- `frontend/src/lib/utils/serverSpectrumFallback.ts` (new) owns the whole workaround. It probes the analyser every 250 ms; only after four consecutive all-zero reads **while audio is playing through a running context** does it open the spectrum stream, and it adopts the server bins only once a column shows energy the analyser cannot see. A single non-zero analyser read tears the fallback down permanently. No UA sniffing — the decision is made from observed behaviour.
- Columns carry their capture time, are queued, and are rendered against the playhead wall clock via the existing `computeWallClockAtPlayhead()` — the same helper and clock the detection-label promotion already uses. Without this the waterfall would run seconds ahead of the buffered audio, and ahead of the labels drawn over it.
- If the stream drops, the waterfall blanks after 1.5 s rather than smearing one frozen column across the canvas. If the playhead never reaches the queue (a browser and server whose clocks disagree, which the seekable-range estimate cannot detect), alignment is abandoned with a warning rather than leaving the user blank.
- `SpectrogramCanvas` gains one boolean prop, `externalFrequencyData`; the render path is otherwise unchanged. The server bins match the browser analyser's shape (1024-point FFT, 512 bins, same dB range), so `pixelToBinMap` works as-is.

## Testing

- [x] Go tests pass — 675 across `internal/analysis`, `internal/api/v2/audio`, `internal/audiocore`, including 16 new ones (`-race` on the new suites)
- [x] Frontend tests pass — 2910/2910, including 27 new unit tests for the decode, queue-trim, column-selection and render-decision logic
- [x] Linting passes — `golangci-lint run` clean on the changed packages; frontend `format:check`, `lint`, `lint:css`, `typecheck` (0 errors), `ast:all`, `analyze:circular` and `build` all clean
- [x] Benchmark included for the per-column FFT cost
- [ ] Manual testing on a real Safari/iOS device — **not yet done, see below**

### What I could not verify

I do not have a WebKit device in this environment, so two things are reasoned about rather than observed, and I would value a second pair of eyes (or hands) on them:

1. **That Safari actually latches the fallback.** The probe logic follows from the bug report's description (analyser returns zeros while playback works), but I have not watched it happen.
2. **Clock agreement on native iOS HLS.** With hls.js, alignment uses `playingDate` (stream time) and is exact. Native HLS has no `playingDate`, so the playhead is derived from the browser clock while columns carry the server clock. Badly disagreeing clocks degrade to unaligned rendering — handled deliberately, but unconfirmed in the field.

## Notes for review

Scoped deliberately to this one bug, after the feedback on #2745. No axis labels, no colormap persistence, no unrelated refactors. Most of the code is in five new files; existing files are touched in ~155 lines across 25 hunks, mostly single lines.

`internal/api/v2/audio/audio_level.go` has one change beyond the opt-in wiring: `updateAudioLevel` now takes `*audiocore.AudioLevelData`. The three new fields pushed the struct past gocritic's 80-byte `hugeParam` threshold, and copying it per update per client was pointless anyway.

The FFT is a small package-local copy rather than a shared helper. The only other FFT in the tree is private to `internal/audiocore/ultrasonic`, and I did not want the live audio-level path depending on a bat-detection package. Extracting a common DSP helper felt like a separate cleanup rather than something to smuggle into this PR — happy to do it either way.

## Related Issues

Refs #2702

## Summary by CodeRabbit

* **New Features**
  * Live spectrograms now remain available for HLS playback in browsers where local audio analysis is unavailable.
  * Spectrogram data can be requested through the audio-level stream for all sources or a specific source.
  * Server-provided spectrum data stays synchronized with playback timing and handles stalls and recovery gracefully.

* **Documentation**
  * Added audio-level stream documentation for requesting spectrum data.

* **Tests**
  * Expanded coverage for spectrum generation, streaming, rendering, alignment, and fallback behavior.

### DIFF
--- a/frontend/src/lib/desktop/components/media/SpectrogramCanvas.svelte
+++ b/frontend/src/lib/desktop/components/media/SpectrogramCanvas.svelte
@@ -31,6 +31,12 @@
     analyser: AnalyserNode | null;
     /** Pre-allocated frequency data buffer */
     frequencyData: Uint8Array<ArrayBuffer>;
+    /**
+     * Render `frequencyData` as supplied instead of polling the analyser.
+     * Set when the bins come from the server because the browser's
+     * AnalyserNode cannot read HLS-backed media (Safari/WebKit).
+     */
+    externalFrequencyData?: boolean;
     /** Sample rate for bin-to-frequency mapping */
     sampleRate: number;
     /** FFT size for bin count calculation */
@@ -64,6 +70,7 @@
   let {
     analyser,
     frequencyData,
+    externalFrequencyData = false,
     sampleRate,
     fftSize,
     frequencyRange = [0, 15000],
@@ -203,7 +210,7 @@
 
   // Main animation loop
   $effect(() => {
-    if (!analyser || !isActive || !canvasEl) return;
+    if ((!analyser && !externalFrequencyData) || !isActive || !canvasEl) return;
 
     const ctx = canvasEl.getContext('2d');
     if (!ctx) return;
@@ -266,8 +273,11 @@
       const deltaTime = (now - lastFrameTime) / 1000;
       lastFrameTime = now;
 
-      // Read frequency data from analyser
-      analyser.getByteFrequencyData(frequencyData);
+      // Read frequency data from the analyser, unless the caller is filling
+      // frequencyData itself from the server.
+      if (!externalFrequencyData) {
+        analyser?.getByteFrequencyData(frequencyData);
+      }
 
       // Compute device pixels to scroll; cap deltaTime to prevent OOM after tab resume
       const clampedDelta = Math.min(deltaTime, MAX_FRAME_DELTA_SEC);
--- a/frontend/src/lib/desktop/features/dashboard/components/MiniSpectrogram.svelte
+++ b/frontend/src/lib/desktop/features/dashboard/components/MiniSpectrogram.svelte
@@ -97,7 +97,11 @@
   const MAX_OVERLAY_SLOTS = 4;
 
   // Initialize composable during component init (must be at top level for $effect cleanup)
-  const spectro = useSpectrogramAnalyser({ fftSize: FFT_SIZE, audioOutput: false });
+  const spectro = useSpectrogramAnalyser({
+    fftSize: FFT_SIZE,
+    audioOutput: false,
+    getPlayingDate: () => hls?.playingDate ?? null,
+  });
 
   function shouldAutoStart(): boolean {
     if (appState.liveSpectrogram) return true;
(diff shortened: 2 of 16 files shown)


alexbelgium:
Sorry this was a minimalistic revision of https://github.com/tphakala/birdnet-go/pull/2745 that was not intend to replace the initial pr

PR #4238, opened 2026-09-01
fix(inference): probe OpenVINO devices in a child process so a driver-stack crash falls back to ORT

### LINKED ISSUE #4236: bug: SIGABRT (glibc "free(): invalid next size") in ovbind_get_available_devices on first OpenVINO GPU enablement, native binary in LXC — later became unreproducible
## Summary

On first enabling the OpenVINO backend with an Intel Gen12 iGPU (native Linux
binary, release 20260823, unprivileged Proxmox LXC), the process aborted
reliably with glibc heap-corruption detection inside
`ov_core_get_available_devices`, crash-looping the service. The next day, with
no library or binary changes, the crash became unreproducible and the backend
now runs both BirdNET v2.4 (fp32-dfttrunc) and Perch v2 (no-dft-fp32) on the
GPU without issues. I am filing this for the signature and forensics rather
than a repro, since GPU-on-native-binary only shipped in this release (#3961)
and other early adopters may hit the same thing.

## Environment

- BirdNET-Go 20260823 (release tarball `birdnet-go-linux-amd64`), native
  binary, not Docker
- Debian 13 (trixie), unprivileged LXC on Proxmox VE, host CPU i5-12500T
  (Gen12 UHD 770)
- GPU passthrough: `dev0: /dev/dri/renderD128,gid=<render gid>` — device
  visible and root-accessible in the container
- Intel NEO 26.22.38646.4 + IGC 2.36.3+21719 + gmmlib 22.10.0 (GitHub debs;
  `intel-opencl-icd` no longer exists in Debian 13)
- OpenVINO 2026.2.0.21903.52ddc073857 runtime libs (the exact build your
  Dockerfile pins, sha256-verified) copied to `/usr/local/lib`, `ldconfig` run
- `clinfo` in the same container enumerates the Intel OpenCL platform and GPU
  correctly
- Config: `birdnet.backend: openvino` (also crashed with `auto`),
  `openvinodevice: auto`

## Crash signature

At the time of the crashes the container had 4 cores / 4 GB RAM,
`birdnet.threads: 0`, six live RTSP streams, and the crash-triggering call
path was a hot model reload (installing the `fp32-dfttrunc` variant through
`POST /api/v2/models/install/birdnet-v2.4` while the builtin TFLite model was
analyzing). After that install, every subsequent service start also crashed
at startup on the same path, until the backend was set to `onnx`.

```text
free(): invalid next size (fast)
SIGABRT: abort
PC=0x769dd04a095c m=9 sigcode=18446744073709551610
signal arrived during cgo execution
goroutine 332 [syscall, locked to thread]:
runtime.cgocall(0x17be1f9, 0x21ebdabe6ab0)
github.com/tphakala/birdnet-go/internal/inference/openvino._Cfunc_ovbind_get_available_devices(0x769d780282e0, 0x21ebd6bff290)
github.com/tphakala/birdnet-go/internal/inference/openvino.AvailableDevices()
        internal/inference/openvino/backend_openvino.go:817
github.com/tphakala/birdnet-go/internal/inference.OpenVINOHasDevice({0x1b46d06, 0x3})
        internal/inference/openvino.go:90
github.com/tphakala/birdnet-go/internal/classifier.openVINOGPUAvailable(...)
        internal/classifier/model_openvino.go:147
github.com/tphakala/birdnet-go/internal/classifier.openVINOPlanFor(...)
        internal/classifier/model_openvino.go:80
github.com/tphakala/birdnet-go/internal/classifier.(*BirdNET).openVINOPlan(...)
        internal/classifier/model_openvino.go:160
github.com/tphakala/birdnet-go/internal/classifier.(*BirdNET).initializeModel(...)
        internal/classifier/birdnet.go:325
github.com/tphakala/birdnet-go/internal/classifier.(*BirdNET).reloadModelInternal(...)
        internal/classifier/birdnet.go:1499
github.com/tphakala/birdnet-go/internal/classifier.(*ModelManager).replacePrimaryVariant.(*Orchestrator).ReloadPrimaryForVariantSwap.func2(...)
```

`free(): invalid next size (fast)` is glibc detecting corrupted heap chunk
metadata during a free performed inside the OpenVINO call, i.e. the heap was
overflowed at some earlier point and detected here.

## What I ruled out the next day (all on the same container, same libs)

1. A minimal C program doing exactly what `ovbind_load` + `AvailableDevices`
   do — `dlopen("libopenvino_c.so", RTLD_NOW | RTLD_GLOBAL)`,
   `ov_core_create`, `ov_core_get_available_devices`,
   `ov_available_devices_free` — runs clean, returns `CPU GPU`.
2. The same repro under valgrind: 0 errors through the whole NEO/IGC device
   enumeration.
3. An enriched repro that first loads `libtensorflowlite_c.so`
   (RTLD_GLOBAL, mirroring the DT_NEEDED link) and `libonnxruntime.so`
   (RTLD_LAZY, as yalue/onnxruntime_go does), then enumerates in a loop with
   interleaved heap churn and `MALLOC_CHECK_=3`: clean.
4. Dynamic-symbol overlap between the tflite/onnxruntime exports and the
   imports of `libopenvino.so` / `libopenvino_intel_gpu_plugin.so`: zero
   overlapping symbols, so the classic two-runtimes ODR clash is not it.
5. The actual release binary, `backend: openvino`: initialized the GPU
   ("OpenVINO model initialized ... device=GPU") and has now been running
   both models on GPU with no errors.

## What changed between "crashes reliably" and "cannot reproduce"

Confounders I cannot separate, listed in case one of them rings a bell:

- LXC cores 4 -> 6 (`pct set --cores`, live cpuset change) and RAM 4G -> 6G
- `birdnet.threads` 0 -> 4 (set to silence the onnxruntime
  `pthread_setaffinity_np` cpuset warnings, error code 22, which were present
  in the crashing runs)
- NEO shader/kernel cache went from cold to warm (first successful GPU model
  init took 2.27 s, subsequent ones ~360 ms)
- The crashing runs went through the hot-reload/variant-swap path first; the
  working runs initialized OpenVINO at startup

The cpuset + `threads: 0` combination seems the most suspicious to me given
the affinity errors in the same logs, but I have no proof.

## Ask

Nothing urgent since it self-resolved, but: if the enumeration probe could be
made crash-proof (e.g. run once in a short-lived child process, or a clearer
fallback when the GPU stack aborts), a driver-level abort would degrade to
"OpenVINO unavailable, using ORT" instead of a systemd crash-loop, which
matches the fallback philosophy described in backend_openvino.go. Happy to
provide the full journal, container config, or test patches on this
environment (it is the same iGPU generation your wiki benchmarks).

beaglemoo:
## Summary

Fixes #4236.

`ov_core_get_available_devices` walks the vendor driver stack (NEO, IGC, Level Zero), and a fault anywhere in that stack aborts the process before Go can recover: glibc heap-corruption detection raises SIGABRT during cgo execution, which crash-looped the whole analyzer under systemd on first GPU enablement of the native binary. This PR runs the first device enumeration in a short-lived child process, so that worst case degrades to the ordinary "OpenVINO unavailable, using ONNX Runtime" fallback the backend already promises for every other failure mode.

## Design

- New hidden subcommand `support openvino-probe`: loads the OpenVINO core in its own process, enumerates devices, and reports them on stdout as marker-prefixed lines (`OVPROBE_DEVICE=GPU`, terminated by `OVPROBE_OK`). Markers are required because the CLI bootstrap can interleave log lines on stdout; the completion marker means truncated output from a dying child is never mistaken for a successful empty result.
- `inference.OpenVINOProbeDevices(libraryPath)` launches this same executable as that subcommand (60s timeout), parses the markers, and caches the outcome per (process, library path). A cached failure means a crashing driver stack is not re-executed on every hot reload; a changed `openvinopath` re-probes.
- `classifier.openVINOGPUAvailable` now gates the GPU decision on the probe instead of the in-process enumeration. The in-process `InitOpenVINO` stays where it is: dlopen/core-create was not the crashing step, and it preserves the existing absent-vs-broken library log classification.
- `inference.OpenVINOHasDevice` (status/capability surfaces: system inference status, models API, hwprofile) serves from the cached probe result once the planner has probed, and keeps its previous in-process behavior before any probe has run. Note the pre-probe in-process fallback only ever enumerates when the core is already initialized, same as before this change.

## Testing

- `go test -race` (probe unit tests, tagged and untagged): child crash, missing completion marker, timeout, success/failure caching, re-probe on library path change, interleaved-log parsing. Fake children via test-binary re-invocation seams.
- `golangci-lint run --build-tags openvino` on the touched packages: 0 issues (the second commit fixes the one pre-existing `modernize` finding this surfaces in `openvino_parity_functional_test.go`, which CI does not lint because it does not use the tag).
- Happy path on the hardware from #4236 (i5-12500T Gen12 iGPU, unprivileged Proxmox LXC, Debian 13, OpenVINO 2026.2, NEO 26.22): probe child reports `CPU`,`GPU`; `BirdNET v2.4 fp32-dfttrunc` initializes on GPU (f32) and `Perch v2 no-dft` on GPU (f16), identical to the pre-patch behavior.
- Fault injection: with `abort()` compiled into the enumeration shim (local test build), the probe child dies with the same SIGABRT signature as #4236 and the service logs `OpenVINO device probe failed; treating GPU as unavailable`, initializes both models on ONNX Runtime, and keeps running - previously this configuration was a systemd crash-loop.

## Notes

- The probe warning is logged once per model plan that consults it (the underlying child runs once); happy to dedupe if preferred.
- Cost is one extra process execution per library path per service lifetime; on the test hardware the child completes in well under 2s warm, ~2.5s with a cold NEO cache.

## Summary by CodeRabbit

* **New Features**
  * Added safer OpenVINO device detection using an isolated probe process.
  * Added a hidden support command for diagnosing OpenVINO device availability.

* **Bug Fixes**
  * Prevented vendor-driver crashes during hardware detection from terminating the application.
  * Automatically falls back to ONNX Runtime when OpenVINO detection fails.

* **Tests**
  * Added coverage for successful probes, crashes, timeouts, malformed output, caching, and unsupported environments.

### DIFF
--- a/cmd/support/openvino_probe.go
+++ b/cmd/support/openvino_probe.go
@@ -1,0 +1,39 @@
+package support
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/tphakala/birdnet-go/internal/conf"
+	"github.com/tphakala/birdnet-go/internal/inference"
+)
+
+// OpenVINOProbeCommand creates the hidden `support openvino-probe` subcommand.
+// It is the child half of the out-of-process OpenVINO device probe
+// (inference.OpenVINOProbeDevices): it loads the OpenVINO core in this
+// process, enumerates the available devices, and reports them on stdout as
+// marker-prefixed lines the parent parses. A crash anywhere in the vendor
+// driver stack during enumeration then kills only this child, and the parent
+// falls back to ONNX Runtime instead of crash-looping (issue #4236). The
+// command is hidden because it is an internal protocol, not a user surface.
+func OpenVINOProbeCommand(settings *conf.Settings) *cobra.Command {
+	var libraryPath string
+	cmd := &cobra.Command{
+		Use:    "openvino-probe",
+		Short:  "Internal: enumerate OpenVINO devices in an isolated process",
+		Hidden: true,
+		// Errors exit non-zero; the parent treats that (and any signal) as
+		// "no OpenVINO devices". Usage help on failure would only pollute the
+		// parsed stdout stream.
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			path := libraryPath
+			if path == "" && settings != nil {
+				path = settings.BirdNET.OpenVINOPath
+			}
+			return inference.RunOVProbeChild(path, cmd.OutOrStdout())
+		},
+	}
+	cmd.Flags().StringVar(&libraryPath, "library-path", "",
+		"Path to libopenvino_c (defaults to birdnet.openvinopath from the config)")
+	return cmd
+}
--- a/cmd/support/support.go
+++ b/cmd/support/support.go
@@ -14,6 +14,7 @@ func Command(settings *conf.Settings) *cobra.Command {
 
 	// Add subcommands here
 	supportCmd.AddCommand(CollectCommand())
+	supportCmd.AddCommand(OpenVINOProbeCommand(settings))
 
 	return supportCmd
 }
--- a/internal/classifier/model_openvino.go
+++ b/internal/classifier/model_openvino.go
@@ -3,6 +3,7 @@ package classifier
 import (
 	"os"
 	"runtime"
+	"slices"
 	"strings"
 	"time"
 
@@ -155,7 +156,20 @@ func openVINOGPUAvailable(libraryPath string) bool {
 		}
 		return false
 	}
-	return inference.OpenVINOHasDevice(inference.OVDeviceGPU)
+	// Enumerate devices in a short-lived child process rather than in-process:
+	// ov_core_get_available_devices walks the vendor driver stack, and a fault
+	// there aborts the process before Go can recover (glibc heap-corruption
+	// SIGABRT during cgo, issue #4236 - a systemd crash-loop on the first GPU
+	// enablement). A crashing probe child degrades to the ordinary ORT
+	// fallback instead. The probe result is cached per process, so this costs
+	// one child run per library path.
+	devices, err := inference.OpenVINOProbeDevices(libraryPath)
+	if err != nil {
+		GetLogger().Warn("OpenVINO device probe failed; treating GPU as unavailable",
+			logger.Error(err))
+		return false
+	}
+	return slices.Contains(devices, inference.OVDeviceGPU)
 }
 
 // isLibraryAbsent reports whether err indicates the OpenVINO shared library is
--- a/internal/inference/openvino.go
+++ b/internal/inference/openvino.go
@@ -86,7 +86,19 @@ func NewOpenVINOClassifier(modelPath string, opts OpenVINOClassifierOptions) (Cl
 // OpenVINOHasDevice reports whether the named OpenVINO device (e.g. ov.DeviceGPU)
 // is available. It returns false on any error (backend not compiled in, core not
 // initialized, or query failure), so callers can use it as a plain gate.
+//
+// When the out-of-process probe has completed (the device planner runs it before
+// any classifier is built), its cached result answers without touching the
+// in-process driver stack, so a status or capability query can never trip the
+// enumeration crash the probe exists to contain (issue #4236). Before any probe
+// has run it falls back to the in-process query, preserving prior behavior.
 func OpenVINOHasDevice(name string) bool {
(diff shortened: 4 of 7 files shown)


beaglemoo:
Withdrawing this for now.

PR #4245, opened 2026-09-02
fix(audiocore): carry per-stream RTSP transport through the pipeline (#4240)

### LINKED ISSUE #4240: RTSP config from wiki guide fails on startup: migration clears global transport, startup reads only global
## Summary

Following the RTSP configuration shown in the [BirdNET-Go-Guide](https://github.com/tphakala/birdnet-go/wiki/BirdNET-Go-Guide#configuration-file-format) produces a working-looking config that fails on the first run with an empty `-rtsp_transport` argument. The legacy `urls:` + global `transport:` format the wiki documents is migrated on load to the `streams:` format, and the migration clears the global `transport`. The startup code path reads only the global, so FFmpeg receives `-rtsp_transport ""` and never opens the stream.

## Error

```
{
  "level": "WARN",
  "msg": "FFmpeg process ended",
  "module": "audiocore",
  "url": "rtsp://192.168.42.1:7447/tx3iVcnBsBWUxCby",
  "error": "...[RTSP demuxer @ ...] Undefined constant or missing '(' in ''\n[RTSP demuxer @ ...] Unable to parse option value \"\"\n[RTSP demuxer @ ...] Error setting option rtsp_transport to value .\n[in#0 @ ...] Error opening input: Invalid argument\nError opening input file rtsp://...\nError opening input files: Invalid argument\n",
  "runtime_seconds": 0.101,
  "component": "ffmpeg-stream",
  "operation": "process_ended"
}
```

`runtime_seconds` is ~0.1 and the process exits immediately, which distinguishes this from a network or auth failure.

## Reproduction

1. Start from the wiki guide's RTSP block:

   ```yaml
   realtime:
     rtsp:
       transport: "tcp"
       urls:
         - rtsp://192.168.42.1:7447/xxxxxxxx
   ```

2. Start the container once. On load, `MigrateRTSPConfig()` runs and rewrites the file on disk to:

   ```yaml
   realtime:
     rtsp:
       streams:
         - name: Stream 1
           url: rtsp://192.168.42.1:7447/xxxxxxxx
           enabled: true
           type: rtsp
           transport: tcp
       # global transport: gone
   ```

3. Restart. FFmpeg fails with `Error setting option rtsp_transport to value .`.

The `transport: tcp` value is now present on the stream entry but absent from the global, and the global is what the startup path reads.

## Root cause

Checked against commit `5b7aa9c` (Docker image build `20260823`).

The migration and the startup path disagree about where the transport lives.

`internal/conf/migrations.go` `MigrateRTSPConfig()` copies the global `tcp` into each per-stream `Transport` (line 282), then clears the global:

```go
// Clear legacy fields
rtsp.URLs = nil
rtsp.Transport = ""
```

The doc comment on `MigrateRTSPConfig` says it "Preserves the global Transport setting for RTSP/RTMP streams." It preserves the value into per-stream entries, then deletes it from the global field the startup path reads.

`cmd/serve/serve.go:52` builds the audio engine from the global:

```go
audioEngine := engine.New(cmd.Context(), &engine.Config{
    ...
    Transport:            settings.Realtime.RTSP.Transport,
    ...
}, nil)
```

`internal/audiocore/engine/engine.go:444` passes that global to every stream:

```go
streamCfg := &ffmpeg.StreamConfig{
    ...
    Transport:        e.transport,
    ...
}
```

`internal/audiocore/source.go` `SourceConfig` has no `Transport` field, so when streams are registered in `internal/analysis/audio_pipeline_service.go` (~line 1434), the per-stream `transport: tcp` written by the migration is dropped. It is not carried into the `StreamConfig` the engine builds.

`internal/audiocore/ffmpeg/process.go:64` appends the flag unconditionally:

```go
func appendRTSPMediaArgs(args []string, transport string, audioOnly bool) []string {
    args = append(args, "-rtsp_transport", transport)
    ...
}
```

With the global cleared, `transport` is `""`, and FFmpeg fails.

The per-stream value written by the migration is read on a different path: `internal/audiocore/schedule/quiet_hours.go:253` uses `stream.Transport` when restarting a stream after quiet hours. So the initial start and the quiet-hours restart read transport from different fields.

## Why the Viper default does not save it

`internal/conf/defaults.go:274` sets a Viper default:

```go
viper.SetDefault("realtime.rtsp.transport", DefaultTransport)
```

This applies when the key is absent. After migration, `rtsp.Transport` is set to the empty string and persisted, so the key is present with value `""` and the default does not apply. The unmarshalled struct gets an empty string, not `tcp`.

## Why the docs make it worse

The [Configuration Reference](https://github.com/tphakala/birdnet-go/wiki/configuration-reference) lists:

```
| `realtime.rtsp.streams` | stream-config[] | Stream configurations |
| `realtime.rtsp.urls` | string[] | Legacy: accepts old format, migrated on load |
| `realtime.rtsp.transport` | string | Legacy: global default, migrated on load |
```

It marks `realtime.rtsp.transport` as "Legacy: global default, migrated on load," which reads as "do not use this." The startup path requires it.

It lists `realtime.rtsp.streams` as the current format with type `stream-config[]`, but never expands the sub-fields, so there is no documented schema for the per-stream `transport` value. The [RTSP Troubleshooting](https://github.com/tphakala/birdnet-go/wiki/rtsp-troubleshooting) guide examples still use the legacy `urls:` + global `transport:` form.

Net, there is no accurate, complete document for the current RTSP config. A user following the wiki guide lands on the failing path.

## Suggested fix

The smallest code fix is in `MigrateRTSPConfig`: after building the streams, keep the global instead of clearing it:

```go
// Clear legacy URLs only
rtsp.URLs = nil
// rtsp.Transport = ""  // keep; the startup path (serve.go) reads this
```

That is the one-line change that stops the migration from breaking new users on the documented config.

Two follow-ons, either of which also prevents the class of bug:

- In `cmd/serve/serve.go`, fall back to `DefaultTransport` when the global is empty:

  ```go
  transport := settings.Realtime.RTSP.Transport
  if transport == "" {
      transport = conf.DefaultTransport
  }
  ```

- Carry per-stream `Transport` through `SourceConfig` into `StreamConfig` on the add path, so the per-stream value the migration writes is the value that gets used. This also removes the initial-start vs. quiet-hours-restart inconsistency.

On the docs side:

- Update the BirdNET-Go-Guide RTSP example to the `streams:` format, or add a note that the legacy form is rewritten on load and the global `transport` must remain set.
- In the Configuration Reference, remove the "Legacy" annotation from `realtime.rtsp.transport` (or clarify that it is still read at startup), and expand the `stream-config` sub-fields.
- Update the RTSP Troubleshooting examples to the `streams:` format.

## Environment

- Image build: `20260823` (commit `5b7aa9c`)
- Source: RTSP from UniFi Protect (port 7447)
- Config source: copied from the BirdNET-Go-Guide wiki page

tphakala:
## Summary

Following the wiki's legacy RTSP config (`urls:` plus a global `transport:`) produced a config that failed on the first start with an empty `-rtsp_transport` argument, so FFmpeg refused to open the stream. `MigrateRTSPConfig` copied the global transport into each stream and then cleared the global, but the startup path read only the global, so FFmpeg received `-rtsp_transport ""`. The per-stream value the migration wrote was dropped because `audiocore.SourceConfig` had no transport field, and only the quiet-hours restart path read it, so the initial start and the restart disagreed about the transport.

This fixes the transport handling end to end:

- `MigrateRTSPConfig` keeps the global transport instead of clearing it. The startup path reads it as the engine-wide default.
- `audiocore.SourceConfig` and `audiocore.AudioSource` gain a `Transport` field, carried from the per-stream config through the registry into the FFmpeg `StreamConfig`, so the per-stream transport is honored at initial start (previously only the quiet-hours restart read it).
- `buildSourceConfigsWithModels` resolves the transport to a concrete value via a new `RTSPSettings.ResolveTransport` (per-stream else global else default), so change detection compares like with like and a non-tcp global is not masked.
- `sourceNeedsReconfigure` diffs the transport so a per-stream transport edit hot-reloads, and `SyncReconfiguredParams` writes the transport back to the registry so the change does not re-fire on every later reload. The write-back stores the raw pipeline value (empty for non-RTSP stream types) so non-RTSP streams do not enter a restart loop.
- `appendRTSPMediaArgs` falls back to the default transport when empty, so FFmpeg never receives an empty `-rtsp_transport`.
- `serve.go` resolves the engine-wide default the same way.

The wiki docs (guide RTSP example, configuration reference, RTSP troubleshooting) are updated separately since they live outside the repository.

## Testing

- `go test -race` across `internal/conf`, `internal/audiocore`, `internal/audiocore/engine`, `internal/audiocore/ffmpeg`, `internal/analysis` (all pass)
- `golangci-lint run` clean
- New tests: `RTSPSettings.ResolveTransport` truth table; migration preserves the global; empty transport falls back to the default at the FFmpeg arg layer; per-stream transport flows into `SourceConfig` (explicit and global-resolved); transport change triggers reconfigure while an unchanged transport does not; registry write-back syncs transport; a non-RTSP source stays empty through reconfigure (no restart loop)

## Related Issues

Fixes #4240

## Summary by CodeRabbit

* **New Features**
  * Added per-stream RTSP/RTMP transport selection, with stream-specific settings taking precedence over global and default values.
  * Preserved the global transport setting during configuration migration for use as the engine-wide default.

* **Bug Fixes**
  * Prevented empty RTSP transport values from being passed to media processing.
  * Transport changes now correctly trigger stream reconfiguration, while unchanged settings avoid unnecessary restarts.
  * Non-network audio sources remain unaffected by network transport defaults.

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -46,10 +46,17 @@ The "realtime" command is an alias for backward compatibility.`,
 			// depends on SunCalc and ControlChan that are only available
 			// after APIServerService.Start(), so it is set later via
 			// AudioEngine.SetScheduler().
+			// Resolve the engine-wide default transport. A migrated config can
+			// persist an empty global transport, which would otherwise reach
+			// FFmpeg as -rtsp_transport "" and fail every stream that does not
+			// carry its own per-stream transport. ResolveTransport("") falls back
+			// to the global, then to conf.DefaultTransport.
+			engineTransport := settings.Realtime.RTSP.ResolveTransport("")
+
 			audioEngine := engine.New(cmd.Context(), &engine.Config{
 				FFmpegPath:           settings.Realtime.Audio.FfmpegPath,
 				SoxPath:              settings.Realtime.Audio.SoxPath,
-				Transport:            settings.Realtime.RTSP.Transport,
+				Transport:            engineTransport,
 				FFmpegParameters:     settings.Realtime.RTSP.FFmpegParameters,
 				Debug:                settings.Debug,
 				CaptureBufferSeconds: settings.Realtime.ExtendedCapture.EffectiveCaptureBufferSeconds(settings.Realtime.Audio.Export.PreCapture),
@@ -97,11 +104,12 @@ func setupFlags(cmd *cobra.Command, settings *conf.Settings) error {
 	// Every flag default comes from the loaded settings rather than from viper, for
 	// the same reason as --locale in cmd/root.go: cobra assigns the default into the
 	// bound variable at registration time, so passing the raw viper value writes it
-	// straight back over whatever load produced. For --source, --rtsp and
-	// --rtsptransport that would resurrect the legacy keys MigrateAudioSourceConfig
-	// and MigrateRTSPConfig deliberately cleared, and the next save would write them
-	// back into config.yaml. For --listen it would undo the normalized telemetry
-	// address, leaving the endpoint with nothing to bind.
+	// straight back over whatever load produced. For --source and --rtsp that would
+	// resurrect the legacy keys MigrateAudioSourceConfig and MigrateRTSPConfig
+	// deliberately cleared, and the next save would write them back into config.yaml.
+	// (MigrateRTSPConfig no longer clears the global rtsp.transport, so --rtsptransport
+	// simply re-defaults from the preserved value.) For --listen it would undo the
+	// normalized telemetry address, leaving the endpoint with nothing to bind.
 	//
 	// The remaining flags are read from settings too, even though nothing rewrites
 	// their keys today. Since the bound field already holds the loaded value, each
--- a/internal/analysis/audio_pipeline_service.go
+++ b/internal/analysis/audio_pipeline_service.go
@@ -1106,13 +1106,33 @@ func sourceNeedsReconfigure(running *audiocore.AudioSource, desired *audiocore.S
 	// running FFmpeg, silently breaking hot-reload.
 	mediaModeChanged := conf.MediaMode(running.MediaMode).Canonical() !=
 		conf.MediaMode(desired.MediaMode).Canonical()
+	// Both sides carry the resolved concrete transport (buildSourceConfigsWithModels
+	// resolves it via RTSPSettings.ResolveTransport, and the registry stores that
+	// same value), so a direct compare is correct. It must NOT canonicalize an
+	// empty value to conf.DefaultTransport here: the true default is the engine
+	// global, which can be "udp", and hardcoding "tcp" would mask a real
+	// udp->tcp change (issue #4240 hot-reload path).
+	transportChanged := running.Transport != desired.Transport
 	return running.SampleRate != desired.SampleRate ||
 		sourceSampleRateChanged ||
 		running.BitDepth != desired.BitDepth ||
 		running.Channels != desired.Channels ||
 		channelModeChanged ||
 		mediaModeChanged ||
+		transportChanged ||
 		sourceChannelsChanged
+}
+
+// rtspStreamTransport resolves the concrete transport for an rtsp.streams entry:
+// the per-stream value if set, else the global default (via ResolveTransport),
+// and only for RTSP/RTMP types where transport applies. Resolving here keeps the
+// built SourceConfig, the registry entry, and the engine's FFmpeg args all in
+// agreement on one concrete value, so change detection compares like with like.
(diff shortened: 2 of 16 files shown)


tphakala:
Addressed the transport-constant nitpick in 6fa587205: the RTSP transport tests now use package-local `transportTCP`/`transportUDP` constants instead of raw "tcp"/"udp" literals across the pipeline, engine, and registry tests.

Issue #4206, opened 2026-08-28
bug: Perch v2 detection issue with FP32 (no-dft) (Central Europe)


Aimache:
### BirdNET-Go version / build date

20260823

### Install method

Docker run / docker-compose (manual, ghcr.io/tphakala/birdnet-go)

### Device type

Raspberry Pi 5

### Operating system and version

Raspberry Pi OS / Bookworm 64 bits

### Audio source

Both a sound card and a network stream

### Hardware details

- Raspberry Pi 5 
- RAM : 8 Gb
- Disk : Nvme 512 GB
- 1 USB microphone (Rode lava something)
- 1 RTSP stream (Clippy EM272)

### Description

After upgrading to the last nightly, I saw a change in detection, especially for the Perch V2 model.

I did investigate a little and found that Perch detection (looking at the inference view in System) is just always off. What I mean by that is that the detection is always wrong and not showing the right bird, even with great sound quality and a long duration. 

Note that it has nothing to do with location or threshold. It's really the initial detection which is wrong and showing a parrot instead of a pigeon (that's an example, not a real one). Of course, cross-model consensus never happens if running Birdnet V2.4 on the same source.

I did try on another machine, using a different architecture (AMD64) and things were OK. 

So, I uninstalled the Perch v2 model on the Raspberry and installed it again without picking the recommended version. And everything is now OK again.

### Steps to reproduce

1 - Install the recommended Perch V2 model on a Raspberry Pi5 (FP32 (no-dft) (Central Europe) / 83.87 MB / 873 species
2 - See that detections are always off
3 - Uninstall the recommended Perch V2 model
4 - Install the non-optimized version (ARM64 CPU / FP32 / 406.82 MB / 14 795 species)
5 - Perch V2 is performing as expected / as before

### Expected behaviour

Proper Perch V2 detections on Rapberry Pi 5

### Actual behaviour

Perch v2 is off (as if picking a specie with a offset int the list)

### When did this start?

Worked before, started after I updated

### Relevant log output

```shell
Nothing relevant.
```

### Pre-submission checks

- [x] I am running the [latest release](https://github.com/tphakala/birdnet-go/releases) or have tested with the `:dev` container image.
- [x] I have searched existing open and closed issues and this does not appear to be a duplicate.

Aimache:
system ID : 1FEE-896B-6952
Dump ID disappeared too fast :)

LumenPrima:
[...]

PR #4251, opened 2026-09-03
fix(classifier): force f32 for Perch v2 on the OpenVINO GPU device

LumenPrima:
## Description

Perch v2 compiled at the OpenVINO f16 default returns **all-NaN logits on an Intel Arc A380** (dGPU `8086:56a5`, OpenVINO 2026.2). Every NaN score then passes the `confidence <= threshold` filter, so each 5 s window emits detections for the first few labels of the label set and writes a clip named with a `NaNp` confidence token that the disk manager can never parse or delete. On the reporting host this filled a 32 GB volume in about 11 hours.

The Perch f16-GPU path was only ever validated on an Iris Xe iGPU. BirdNET v2.4 is already forced to f32 on the GPU for a similar f16 miscompile (#3551), so this extends the same guard to Perch in `openVINOPrecisionFor`.

Verified on the Arc A380 with the `north-america-east` `no_dft` fp32 regional model:

| Backend / precision | Result |
|---|---|
| OpenVINO GPU f16 (before) | every logit NaN, 0 real detections, 4 bogus `NaNp` clips per window |
| OpenVINO CPU (f32 internally on this Xeon) | correct detections |
| ONNX Runtime CPU (same file, x86) | correct on silence, DC, clipping, real clips |
| **OpenVINO GPU f32 (this PR)** | correct detections, ~19 ms per inference |

Also:
- Perch's reported runtime precision is now derived from the compiled `INFERENCE_PRECISION_HINT` via `openVINOEffectivePrecision` instead of a hard-coded FP16, so the inference status card shows FP32 on the GPU.
- Precision policy comments, the OpenVINO wiki table, and the hardware-gated Perch functional test updated to match.

**Behavioral note for the maintainer:** the plan carries only the device class (`GPU`), so the override also applies to iGPUs where f16 worked (Iris Xe, ~45 ms at f16 per the wiki). There is no config knob for the precision hint. Correctness was preferred over the f16 throughput there; scoping to discrete Arc parts would need adapter-name probing that does not exist today. Issue #4206 reports the same class of failure (wrong species from the regional `no_dft` model) on a Raspberry Pi 5, which is the OpenVINO **CPU** f16 path; this PR deliberately does not change the CPU path, since I could not validate ARM, but it is very likely the same root cause.

## Related issue

Related: #4206 (Perch `no_dft` on OpenVINO f16, Pi 5 CPU path). Companion PRs: #4252 (non-finite score guard in the classifiers), #4253 (non-finite confidence filter in the processor), #4254 (disk manager tolerance for `NaNp` clip names).

## Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/tphakala/birdnet-go/blob/main/CONTRIBUTING.md)
- [x] Tests pass locally (`go test -race ./...` and/or `npm test`)
- [x] Linters are clean (`golangci-lint run` and/or `npm run check:all`)
- [x] New exports and user-facing changes are documented

## Preflight Status

- [x] Single concern: PR contains exactly ONE fix
- [x] Preflight passed: all Phase 1-3 findings resolved (stale precision docs/comments/wiki updated; functional test asserts the f32 hint on GPU)
- [x] Linters clean: `golangci-lint run` on the changed packages and the whole tree, zero issues
- [x] Tests pass: `go test -race` on `internal/classifier` and the whole tree
- [x] No unrelated changes
- [x] Scope complete; no TODO/FIXME
- [x] No regression/backward-compat break: no config keys, API fields, or migrations touched. Behavioral change: Perch on any Intel GPU now runs f32 (see note above); the status card's `quantization` value for Perch on GPU changes FP16 → FP32 (reporting only).
- [x] Breaking changes documented: wiki table and perf note updated
- [ ] Maintainer-confirm item: scope of the f32 override (all Intel GPUs vs. Arc only) — see note above
- [x] i18n complete: no user-facing strings added
- [x] No secrets or PII
- Secondary-model cross-validation: **not run** (no second model available in the review environment); all review passes were single-model.

## Licensing

- [x] My contribution is licensed under **CC BY-NC-SA 4.0**, and I grant the maintainer the right to relicense it under any **OSI-approved open source license**, as described in the [Relicensing Grant](https://github.com/tphakala/birdnet-go/blob/main/CONTRIBUTING.md#relicensing-grant).

https://claude.ai/code/session_01HBXUYwsiTWinmcQbW8Noo7

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed NaN inference results for Perch v2 when running on Intel Arc GPUs by using 32-bit precision.
  * Runtime precision reporting now reflects the precision actually used by the accelerator.

* **Documentation**
  * Updated OpenVINO performance guidance with revised Iris Xe details and Intel Arc A380 timing information.

* **Tests**
  * Added coverage confirming GPU precision selection and fallback behavior.

### DIFF
--- a/doc/wiki/openvino-acceleration.md
+++ b/doc/wiki/openvino-acceleration.md
@@ -24,7 +24,7 @@ OpenVINO is applied per model, only where it is known to be correct and faster. 
 | Model                                         | OpenVINO eligible | GPU precision | Notes                                                                                                                                   |
 | --------------------------------------------- | :---------------: | :-----------: | --------------------------------------------------------------------------------------------------------------------------------------- |
 | BirdNET v2.4 (stock classifier)               |        Yes        |      f32      | The GPU f16 kernel miscompiles this model, so it is forced to f32 on the iGPU (still faster than ORT CPU). f16 on the ARM A76 CPU path. |
-| Perch v2 (`no_dft` variant)                   |        Yes        |      f16      | Runs its own OpenVINO path. The stock `perch_v2.onnx` (with the DFT layer) is **not** OpenVINO-eligible; the `no_dft` variant is.       |
+| Perch v2 (`no_dft` variant)                   |        Yes        |      f32      | Forced to f32 on the GPU: the GPU f16 kernel returns NaN logits on Intel Arc. f16 on the ARM A76 CPU path. The stock `perch_v2.onnx` (with the DFT layer) is **not** OpenVINO-eligible; the `no_dft` variant is. |
 | BattyBirdNET (bat embedding)                  |        Yes        |      f32      | Forced to f32 on every device (its embedding head overflows at f16).                                                                    |
 | INT8 models (e.g. the arm64 INT8-ARM default) |        No         |       -       | INT8 stays on ONNX Runtime CPU.                                                                                                         |
 | Custom / other models                         |        No         |       -       | Fall back to ONNX Runtime.                                                                                                              |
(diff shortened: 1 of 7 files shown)


LumenPrima:
Addressed the table-driven test nitpick in f046514ad: `TestOpenVINOPrecisionFor` is now a table with one `t.Run` subtest per (model, device) row. No code change.

tphakala:
Thanks for the thorough writeup and the verification table, that made this easy to reason about. Merged as-is: correctness on every Intel GPU is the right call over the bit of f16 throughput on Iris Xe, and having the status card report the compiled precision hint is a nice touch. I'll take the Pi 5 CPU path from #4206 separately.

PR #4268, opened 2026-09-05
fix(audiocore): round native MP3 bitrate to nearest MPEG-1 rate

tphakala:
## Summary

The native go-mp3 clip encoder (opt-in via `BIRDNET_MP3_ENCODER=native`) codes only the 14 fixed MPEG-1 Layer III CBR bitrates, but the export bitrate setting accepts any value in the 32-320k range. A non-standard value such as 100k was rejected by the encoder wrapper, which pushed the clip onto FFmpeg and, on an FFmpeg-less install, silently downgraded it to a lossless WAV. Swapping the user's chosen lossy MP3 for a WAV is surprising and defeats the point of selecting MP3.

This rounds the requested bitrate to the nearest of the 14 valid rates inside the mp3 wrapper (`RoundBitrateKbps`), so the native path always encodes MP3. With rounding in place, the bitrate check drops out of encoder selection and the stranded-clip fallback, leaving only the sample rate (32/44.1/48 kHz) as a reason to fall back. The rounded bitrate is reported through `clipEncoding.BitrateKbps` so the export log and the file on disk agree. The FFmpeg MP3 path is unchanged and keeps the unrounded configured bitrate, which is what FFmpeg is handed.

The default FFmpeg path (gate off) is byte-for-byte unchanged; this only affects installs that opted into the native MP3 encoder.

## Test Plan

- [x] `go test -race ./internal/audiocore/mp3/... ./internal/analysis/processor/...`
- [x] `golangci-lint run` (0 issues)
- [x] `TestRoundBitrateKbps`: nearest-rate mapping, clamping below 32k / above 320k, ties round up
- [x] `TestEncodePCM_RoundsNonStandardBitrate`: encoding 100k succeeds and the frame header carries the rounded 96 kbps bitrate index
- [x] `TestEncodeClip_NativeMP3RoundsBitrate`: the native path records the rounded bitrate in the export log
- [x] `TestEncodeClip_FFmpegMP3KeepsUnroundedBitrate`: the FFmpeg path keeps the unrounded configured bitrate

## Summary by CodeRabbit

* **New Features**
  * MP3 exports now round unsupported configured bitrates to the nearest compatible native-encoding rate.
  * Settings such as 100 kbps can remain MP3 exports instead of being converted to WAV.
  * Native MP3 encoding continues to validate audio format requirements, including sample rate, bit depth, and channel count.

* **Bug Fixes**
  * MP3 export routing no longer falls back solely because of the selected bitrate.

### DIFF
--- a/internal/analysis/processor/actions_database.go
+++ b/internal/analysis/processor/actions_database.go
@@ -812,11 +812,9 @@ func (a *SaveAudioAction) logExportFailure(enc *clipEncoding, exportFormat strin
 // fallback for a clip go-opus cannot carry. AAC and MP3 each have a native
 // encoder too, but both are opt-in while they earn field confidence, so AAC
 // reaches go-aac/go-m4a and MP3 reaches go-mp3 only when the matching gate in
-// internal/conf is set and the encoder accepts the clip's shape (for MP3 that
-// includes the bitrate, since go-mp3 codes only the fixed MPEG-1 rates).
-// Everything else, a non-gated AAC or MP3 clip, and an Opus clip go-opus cannot
-// carry go to FFmpeg.
-func selectEncoder(exportFormat string, exportRate, bitrateKbps int) string {
+// internal/conf is set and the encoder accepts the clip's shape. Everything else,
+// a non-gated AAC or MP3 clip, and an Opus clip go-opus cannot carry go to FFmpeg.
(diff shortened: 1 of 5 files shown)


tphakala:
Good call, done in 6381c9d14: `TestRoundBitrateKbps` is now table-driven with a named subtest per case (specials plus each of the 14 valid MPEG-1 rates).

PR #4269, opened 2026-09-05
test(audiocore): pin FFmpeg stream ingest with a producer-agnostic contract suite

tphakala:
## Summary

BirdNET-Go is going to replace the per-stream FFmpeg ingest subprocess with a pure-Go network stream path. Before any of that refactoring starts, this PR pins the current FFmpeg ingest behaviour with characterization tests, so the later work can be proven to preserve it 1:1. Tests come before the refactor, on purpose.

The tests are producer-agnostic. A new importable test-support package `internal/audiocore/streamtest` defines a single contract suite, `RunManagerContract`, that drives any stream producer through a narrow manager interface and reads its health through a small health interface. It is parametrized over a manager factory and a stream fixture, so the exact same cases will run unchanged against the native producer later. Today it runs against `ffmpeg.NewManager` via a thin adapter and a MediaMTX container. Production code never imports `streamtest`.

## What's here

- `internal/audiocore/streamtest/`: the contract suite (13 manager-seam cases: frame shape, PCM fidelity across the codec matrix, data rate, lifecycle, reset callback, health transitions, silence restart, server-gone reconnect, media modes, multistream, high-rate passthrough, error clarity, time-to-first-frame) plus pure Goertzel dominant-frequency and RMS helpers (unit tested, reusing `internal/audiocore/convert` for the PCM decode), and a MediaMTX-backed fixture behind the `integration` build tag.
- `internal/audiocore/ffmpeg/manager_contract_test.go` (`integration`): runs the contract against `ffmpeg.NewManager` over a MediaMTX container across the RTSP codec matrix (G.711 mu/A-law, AAC-LC, Opus, L16 at 48k and 96k).
- `internal/audiocore/engine/engine_contract_test.go` (`integration`): engine reconfigure and stop/start round-trips, plus the router + liveness-watchdog + manager recovery chain.
- The one production change: a new `ffmpeg.NewManagerWithOptions` constructor taking `ffmpeg.Options{SilenceTimeout}`, so the silence-restart path can be driven in seconds instead of the 90s production default. `NewManager` delegates with a zero `Options` (behaviour unchanged), and the timeout is read per stream via a new `StreamConfig.SilenceTimeout` field that falls back to the existing 90s constant. It is unit-tested directly (`handleSilenceTimeout`) and never mutates the caller's config.
- `internal/testutil/containers`: the MediaMTX publisher now covers the codec matrix (`PublishToneToMediaMTX`), and the container gained `Stop`/`Start` for the server-outage case.

Integration tests skip cleanly when no container runtime is present, so they do not affect the normal unit-test run.

## Characterization findings recorded as the baseline

Two current FFmpeg behaviours are documented (not treated as bugs to fix here):

- Health error classification only runs during the stream's early-detection and quick-exit windows (both about 5s), so fast failures classify (`rtsp_404`, `connection_refused`) but a slow connect timeout (10s, unreachable host) leaves the health error context empty. The error-clarity case asserts the fast failures and records the timeout as an accepted gap the native path is expected to close.
- After manager shutdown the only lingering goroutine is the Go stdlib `os/exec.(*Cmd).watchCtx` for the just-killed child, which the runtime reaps; the lifecycle goleak check ignores that function.

## Test Plan

- [x] `go build ./...`, `go vet ./...`, and `golangci-lint run` (default and `--build-tags=integration`) are clean.
- [x] Unit tests: `go test -race ./internal/audiocore/...`.
- [x] Integration (local, podman + FFmpeg 7.1.1): all 13 manager-seam cases and both engine-level cases pass; baselines captured and reproduced across runs.

## Summary by CodeRabbit

- **New Features**
  - Added configurable silence timeouts for audio streams, with manager-level defaults and per-stream overrides.
  - Improved stream recovery after interruptions, reconnects, and server restarts.

- **Bug Fixes**
  - Improved detection and reporting of silent or unhealthy audio streams.

- **Tests**
  - Expanded coverage for audio fidelity, stream health, lifecycle behavior, codec compatibility, and recovery scenarios.

### DIFF
--- a/internal/audiocore/engine/engine_contract_test.go
+++ b/internal/audiocore/engine/engine_contract_test.go
@@ -1,0 +1,234 @@
+//go:build integration
+
+package engine
+
+import (
+	"context"
+	"os/exec"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/tphakala/birdnet-go/internal/audiocore"
+	"github.com/tphakala/birdnet-go/internal/audiocore/buffer"
+	"github.com/tphakala/birdnet-go/internal/audiocore/ffmpeg"
+	"github.com/tphakala/birdnet-go/internal/audiocore/streamtest"
+	"github.com/tphakala/birdnet-go/internal/testutil/containers"
+)
+
+// Engine-level contract timing (integration; real container and FFmpeg children).
+const (
+	engineHealthyBudget = 25 * time.Second
+	enginePollInterval  = 250 * time.Millisecond
+	engineSampleRate    = 48000
+	engineBitDepth      = 16
+	engineChannels      = 1
+
+	// Shortened liveness thresholds so the chain runs in seconds.
+	livenessCheckInterval = 1 * time.Second
+	livenessSilence       = 3 * time.Second
+	livenessMaxRetries    = 20
+	livenessRetryBackoff  = 2 * time.Second
+	livenessCooldown      = 2 * time.Second
+	livenessEscalation    = 90 * time.Second
+	livenessRestartBudget = 30 * time.Second
+	livenessRecoverBudget = 60 * time.Second
+
+	// managerShutdownCleanupTimeout bounds the manager shutdown in test cleanup.
+	managerShutdownCleanupTimeout = 15 * time.Second
+)
+
+// TestEngineIngestContract runs the Phase 0 engine-level characterization cases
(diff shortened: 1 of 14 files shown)


tphakala:
@coderabbitai both nitpicks fixed in fd3642cb7: caseLifecycle now registers a best-effort `t.Cleanup` shutdown right after building the manager (so a require failure before the explicit shutdown cannot leak the manager and its FFmpeg child into the sibling subtests that share the fixture; the explicit `ShutdownWithContext` still owns the goleak assertion), and `RefusedPortURL` now uses a named `fallbackRefusedPort` constant instead of the bare literal. Re-ran the Lifecycle integration case, still green.

PR #4282, opened 2026-09-06
fix(audiocore): persist lossy export bitrate default and resample sub-48k clips

tphakala:
## Summary

An enabled lossy audio export (aac/opus/mp3) with a blank bitrate on disk (`bitrate: ""`) was repaired in memory on every load but never written back, so the "audio export is enabled with the lossy format ... but no bitrate is set; using the default 96k" warning re-emitted on every restart. This heals it once: a load-time migration fills the documented default and persists it, running before the incomplete-feature normalization so a switched-on-but-unconfigured integration's `enabled: true` is never flipped to `false` on disk. `omitempty` is added to the `Bitrate` YAML tag so a blank is never written back to defeat the viper default on reload; the JSON tag is unchanged, so the settings API still always emits the field.

It also stops stranding sub-48kHz clips to WAV on the native encoders: when the configured lossy native encoder cannot carry the source rate and no FFmpeg is present, the clip is resampled up to 48kHz (accepted by every native encoder) instead of silently downgrading the format to WAV. This is native-path only; FFmpeg-present installs and bat clips are unchanged, and a resample failure still falls back to WAV so the recording always survives. The existing downsample path and the new upsample path now share one `resampleExportTo` helper.

## Test Plan

- conf round-trip per format (aac/opus/mp3): a config carrying an empty lossy bitrate self-heals on load, persists `96k` to disk, and does not re-warn on the next load
- a save made while export is disabled no longer writes `bitrate: ""` (omitempty), and the load-time write-back never persists an incomplete integration's disable
- `resolveExportParams` resamples sub-48k native clips to 48k instead of stranding, keeps encoder-supported rates at their source rate, still strands bat clips to WAV, and falls back to WAV on a resample failure
- `go test -race ./internal/conf/ ./internal/analysis/processor/` and `golangci-lint run` pass

## Summary by CodeRabbit

* **New Features**
  * Bird audio exports below 48 kHz can be resampled to 48 kHz when required, preserving the selected lossy format instead of falling back to WAV.
  * Empty AAC, Opus, and MP3 bitrate settings are automatically repaired with documented defaults and saved for future use.

* **Bug Fixes**
  * Improved handling of unsupported sample rates and native encoding scenarios.
  * Prevented temporary configuration repairs from unintentionally disabling incomplete audio exports.

### DIFF
--- a/internal/analysis/processor/actions_database.go
+++ b/internal/analysis/processor/actions_database.go
@@ -1497,9 +1497,12 @@ func (a *SaveAudioAction) planNativeNormalizationGain(ctx context.Context, sampl
 }
 
 // resolveExportParams determines the export sample rate, format, and output
-// path. Bird audio at rates above 48kHz is downsampled. Bat audio keeps the
-// native rate; if the configured format cannot carry it, the format is
-// silently switched to WAV.
+// path. Bird audio above 48kHz is downsampled to 48kHz. Bird audio below 48kHz
+// whose configured lossy native encoder cannot carry the source rate is resampled
+// UP to 48kHz (which every native lossy encoder accepts) so the configured format
+// is kept rather than stranded to WAV, on an install with no FFmpeg to take the
+// source rate directly. Bat audio keeps its native rate and is never resampled; if
+// the configured format cannot carry it, the format is switched to WAV.
 func (a *SaveAudioAction) resolveExportParams(outputPath string) (rate int, format, path string) {
 	rate = a.sourceSampleRate
 	if rate <= 0 {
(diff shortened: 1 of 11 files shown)


tphakala:
Good catch on the repeated bat model literal, switched those three test sites to the existing batModelName constant in 3e3c795.

tphakala:
On bundling the two fixes: that's intentional. Both changes live in the native clip-export path and came out of the same investigation into export correctness, so I'm keeping them in one PR. They don't overlap in code (config-load migration vs resolveExportParams resampling) and each carries its own tests, so a revert of either is still clean on its own.

PR #4290, opened 2026-09-07
fix(classifier): stop the model-unload predict storm and BirdNET v3.0 preview follow-ups

tphakala:
## Summary

Follow-ups for the BirdNET v3.0 developer preview, plus a general fix for a model-unload predict storm that affects every secondary model, not just v3.0.

**Stop the model-unload predict storm (Sentry BIRDNET-GO-2G6, BIRDNET-GO-1S1).** When a model is unloaded (uninstall, reinstall, or variant switch) the orchestrator drops it from its loaded set immediately, but the analysis monitor for that source is torn down asynchronously, after the topology-reconfigure debounce. During that window the monitor kept dispatching audio windows to a model the orchestrator no longer had, logging an "unknown model" error on every window: 2G6 fired for BirdNET v3.0 and 1S1 for Perch v2 (285 events). The monitor now skips inference for a model that is not loaded, warning once and resuming automatically when the same model is re-registered, so a reconfigure no longer spams the log or Sentry. The monitor keeps running rather than stopping, so a reinstall or variant switch resumes immediately while the reconcile still tears it down when the model is gone for good. The predict path also turns the bare "unknown model" into an ErrModelNotLoaded carrying an actionable reason (shut down, unknown registry, not enabled, failed to load with the last error, unloaded, or still loading), and startup optional-model load failures are now recorded so a later not-loaded predict can explain itself.

**Make the fp16-native recommender reward backend-aware.** The model-gallery recommender gave an fp16 variant a native-f16 speed bonus on any host with native half-precision SIMD. After BirdNET v3.0 was forced to FP32 on every OpenVINO device (PR #4289, Sentry BIRDNET-GO-2H6), that bonus is wrong on the OpenVINO path, where the fp16 file runs at f32. The reward is now suppressed when the variant's selected backend forces f32, via a predicate that wraps the existing per-model precision policy rather than hardcoding a model ID in the recommender. It still applies on the ONNX Runtime path and for a model whose OpenVINO path keeps f16. The rule only ever removes a reward, and the measured scores show no shipped recommendation flips.

**Derive Bat OpenVINO precision from the compiled plan.** NewBat hardcoded FP32 rather than reading the precision the OpenVINO plan compiled at, the way the v3.0 path does. It is correct only because the bat embedding model is forced to f32 on every device today; deriving it from the plan removes the latent silent-lie if that policy ever changes.

**Document the v3.0 RAM floors.** Both v3.0 RAM floors were sourced from the upstream manifest under f16 compilation. Forcing f32 on OpenVINO raises the real footprint, but no f32-on-OpenVINO RSS has been measured for either variant, so the floors are left as sourced with a comment recording why.

## Related Issues

Follows up PR #4289 (the BirdNET v3.0 OpenVINO FP32 fix). No public issue is linked.

## Test Plan

- [x] `go build ./...`, `go vet` (default, `notflite`, `noembed` tags)
- [x] `golangci-lint run` clean on the changed packages
- [x] `go test -race ./internal/classifier/... ./internal/analysis/... ./internal/api/v2/models/... ./internal/api/v2/system/...`
- [x] New tests: monitor skips a not-loaded model without dispatching; the not-loaded reason distinguishes each cause and survives a fail-recover-unload sequence; the fp16 reward is suppressed on OpenVINO for v3.0 but kept on ONNX Runtime and for a model whose OpenVINO path keeps f16; the Bat OpenVINO decline path reports no precision.

## Summary by CodeRabbit

* **New Features**
  * Added clearer model availability diagnostics, including specific reasons when a model cannot be used.
  * Added tracking for model loading failures and successful reloads.
  * Added backend-aware precision handling for OpenVINO and ONNX Runtime configurations.

* **Bug Fixes**
  * Monitoring now continues safely when a configured model is unavailable and resumes inference once it returns.
  * Prevented repeated warnings during consecutive monitoring cycles.
  * Improved model recommendations based on backend-specific precision capabilities.

### DIFF
--- a/internal/analysis/buffer_manager.go
+++ b/internal/analysis/buffer_manager.go
@@ -30,6 +30,16 @@ type monitorConfig struct {
 	overlapSize int // bytes, scaled from user config, PCM-aligned
 }
 
+// monitorTickState is the per-monitor state threaded across processMonitorTick
+// calls by analysisBufferMonitor. hasReadBuffer flips the "buffer removed" log
+// level once the buffer has been seen at least once; notLoadedWarned is the
+// warn-once latch for the model-not-loaded skip, so a monitor that outlives its
+// model during a reconfigure logs one warning rather than one per window.
+type monitorTickState struct {
+	hasReadBuffer   bool
+	notLoadedWarned bool
+}
+
 // BufferManager handles the lifecycle of analysis buffer monitors
 type BufferManager struct {
 	monitors  sync.Map // keyed by monitorKey -> chan struct{}
@@ -392,7 +402,7 @@ func (m *BufferManager) analysisBufferMonitor(quitChan chan struct{}, cfg *monit
 	const pollInterval = 100 * time.Millisecond
 
 	analysisWindowBytes := cfg.readSize
-	hasReadBuffer := false
+	var state monitorTickState
 	var tickCount int64
 
 	ticker := time.NewTicker(pollInterval)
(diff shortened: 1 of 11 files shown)


tphakala:
Addressed the CodeRabbit nitpick in 44b692efa: the raw "cuda" token in TestBackendForcesFP32 is now a named cudaBackendToken constant, per the no-magic-strings rule. No behavior change.

PR #4359, opened 2026-09-13
refactor(classifier): restore unload-failure rollback coverage; stop range-filtering unknown-ID classifiers

tphakala:
## Summary

Two folded follow-ups from the classifier model de-privilege work.

Restore unload-failure rollback test coverage. Phase 3 removed the primary-model unload refusal that three model-manager tests used to drive the Uninstall, Reinstall, and replaceVariant unload-failure abort branches, leaving those tests skipped. This adds a test-only unload seam on ModelManager (a nil-in-production `unloadFn` function field, the same shape as the existing `freeSpaceFn` seam) and injects a failure in the three tests. The abort branches are still reachable in production via a concurrent unload or delete racing the loaded-check, so the coverage loss was real, not dead code.

Stop range-filtering unknown-ID classifiers. `ParticipatesInRangeFilter` returned true for IDs absent from the model registry, preserving a legacy display-name gate; it now returns false. A custom classifier has its own label space, so gating it against an inclusion list built for a known model would silently drop labels, and the registry already reports no range-filter compatibility for unknown IDs. The branch is unreachable in production (the orchestrator rejects unregistered IDs at load and skips them at startup), so this pins the intended semantics with no observable behavior change.

## Test plan

- `go test -race ./internal/classifier/... ./internal/analysis/processor/...` passes, including the three restored unload-failure tests and the new `TestParticipatesInRangeFilter_UnknownNotFiltered`.
- `go test -tags notflite ./internal/classifier/... ./internal/analysis/processor/...` passes.
- `golangci-lint run` reports 0 issues.

## Summary by CodeRabbit

- **Bug Fixes**
  - Unknown or custom model IDs no longer participate in geographic range filtering; only recognized models use configured filtering behavior.
  - Model uninstall, reinstall, and replacement operations now stop safely if a model cannot be unloaded, preserving the existing installation and files without starting downloads or overwrites.

- **Reliability**
  - Improved handling of model-management failures to prevent unintended changes when models remain in use.


tphakala:
Two notes on the review feedback:

Scope: the two changes are folded on purpose. They're both Phase 3 test-coverage follow-ups on the same classifier surface (restoring the unload-failure rollback tests, and pinning the unknown-ID range-filter gate), and I'd rather keep tightly related follow-ups together than split them into two tiny PRs.

threads_test.go named constants: leaving the table fixtures as inline literals. In a table-driven test the inputs are meant to read inline; the values here are self-documenting test inputs (a thread count and throwaway synthetic model names), not domain magic numbers, and the real registry IDs already use the RegistryID* constants. The six tests this consolidates used the same inline literals.

PR #4379, opened 2026-09-16
refactor(classifier): DefaultTargets final semantics and one-shot source-target migration (Phase 4 PR 1)

tphakala:
## Summary

Phase 4 PR 1 of the model de-privilege work. It gives `Orchestrator.DefaultTargets()` its final semantics and makes every consumer of an empty per-source model list consistent with it, with no behavior change for existing installs. There is no N=0 runtime, no audio-engine change, and no frontend here; those land in later PRs.

`DefaultTargets()` now returns every loaded model that is not schedule-gated (today only the bat model), BirdNET v2.4 first and the rest in registry-ID byte order, nil when none qualify. v2.4 leads so the audio engine's pre-allocated analysis buffer keeps its geometry until the engine's primary concept is removed later. The schedule-gated decision moves from a hard-coded `modelID != RegistryIDBat` in `IsModelActive` and `ModelScheduleStatus` to a `scheduleGated` flag on the bat registry entry, read through one `isScheduleGated` predicate the three call sites share.

An empty per-source or per-stream model list changed meaning: it used to mean "the built-in BirdNET v2.4" and now means "the orchestrator's default targets", which can include secondaries. To keep every existing install on exactly the models it analyzed with before, the per-load fill in `MigrateSourceModels` becomes a one-shot `MigrateSourceTargetDefaults` gated by a new `Settings.ConfigVersion`: it pins every pre-Phase-4 empty list to `["birdnet"]` once, then stamps the version so an intentionally emptied list (a source uninstalled down to the defaults) survives a restart instead of being re-filled. `ConfigVersion` is persisted but hidden from the settings API and copied by value through `CloneSettings`, so writers never drop it.

The fallback for a source that resolves to no loaded target splits by intent, to hold the epic's I1 invariant that an upgrade never makes an existing install gain or lose a model: an empty list fans out to every default target, but a non-empty list whose models are all unknown or unloaded (a misconfigured source) falls back to the first default only (v2.4), matching pre-Phase-4 behavior. The runtime allocator, the reconfigure change-detection and the `/system/inference` status view all apply the same split, so they stay in parity and no reconfigure storm or status desync can arise.

Uninstalling a model no longer re-pins an emptied source or stream list to `["birdnet"]`; an emptied list means the default targets.

## Test Plan

- [x] `golangci-lint run` clean across the changeset
- [x] `go test -race` on `internal/classifier`, `internal/analysis`, `internal/api/v2/system`, `internal/conf`
- [x] `go vet` across the tflite / onnx / openvino / noembed / notflite build-tag matrix
- [x] Config schema regenerated; `cmd/gen-schema` `TestSchemaUpToDate` green
- [x] New table tests pin the v2.4-first ordering (including a default that byte-sorts before "BirdNET_V2.4"), the one-shot migration gate, the empty-vs-unresolvable fallback split, and the status fan-out
- [x] The multimodel invariance golden records the expanded default-target set while `EngineDims` stays byte-identical, proving the engine's pre-allocated buffer geometry is preserved

## Summary by CodeRabbit

- **New Features**
  - Added automatic configuration migration tracking with a preserved version value.
  - Added support for multiple default analysis targets.
  - Empty model assignments now use all eligible defaults; unresolved explicit assignments use the first default.
  - Schedule-limited models are excluded from default targets and remain inactive outside their schedules.

- **Bug Fixes**
  - Model removal no longer unnecessarily reassigns sources or streams to a specific model.
  - Existing configurations are migrated once while preserving their intended default behavior.

- **Documentation**
  - Documented configuration migration tracking and preservation behavior.


tphakala:
Fixed in 43c42a6: registered the new ConfigVersion field in hotReloadRegistry as hotReloadRuntime (an internal, non-user-editable migration marker), which is what TestSettingsHotReloadCoverage was flagging.

Issue #3953, opened 2026-07-17
bug: Unifi rtsp "silence detected" and "recovered" loop


klushie:
### BirdNET-Go version / build date

2026-07-16T20:05:03Z

### Install method

Home Assistant add-on (community, third-party)

### Device type

PC / Server (x86_64)

### Operating system and version

Intel NUC running HAOS (Debian Linux)

### Audio source

Network audio stream - RTSP / RTSPS (e.g. IP camera)

### Hardware details

RAM: 8GB
Storage: 120GB
Audio: RTSP stream from Unifi G3 Flex

### Description

Recently installed BirdNet-GO over HAOS and it was working fantastically for the first couple weeks, but recently it's looping a few errors regarding the rtsp stream. 

The most common error is:
> Audio source silence detected
> Source rtsp_c5168874: silence detected

After a min or two, it recovers the audio source.

But I am also getting:
> Audio stream error
> Deck Camera: early FFmpeg error (transient): Stream ended unexpectedly

> Audio stream error
> Deck Camera: FFmpeg process failed to start properly: Stream ended unexpectedly

I'm still getting plenty of detections and audio clips saved, so it's still working I suppose, but the constant errors are less than ideal.

### Steps to reproduce

Wait for the 300 second timeout and get another discord msg push with an error.

### Expected behaviour

Listen to RTSP stream without errors.

### Actual behaviour

Audio source silence detected
Source rtsp_c5168874: silence detected

Audio stream error
Deck Camera: early FFmpeg error (transient): Stream ended unexpectedly

Audio stream error
Deck Camera: FFmpeg process failed to start properly: Stream ended unexpectedly

### When did this start?

Worked before, started without any update

### Relevant log output

```shell

```

### RTSP source (only if the bug is RTSP related)

Unifi G3 Flex.
High res RTSP link (I also tried medium and low res. Still got the errors)

### Pre-submission checks

- [x] I am running the [latest release](https://github.com/tphakala/birdnet-go/releases) or have tested with the `:dev` container image.
- [x] I have searched existing open and closed issues and this does not appear to be a duplicate.

PR #4407, opened 2026-09-22
fix(analysis,alerting): name the audio source in its notifications

gilindoesgit:
## Problem

Audio source notifications identify the source by its internal ID, or not at all, so a user running several cameras cannot tell which one to go and check.

This is visible in #3953, where the reporter's own notifications show both halves of the problem side by side:

> Audio source silence detected
> **Source rtsp_c5168874: silence detected**

> Audio stream error
> Deck Camera: early FFmpeg error (transient): Stream ended unexpectedly

The first is the liveness watchdog naming the source `rtsp_c5168874`. The second gets it right — but only because that notification went out as the English fallback. In the notification bell, where the message is rendered from its i18n key, the name is dropped entirely and the body reads just `error reading from FFmpeg: stream ended without producing data`.

Two independent causes:

### 1. The liveness watchdog formats bodies from the source ID

`livenessNotifier.notify` built every body out of `sourceID`, so flapping, silence, recovery and escalation alerts all read `Source rtsp_832ca5de: …`.

It now takes a name resolver, wired in `audio_pipeline_service.go` to the source registry's `DisplayName`. The name is resolved per notification rather than captured once, so renaming a source in settings applies without a restart.

Two details worth calling out:

- Burst coalescing still keys on `sourceID`, not the name. The ID is stable for the life of the source, whereas keying on a mutable name would let a rename mid-window split one flapping incident into two buckets.
- The fallback runs the ID through `privacy.SanitizeRTSPUrl`, because legacy call paths can pass a raw connection string, and an RTSP URL with embedded credentials must never reach a user-visible notification.

### 2. The localized alert template discarded the source name

Stream and device error events already carry the name in `stream_name`, and `formatErrorFallback` prepends it to the English fallback. But the i18n template was bare:

```json
"errorOccurred": "{error}"
```

so the translated path — which is what the bell renders — threw the name away. That is why #3953 shows the name via Discord but the bell does not.

Error events that name a source now use a new `errorWithSource` key rendering `"{source_name}: {error}"`, matching what `formatErrorFallback` already produced. A classified error (`timeout`, `authError`, …) passes its own key as a nested `error_key` param, resolved on the frontend before substitution — the same nested-key pattern already used for `rule_name_key`, now factored into a shared `resolveNestedKeyParam` helper. Events that cannot name a source keep the existing keys, so nothing renders with an empty `": "` prefix.

## Related

- Fixes the notification-naming half of #3953 (`Source rtsp_c5168874: silence detected`). The underlying flapping loop in that issue was addressed separately; this makes the alerts it produces legible.
- Context: #4080 (liveness watchdog / RTSP recovery backoff) — same notifier, unrelated cause.
- Adjacent theme, not addressed here: #2813 asks for per-stream attribution on *detections*, a different subsystem.

## Testing

- `golangci-lint run` — 0 issues across the project
- `go test -race ./internal/alerting/ ./internal/analysis/` — pass
- `npm run typecheck` (svelte-check) — 5346 files, 0 errors
- `npx vitest run src/lib/utils/notifications.test.ts` — 35 pass
- `npx prettier --check` — clean

New coverage:

- `TestLivenessNotifier_UsesDisplayName` — every body names the source and no body leaks the internal ID, across the transient, summary and critical-bypass paths
- `TestLivenessNotifier_UnknownSourceFallsBackToID` — a registry miss still identifies the source rather than rendering blank
- `TestDispatcher_DefaultTemplate_ErrorMessage_UnclassifiedWithSource` — the raw FFmpeg read error case from #3953
- `TestDispatcher_DefaultTemplate_ErrorMessage_ClassifiedNoSource` — no empty prefix when the event cannot name a source
- Three frontend cases covering the wrapper, nested `error_key` resolution, and fallback when the nested key is unknown

## Notes for review

- `errorWithSource` is a punctuation-only template, so `i18n:sync` filled the 15 non-English locales with the English string verbatim. That is correct for all but `fr`, which I gave the French narrow no-break space before the colon (`{source_name} : {error}`). The sync script still flags the key as untranslated.
- This touches `frontend/static/messages/*.json`, `types.generated.ts` and `src/test/setup.ts`, which overlap with my open #4327 and #4387 — expect a trivial i18n merge.

## Summary by CodeRabbit

* **New Features**
  * Alert notifications now include the affected source name alongside the error.
  * Classified errors are displayed using their translated descriptions, with raw-error fallback when no translation is available.
  * Liveness notifications now show friendly source names instead of internal identifiers.
  * Missing source names fall back to a sanitized identifier without exposing connection credentials.

* **Localization**
  * Added translated alert templates for source-aware error messages across supported languages.


gilindoesgit:
@coderabbitai review

